PR #8, opened 2019-12-16
WIP: Login code split

sorja:
Added codesplitting and removed extra config from webpack configuration.

TODO: 
- Validate if removing `loginHandler.init(app)` affects the app


webapp/commonComponents/dynamicImport.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import { injectReducers } from '../state/store'
+
+export default class DynamicImport extends React.Component {
```
minotogna: can we make it functional ?

webapp/loggedin/appView.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import LoggedInView from './loggedInView'
+
+const AppView = props => {
+  return (
+    <>
```
minotogna: you don't need this component. AppView is LoggedinView

PR #22, opened 2020-01-21
originalDataPointView: Use history.goBack not window.location

sorja:


webapp/originalDataPoint/originalDataPointView.js
```diff
@@ -35,13 +36,17 @@ const OriginalDataPoint = (props) => {
...
   } = props
-
   const { countryIso, tab } = match.params
-
+  const history = useHistory()
   const saveControlsDisabled = () => !odp.odpId || autoSaving
   const yearValidationStatusClass = () => odp.validationStatus && !odp.validationStatus.year.valid ? 'error' : ''
   const unselectable = R.defaultTo([], odp.reservedYears)
 
+  const handleSave = () => {
```
minotogna: I would pass the history to the action and after awaiting the response from the server, I would call ```history.goBack()```. This is to make sure that the changes were applied on the server side

PR #45, opened 2020-02-17
Feature/table to csv

sorja:
Uses cheerio-tableparser inspired algorithm (`utils.js`) for parsing table data with col and row spans.

### DIFF
--- a/webapp/traditionalTable/singleTraditionalTableView.js
+++ b/webapp/traditionalTable/singleTraditionalTableView.js
@@ -4,7 +4,7 @@
  * just do a custom view and use TraditionalTable, CommentableDescriptions etc. directly instead!
  */
 
-import React from 'react'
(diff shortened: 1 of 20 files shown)


webapp/traditionalTable/singleTraditionalTableView.js
```diff
@@ -94,10 +94,13 @@ class SingleTraditionalTableView extends React.Component {
             <div className="page-break" />
           }
 
-          <TraditionalTable tableSpec={tableSpecInstance}
-                            countryIso={match.params.countryIso}
-                            disabled={isEditDataDisabled}
-                            tableDataFetched={true}/>
+          <TraditionalTable 
+            tableSpec={tableSpecInstance}
+            countryIso={match.params.countryIso}
```
minotogna: Can we get countryIso from state object?

PR #47, opened 2020-02-21
Initial version of authMiddleware

sorja:
Initial version of the authMiddleware
Note: `serverInit.js`
```
  // accessControl.init(app)

```


server/auth/authApiMiddleware.js
```diff
@@ -0,0 +1,27 @@
...
+  const countryIso = req.params.countryIso || req.query.countryIso
+  const user = getUser(req)
+  const userName = getUserName(req)
+
+  try {
+    checkCountryAccessFromReqParams(req)
+    await allowedToEditDataCheck(countryIso, user, section)
+    next()
+  } catch (error) {
+    sendErr(res, error)
```
minotogna: should this be ```next(error)```?

server/auth/authApiMiddleware.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+
+const requireAdminPermission = async (req, res, next) => {
+  console.log({ req, res, next })
```
minotogna: I guess you can use serverUtils/accessControl.checkAdminAccess

sorja: This has not yet been implemented, thanks for the tip though

server/auth/authApiMiddleware.js
```diff
@@ -0,0 +1,34 @@
+const { checkCountryAccessFromReqParams, checkAdminAccess } = require('../utils/accessControl')
+const { getUser } = require('../utils/requestUtils')
+const { allowedToEditDataCheck } = require('../assessment/assessmentEditAccessControl')
+
+const requireCountryEditPermission = async (req, res, next) => {
+  const section = req.params.section || req.query.section
+  const countryIso = req.params.countryIso || req.query.countryIso
```
minotogna: in RequestUtils we could add a method like:
```
export const getParams = req =>
  R.pipe(
    R.mergeLeft(R.prop('query', req)),
    R.mergeLeft(R.prop('params', req)),
    R.mergeLeft(R.prop('body', req)),
    // Convert String boolean values to Boolean type
    R.mapObjIndexed(val => R.ifElse(v => v === 'true' || v === 'false', R.always(val === 'true'), R.identity)(val)),
  )({})
```
so that your code above would become:
```
const {countryIso, section} = Request.getParams(req)
```

what do you think?

server/auth/authApiMiddleware.js
```diff
@@ -0,0 +1,34 @@
...
+  const user = getUser(req)
```
minotogna: I would import Request as object.
this could become:
```
const user = Request.getUser(req)
```
what do you think?

server/user/userApi.js
```diff
@@ -59,28 +61,18 @@ module.exports.init = app => {
   })
 
   // get all users / only admin can access it
-  app.get('/users', async (req, res) => {
-    try {
-      if (isAdministrator(req.user)) {
-        const url = serverUrl(req)
+  app.get('/users', Auth.requireAdminPermission, async (req, res) => {
+    const url = serverUrl(req)
```
minotogna: I think if you omit try catch, the app will crash

server/user/userApi.js
```diff
@@ -242,35 +224,25 @@ module.exports.init = app => {
...
-        const url = serverUrl(req)
-
-        const invitations = await userRepository.fetchAllInvitations(url)
-        const sendInvitationPromises = invitations.map(async invitation => {
-
-          if (validEmail(invitation)) {
-            await sendInvitation(invitation.countryIso, invitation, req.user, url)
-            return `<p>Email sent to ${invitation.name} (${invitation.email}) invited as ${invitation.role} for ${invitation.countryIso}</p>`
+  app.get('/users/invitations/send', Auth.requireAdminPermission, async (req, res) => {
+    const url = serverUrl(req)
```
minotogna: try catch as above

server/utils/accessControl.js
```diff
@@ -36,8 +36,10 @@ const checkAdminAccess = (user) => {
 // WARNING: the param name needs to be exactly 'countryIso'
 // If it's not, use checkCountryAccess instead
 const checkCountryAccessFromReqParams = (req) => {
-  if (req.params.countryIso) checkCountryAccess(req.params.countryIso, req.user)
-  if (req.query.countryIso) checkCountryAccess(req.query.countryIso, req.user)
+  if (req.user) {
```
minotogna: is this ```if(req.user``` still needed ?

sorja: I need to figure a better way - the underlying methods are using the user and failing if the user doesn't exist I think ( user = 'undefined' )

minotogna: Right, but this method should be invoked only if the user exists. Isn't it so?

PR #53, opened 2020-02-25
Feature/public view auth mid latest schema

sorja:
Also contains public api endpints for:
- GET /odp
- GET /growingStock/:countryIso


server/versioning/service.js
```diff
@@ -0,0 +1,24 @@
...
+// 2b. User doesn't exist: Return latest schema version ('frozen version')
+const getDatabaseSchema = async (req) => {
+  const user = Request.getUser(req)
+  if (user) {
+    return defaultSchema
+  }
+
+  // Destruct array => object => string
+  const [{schemaName}] = await getLatestSchemaVersion()
+  return schemaName
```
minotogna: What if there's no schema? Does it return public?

PR #54, opened 2020-02-26
requireCountryEditPermission: fix section check

sorja:


server/auth/authApiMiddleware.js
```diff
@@ -9,7 +9,8 @@ const requireCountryEditPermission = async (req, res, next) => {
 
   try {
     checkCountryAccessFromReqParams(req)
-    if (section) {
+    // Section is returned as string, check if it's not 'undefined'
+    if (section !== 'undefined') {
```
minotogna: Would it be safer section && section !== 'undefined'

sorja: Sounds good

PR #55, opened 2020-02-26
Feature/public view country selection

minotogna:


webapp/loggedin/countrySelection/components/countryList.js
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <div className="country-selection-list">
+      {
+        isAdministrator(userInfo) &&
+        <div className="country-selection-list-download">
+          <CSVLink
+            className="btn-s btn-primary"
+            target="_blank"
+            filename="FRA-Countries.csv"
+            data={
```
sorja: Should data prop value be defined somewhere else for readability?

minotogna: yes. I have also extracted it to a separate component

PR #69, opened 2020-02-27
Added git revision webpack plugin for version info

sorja:


webapp/landing/views/aboutView.js
```diff
@@ -46,7 +46,7 @@ const AboutView = () => {
 
     <Logos i18n={i18n}/>
 
-    <div className="landing__version">{i18n.t('navigation.support.platformVersion')} {__PLATFORM_VERSION__}</div>
+    <div className="landing__version">{i18n.t('navigation.support.platformVersion')} <span title={__PLATFORM_VERSION__}>{__APPLICATION_VERSION__}</span></div>
```
minotogna: do we still need PLATFORM_VERSION?

package.json
```diff
@@ -75,6 +75,7 @@
     "express": "^4.15.2",
     "express-fileupload": "^1.1.6",
     "express-session": "^1.15.3",
+    "git-revision-webpack-plugin": "^3.0.4",
```
minotogna: this should go to dev dependencies

webapp/landing/views/aboutView.js
minotogna: do we still need PLATFORM_VERSION?

package.json
minotogna: it is still in dependencies. why not in dev dependencies ?

webapp/loggedin/navigation/components/navigationFooter.js
```diff
@@ -62,7 +62,8 @@ const NavigationFooter = () => {
         </>
       }
 
-      <span className="nav__footer-copyright">&copy; {currentYear} FAO</span>
+      <span className="nav__footer-copyright">&copy; {currentYear} FAO <br/>{i18n.t('navigation.support.platformVersion')} {__APPLICATION_VERSION__}
```
minotogna: this should not go here. but replaced to NavigationFooterSendFeedback

PR #110, opened 2020-03-06
AppLinks: divider

sorja:


webapp/app/components/header/components/AppLinks.js
```diff
@@ -20,6 +20,7 @@ const AppLinks = ({ i18n }) => {
       className="app-header__menu-item">
       FRA Platform
     </NavLink>
+    <div className="nav__divider"/>
```
minotogna: This should be visible only to when the menu after is visible

webapp/app/components/header/components/AppLinks.js
```diff
@@ -20,6 +20,7 @@ const AppLinks = ({ i18n }) => {
       className="app-header__menu-item">
       FRA Platform
```
minotogna: Use i18n here

webapp/app/components/header/components/AppLinks.js
sorja: Can you explain?

sorja: Currently looks like:
image

minotogna: Is the divider visible in statisticalfactsheets view? It should not

PR #148, opened 2020-03-13
Table 1a - added chart

minotogna:


common/objectUtils.js
```diff
@@ -0,0 +1,13 @@
+const R = require('ramda')
+
+const types = {
+  fn: 'Function',
+}
+
+const _isType = type => R.pipe(R.type, R.equals(type))
+
+const isFunction = _isType(types.fn)
```
sorja: typeof() ?

webapp/app/assessment/components/dataTable/table/row.js
```diff
@@ -23,7 +23,7 @@ const Row = props => {
     return null
   }
 
-  const renderer = type === 'custom' ? render : rowRenderers[type]``
+  const renderer = type === 'custom' ? render : rowRenderers[type]
```
sorja: maybe types.custom?

common/objectUtils.js
minotogna: Ramda has more specific object types (like array, null). I would prefer Ramda

sorja: ok

minotogna: https://ramdajs.com/docs/#type

webapp/app/assessment/components/dataTable/table/row.js
minotogna: removed completely. that was useless. we don't support custom type

PR #151, opened 2020-03-13
exact path login

### LINKED ISSUE #147: Reset password
Reset password functionality seems to be broken.
To be fixed in master branch

sorja:
Resolves #147

### DIFF
--- a/webapp/login/loginView.js
+++ b/webapp/login/loginView.js
@@ -44,7 +44,7 @@ const LoginView = () => {
             <ResetPasswordForm />
           </Route>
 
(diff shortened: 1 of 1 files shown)


webapp/login/loginView.js
```diff
@@ -44,7 +44,7 @@ const LoginView = () => {
             <ResetPasswordForm />
           </Route>
 
-          <Route path="/login/">
+          <Route exact path="/login/">
```
minotogna: add exact also to resetPassword.

PR #181, opened 2020-03-24
Refactor/forest area validator

### LINKED ISSUE #179: Implement forestAreaLessThanOrEqualToExtentOfForestValidator

minotogna:
Resolves #179

### DIFF
--- a/webapp/app/assessment/fra/sections/extentOfForest/extentOfForestValidatorState.js
+++ b/webapp/app/assessment/fra/sections/extentOfForest/extentOfForestValidatorState.js
@@ -1,5 +1,5 @@
 import * as R from 'ramda'
-import { abs, greaterThanOrEqualTo, lessThanOrEqualTo, sub } from '@common/bignumberUtils'
+import * as NumberUtils from '@common/bignumberUtils'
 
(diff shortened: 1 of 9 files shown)


webapp/app/assessment/fra/sections/extentOfForest/extentOfForestValidatorState.js
```diff
@@ -48,6 +48,19 @@ export const otherWoodedLandValidator = datum => state => {
...
 
+// ==== Common validator
+
+export const lessThanOrEqualToForestValidator = (year, value) => state => {
+  const forest = ExtentOfForestState.getForestByYear(year)(state)
+
+  if (R.isNil(value) || R.isNil(forest)) {
+    return true
+  }
+  const tolerance = -1
```
sorja: can we set this to default value for greaterThan?

minotogna: I noticed that. on one side yes, on the other making it explicitly serves as doc. 
Plus, this logic is used several times, i realized that in the end. I would suggest to add this method to numberUtils and in a separate issue. what do you think?

sorja: Agree

PR #213, opened 2020-04-14
table.js: display csv export if tableRef not null

### LINKED ISSUE #208: Table export
In same cases, table export doesn't work.
Downloaded CSV file is blank. (See AUS 1c, 1d)

sorja:
Resolves #208


webapp/app/assessment/components/dataTable/table/table.js
```diff
@@ -26,11 +26,12 @@ const Table = (props) => {
   const i18n = useI18n()
   const [printView] = usePrintView()
   const tableRef = useRef(null)
+  const displayTableExportButton = !secondary && !printView && tableRef.current != null
```
minotogna, line 29: It's not safe to check tableRef.current != null.
Use React.forwardRef

PR #230, opened 2020-04-22
Fix admin paths

### LINKED ISSUE #139: Update admin routes
- [x] Use /admin for admin routes
- [x] Add route to basePaths
- [x] Update link using basePaths in webapp/app/components/header/components/userInfo.js

sorja:
Resolves #139

### DIFF
--- a/webapp/main/basePaths.js
+++ b/webapp/main/basePaths.js
@@ -1,29 +1,24 @@
-// ==== application paths
-export const root = '/'
-export const login = '/login/'
-export const statisticalFactsheets = '/statisticalFactsheets/'
-export const country = '/:countryIso/'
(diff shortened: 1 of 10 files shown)


webapp/main/basePaths.js
```diff
@@ -8,20 +8,58 @@ export const assessmentPrint = `${country}print/:assessmentType/`
 export const assessmentPrintOnlyTables = `${country}print/:assessmentType/onlyTables/`
 export const odp = `${country}odp/:tab/`
 export const user = '/users/:userId/'
+export const admin = '/admin/'
+
+const pathFragments = {
```
minotogna, line 13: nice one. why not to use pathFragments in the routes above?

PR #322, opened 2020-06-10
Fix/enable trust proxy heorku

minotogna:


server/serverInit.js
```diff
@@ -36,15 +36,16 @@ module.exports = () => {
...
-// http://expressjs.com/en/guide/error-handling.html
-  app.use((err, req, res, next) => {
+  // Custom error-handling for handling custom exceptions and
+  // sending the uncaught errors as json instead of HTML
+  // http://expressjs.com/en/guide/error-handling.html
+  app.use((err, req, res) => {
     if (err) sendErr(res, err)
   })
 
+  app.enable('trust proxy')
```
minotogna, line 46: this is the only change. rest is formatting

PR #323, opened 2020-06-10
Feature/issue 254

### LINKED ISSUE #254: ExportView: Unit selection
Unit variables  in tables 1a, 1b, 1e, 1f, 3a, 3b, 4a, 4b,  5a, 5b, 6b is 1000ha and result value can be converted.

- [x] In webapp/app/assessment/components/section/sectionSpec add unitSpec.js (see below)
- [x] In the aforementioned tableSpecs add meta {unit:UnitSpec.units.haThousand}
- [x] In Table result, next to variable table, add dropdown unit selection if tableSpec has a unit defined (single selection)
- [x] convert values client side when unit changes

**unitSpec.js**
```
import * as NumberUtils from '@common/bignumberUtils'

export const units = {
  haThousand: '1000ha'
} 

export const factors = {
  [units.haThousand]: {
    ha: 1000,
    kmSq: 10,
    mileSq: 3.86102,
    acre1000: 2.47105,
    acre: 2471.05,
    haMillion: 0.001,
  },
}

export const convert = (value, unit, factor) => NumberUtils.formatNumber(NumberUtils.mul(value, factors[unit][factor]))
```

sorja:
Resolves #254

### DIFF
--- a/webapp/app/dataExport/components/resultsTable/resultsTable.js
+++ b/webapp/app/dataExport/components/resultsTable/resultsTable.js
@@ -5,7 +5,47 @@ import { useParams } from 'react-router'
 
 import { useI18n } from '@webapp/components/hooks'
 import ButtonTableExport from '@webapp/components/buttonTableExport'
-import { getValue, getI18nKey } from '../../utils/format'
+import * as SectionSpecs from '@webapp/app/assessment/components/section/sectionSpecs'
+import { UnitSpec } from '@webapp/app/assessment/components/section/sectionSpec'
+import { getValue, getI18nKey, valueConverted } from '../../utils/format'
+
+const ResultsTableTitle = (props) => {
+  const {
+    baseUnit,
+    selection: {
+      variable: { label, labelParam },
+    },
+    resultsLoading,
+    setSelected,
+  } = props
+
+  const i18n = useI18n()
+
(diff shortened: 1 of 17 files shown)


webapp/app/dataExport/components/resultsTable/resultsTable.js
```diff
@@ -15,24 +55,45 @@ const ResultsTable = (props) => {
...
+  }, [unit])
 
   useLayoutEffect(() => {
     setExportDisabled(resultsLoading)
   }, [resultsLoading])
 
   return (
     <div className="results-table">
-      <ButtonTableExport tableRef={tableRef} filename={`${assessmentType}-${section}`} disabled={exportDisabled} />
+      {!showTableExport && tableRef.current != null && (
```
minotogna, line 80: do not hide the button. setting exportDisabled to false and then true should be enough

webapp/app/dataExport/components/resultsTable/resultsTable.less
```diff
@@ -5,3 +5,7 @@
     text-align: right;
   }
 }
+
+.select-s {
```
minotogna, line 9: add this within results-table. otherwise it applies to all select-s

PR #324, opened 2020-06-10
Pan European table 1.4b metadata

apkretov:


webapp/app/assessment/panEuropean/sections/table_1_4b/sectionSpec.js
```diff
@@ -0,0 +1,53 @@
...
+        }),
+        SectionSpec.newColHeader({
+          [SectionSpec.KEYS_COL.labelKey]: 'panEuropean.carbonStockInHarvestedWoodProductsHWP.total_carbon_stock_in_hwp',
+        }),
+      ],
+    }),
+
+    // rows data
+    ...variables.flatMap((variable) =>
+      PanEuropean.years90_15.reverse().map((year) =>
```
minotogna, line 31: this method PanEuropean.years90_15.reverse() update the original array and might create problems when accessing it again.
create a variable years on top of the file like we did in table 1.1a
```const years = [...PanEuropean.years90_15].reverse()

apkretov: The variable has been created in the '1.4b years variable created' commit.

PR #339, opened 2020-06-17
columnSelect: select all

### LINKED ISSUE #327: Data export: add select all to Column

sorja:
Resolves #327


webapp/app/dataExport/components/columnSelect/columnSelect.less
```diff
@@ -0,0 +1,7 @@
+.export-column-select {
```
minotogna, line 1: don't we have a similar case in country selection?

sorja: yes

minotogna: can we generalize the style then?

sorja: yes

PR #407, opened 2020-07-07
statisticalFactsheets/table: display empty rows

### LINKED ISSUE #399: Statistical factsheets - Table missing rows
- [ ] Table rows must be always visible even with no data

Screenshot 2020-07-06 at 19 52 07

sorja:
Resolves #399

### DIFF
--- a/common/i18n/resources/en.js
+++ b/common/i18n/resources/en.js
@@ -63,6 +63,7 @@ module.exports.translation = {
     primaryDesignatedManagementObjective: {
       title: 'Primary Designated Management Objective (1990 - 2020)',
       rowName: '',
+      conservation: 'Conservation',
       multiple_use: 'Multiple use',
       other: 'Other',
       production: 'Production',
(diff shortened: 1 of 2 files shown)


common/i18n/resources/en.js
```diff
@@ -63,6 +63,7 @@ module.exports.translation = {
     primaryDesignatedManagementObjective: {
       title: 'Primary Designated Management Objective (1990 - 2020)',
       rowName: '',
+      conservation: 'Conservation',
```
minotogna, line 66: use designatedManagementObjective.biodiversityConservation and add it to all languages

common/i18n/resources/en.js
```diff
@@ -63,6 +63,7 @@ module.exports.translation = {
     primaryDesignatedManagementObjective: {
       title: 'Primary Designated Management Objective (1990 - 2020)',
       rowName: '',
+      conservation: 'Conservation of biodiversity',
```
minotogna, line 66: reference the key. do not duplicate it.

PR #493, opened 2020-08-29
Pan European table 4.8 metadata

apkretov:


apkretov:
Some unnecessary deletions appeared

PR #495, opened 2020-08-29
Pan European table 6.2 metadata

apkretov:


apkretov:
Some unnecessary deletions appeared

PR #496, opened 2020-08-29
Pan European table 6.3 metadata

apkretov:


apkretov:
Some unnecessary deletions appeared. Recreating the pull request was required.

PR #508, opened 2020-09-03
Dev/react refresh webpack plugin

sorja:
Adds hot reload for react frontend development

Note: The server to use during development is at localhost:**9000**. This will proxy any API requests to port 9001 and things should Just Work TM.


webpack.config.babel.js
```diff
@@ -46,7 +50,7 @@ const plugins = [
     __URL_STATISTICAL_FACTSHEETS__: JSON.stringify(process.env.URL_STATISTICAL_FACTSHEETS),
   }),
   new CleanWebpackPlugin(),
-]
+].filter(Boolean)
```
minotogna, line 53: can we avoid filter?
you can push to plugins if isDev

webpack.config.babel.js
```diff
@@ -70,14 +74,35 @@ const appConfig = {
...
+        use: [
+          // ... other loaders
+          {
+            loader: 'babel-loader',
+            options: {
+              // ... other options
+              plugins: [
+                // ... other plugins
+                isDevelopment && require.resolve('react-refresh/babel'),
+              ].filter(Boolean),
```
minotogna, line 102: avoid filter

minotogna (changes requested):

minotogna (approved):

PR #544, opened 2020-10-01
Feature/custom regions

sorja:
Contains changes:
new table regions
removed from country table columns: `{list_name_,full_name}{en,fr,es,ru}`, region
changed country repository queries to join regions as new field `regions[]`


common/i18n/resources/en/area.js
```diff
@@ -8,6 +8,8 @@ module.exports = {
   NA: { listName: 'North and Central America' },
   OC: { listName: 'Oceania' },
   SA: { listName: 'South America' },
+  EU27: { listName: 'EU-27' },
```
minotogna, line 11: add this to fr, es, ru as well

server/db/migration/migrations/sqls/20200930084905-custom-regions-up.sql
```diff
@@ -0,0 +1,56 @@
+
```
minotogna, line 1: place drop commands on top

server/db/migration/migrations/sqls/20200930084905-custom-regions-up.sql
```diff
@@ -0,0 +1,56 @@
...
+        ('EU27', 'LUX'),
+        ('EU27', 'MLT'),
+        ('EU27', 'NLD'),
+        ('EU27', 'POL'),
+        ('EU27', 'PRT'),
+        ('EU27', 'ROU'),
+        ('EU27', 'SVK'),
+        ('EU27', 'SVN'),
+        ('EU27', 'ESP'),
+        ('EU27', 'SWE');
```
minotogna, line 56: should we insert also the current county_region ?

webapp/app/country/countryState.js
```diff
@@ -10,6 +10,9 @@ const keys = {
   config: 'config',
   status: 'status',
   countries: 'countries',
+  regions: 'regions',
+  countryRegions: 'countryRegions',
+  regionCountries: 'regionCountries',
```
minotogna, line 15: what's the difference between countryRegions and regionCountries?

server/db/migration/migrations/sqls/20200930084905-custom-regions-up.sql
sorja: There is line 24:
` INSERT INTO country_region SELECT region_iso, country_iso FROM country; `
Is this what you meant?

webapp/app/country/countryState.js
sorja: countryRegions means in which regions a country belongs, for ex. Finland can be part of many different regions.
regionCountries is the other way around - what countries belong to a region, ex. EU contains FIN but also other countries

minotogna: but why not to fetch regions together with the countries? so in the UI we have only countries with a property regionCodes (array) ? + the regions themselves .
would make this things easier ?

server/country/countryRepository.js
```diff
@@ -74,27 +59,35 @@ const getAllCountries = (role, schemaName = 'public') => {
...
+    WHERE NOT (message IN ($1))
+    GROUP BY country_iso
+)
+SELECT c.country_iso,
+       c.pan_european,
+       a.type,
+       a.status,
+       a.desk_study,
+       fa.last_edited,
+       json_agg(cr.region_code) AS regions
```
minotogna, line 78: I would call the field region_codes

common/country/country.js
```diff
@@ -3,8 +3,6 @@ const R = require('ramda')
 const keys = {
   countryIso: 'countryIso',
   assessment: 'assessment',
-  fullName: 'fullName',
-  listName: 'listName',
   lastEdit: 'lastEdit',
   region: 'region',
   regionIso: 'regionIso',
```
minotogna, line 8: do we still regionIso prop ?

server/db/migration/migrations/sqls/20201013053954-drop-country-extra-columns-up.sql
```diff
@@ -0,0 +1,15 @@
...
+
+alter table country drop column full_name_es;
+
+alter table country drop column list_name_fr;
+
+alter table country drop column full_name_fr;
+
+alter table country drop column list_name_ru;
+
+alter table country drop column full_name_ru;
```
minotogna, line 15: should we drop also region_iso column ?

webapp/app/country/countryState.js
minotogna: why these 3 properties? can't you filter client side instead fetching all these?

minotogna (changes requested):
Also, please check if data export (from admin section and bulk download) includes regions.

server/db/migration/migrations/sqls/20201013053954-drop-country-extra-columns-up.sql
sorja: It is dropped in another migration 
server/db/migration/migrations/sqls/20201001104113-country-remove-region-region-iso-up.sql

webapp/app/country/countryState.js
minotogna: please answer this

minotogna (changes requested):

webapp/app/country/countryState.js
sorja: This was fixed, missed few keys.
Removing

minotogna (approved):

PR #551, opened 2020-10-08
Paneu/table4.4b

apkretov:


apkretov:
The 2.5 pull request contains the 4.4b one that first needed merging as well. The respective pull-request for 4.4b was still open when the 2.5 one came up.

PR #558, opened 2020-10-14
Pan European table 2.5 metadata

apkretov:


apkretov:
The 2.5 pull request contains the 4.4b one that first needed merging as well. The respective pull-request for 4.4b was still open when the 2.5 one came up.

Speaking about 2.5, an additional table below for the fields Other 1-3 is created (querying table 2.5oth) because they are text unlike the respective numeric data in table 2.5.

sorja (approved):

PR #564, opened 2020-10-19
Pan European table 6.9 metadata

apkretov:


apkretov:
This pull request is included into the next one - Pan European tables1.3a: Indentation.

Issue #542, opened 2020-09-29
Custom Regions


sorja:
- Create new table 'region'
- Create new table 'country_region'
- Remove region from 'country'

sorja:
Possibly reopened later

PR #568, opened 2020-10-20
Custom regions

sorja:
part of #542


webapp/app/user/userManagement/edit/countrySelectionModal.js
```diff
@@ -2,68 +2,90 @@ import React from 'react'
...
+
+  countries.forEach((country) => {
+    const { countryIso, regionCodes } = country
+    regionCodes.forEach((regionCode) => {
+      if (!Array.isArray(regionCountries[regionCode])) {
+        regionCountries[regionCode] = []
+      }
+      regionCountries[regionCode].push(countryIso)
+    })
+  })
```
minotogna, line 26: wouldn't it be better to use useRegions here?

minotogna (changes requested):

minotogna (approved):

webapp/app/user/userManagement/edit/countrySelectionModal.js
sorja: This will be later fixed when the upper component gets some care

PR #599, opened 2020-11-03
Store: Sort country.countries on language change

### LINKED ISSUE #594: CountrySelection: Group by user role
- [x] FRA Country selection doesn't use countries in store, therefore when switching language, countries are not sorted.
Use countries in store

- [x] Country selection dropdown must be updated as follow:
```
Global
____________
...Regions
____________
NATIONAL CORRESPONDANT
Country 1 
Country 2
____________
REVIEWER
Country 3 
Country 4
____________
.... All other countries user has no role for
```

- [x] Add column assessment to user_country_role table with value fra2020

- [x] Update store
  - Do not store user role under country/countries 
  - Assoc roles directly to user object (when fetched from server)
```
{
 role: {
  fra2020: {
    REVIEWER:[iso1,iso2]
  }
 }
}
```

sorja:
Resolves #594 

**Note:
`useOnUpdate` does not work**


webapp/store/country/hooks/useCountries.js
```diff
@@ -0,0 +1,27 @@
...
+export const sortCountries = (countries, i18n) => {
+  const _compareListName = Area.getCompareListName(i18n)
+  const _compareCountries = (country1, country2) => _compareListName(country1.countryIso, country2.countryIso)
+
+  return Object.fromEntries(
+    Object.entries(countries).map(([role, roleCountries]) => [role, [...roleCountries].sort(_compareCountries)])
+  )
+}
+
+export const useCountries = () => {
```
minotogna, line 17: dont we have this in App Store ?

sorja: This is different, see issue comment

minotogna: why don't you iterate over the countries we have in the store and get the role from those?

minotogna:
we will address this in a more stable manner

sorja:
Opened for discussion

PR #668, opened 2021-01-12
Variable & column names are updated in 4.2b

apkretov:


common/i18n/resources/en.js
```diff
@@ -1461,11 +1461,16 @@ The FRA team
       annualForestExpansionAndRegeneration1000ha: 'Annual forest expansion and regeneration (1000 ha)',
       expansionOfForestArea: 'Expansion of forest area',
       regenerationOfForestArea: 'Regeneration of forest area',
-      afforestation: 'Afforestation',
-      natural_expansion: 'Natural expansion',
-      natural_regeneration: 'Natural regeneration',
-      planting_and_seeding: 'Planting and seeding',
-      coppice: 'Coppice',
+      afforestation: 'Expansion: Afforestation',
```
minotogna, line 1464: maybe better to call this label expansion_afforestation so to avoid ugly _ at the end of a label key?

apkretov: The underscores have been replaced in all the similar cases.

minotogna (approved):

PR #676, opened 2021-01-14
Feature/region order 675

### LINKED ISSUE #675: Country selector: Regions order
Country selector must display regions as follow:

FRA regions on top Africa, Asia, Europe, North and Central America, Oceania and South America. Then adding a “line” like the one before the countries and then display the new regions, followed by a “line” and then the countries

sorja:
Resolves #675


webapp/store/app/hooks/index.js
```diff
@@ -45,3 +45,19 @@ export const useRegions = () => {
...
+  return Object.fromEntries(Object.entries(regionGroups).map(([key, regions]) => [key, sortRegions(regions, i18n)]))
+}
+
+export const useRegionGroups = () => {
+  const i18n = useI18n()
+  const dispatch = useDispatch()
+  const regionGroups = useSelector(AppState.getRegionGroups)
+
+  useOnUpdate(() => {
+    dispatch(AppActions.updateRegionGroups(sortRegionGroups(regionGroups, i18n)))
```
minotogna, line 59: region groups must be ordered by 'order' not by label

server/country/countryRepository.js
```diff
@@ -117,6 +117,15 @@ const getRegions = async () => {
   return camelize(result.rows)
 }
 
+const getRegionGroups = async () => {
+  // Exclude Atlantis from region groups
+  const query = `
+        SELECT region_group, ARRAY_AGG(region_code) as regions FROM region  WHERE region_code != 'AT' GROUP BY region_group
```
minotogna, line 123: wouldn't it be better to fetch from region_group table so that you get the 'order' as well?
regions have the regionGroupId for the association in the UI

minotogna (changes requested):

minotogna (changes requested):
please see my comments

webapp/store/app/hooks/index.js
sorja: This only sorts the array in the result set, does not change the order of the groups

minotogna (approved):

PR #683, opened 2021-01-21
typescript: init typescript

sorja:
PR Objective:
- [x] Setup initial configuration for typescript
- [x] Convert `webapp/main.js` to working `webapp/main.tsx`
- [x] Configure tslint
- ...testing?


webapp/main.tsx
```diff
@@ -3,15 +3,15 @@ import '@webapp/components/ckEditor/style.less'
...
+import * as React from 'react'
+import * as ReactDOM from 'react-dom'
 import { Provider } from 'react-redux'
 import { BrowserRouter } from 'react-router-dom'
 
 import PageRoutes from './pages/PageRoutes'
 import store from './main/store'
 
-function renderApp() {
+const App = () => {
```
minotogna, line 14: let's rename this to Main

tslint.json
```diff
@@ -0,0 +1,4 @@
+{
```
minotogna, line 1: if we have tslint, should we remove eslint?

sorja: We can keep both for now as the project has both plain javascript and will incrementally grow in typescript:
> When it comes to linting TypeScript code, there are two major linting options to choose from: TSLint and ESLint. TSLint is a linter that can only be used for TypeScript, while ESLint supports both JavaScript and TypeScript.

webapp/main.tsx
```diff
@@ -3,15 +3,15 @@ import '@webapp/components/ckEditor/style.less'
 
 import '@webapp/utils/polyfill/polyfill'
 
-import React from 'react'
-import ReactDOM from 'react-dom'
+import * as React from 'react'
+import * as ReactDOM from 'react-dom'
```
minotogna, line 7: do not use import *

minotogna (changes requested):

minotogna (approved):

PR #691, opened 2021-01-26
TypeScript: Redux store upgrade

sorja:
Major refactor:
- Fix redux store folder struct under `webapp/` (eg. move `webapp/app/{actions,reducers}` under `webapp/store/`)
- Migrate store to TypeScript
  - This is done incrementally


webapp/store/app/utils/index.js
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export const sortRegions = (regions, i18n) => {
+  const compareListName = Area.getCompareListName(i18n)
+  return [...regions].sort((r1, r2) => compareListName(r1.regionCode, r2.regionCode))
+}
+
+export const sortRegionGroups = (regionGroups) => {
+  return [...regionGroups].sort((rg1, rg2) => rg1.order > rg2.order)
+}
```
ramirobg94, line 16: how is going to grow this file?
maybe it is a good idea starting at this point split the utils folder by domain, ( on folder or file per regions other for countries, ) but maybe I need more context about the domain and the amount of use cases.

webapp/store/app/actions/index.js
```diff
@@ -1,9 +1,63 @@
...
+export const updateRegions = (regions) => ({ type: ActionTypes.updateRegions, regions })
+export const updateRegionGroups = (regionGroups) => ({ type: ActionTypes.updateRegionGroups, regionGroups })
+
+export const initApp = () => async (dispatch) => {
+  const lang = getRequestParam('lang')
+  try {
+    const getCountries = axios.get('/api/countries')
+    const getRegions = axios.get('/api/country/regions/')
+    const getRegionGroups = axios.get('/api/country/regionGroups/')
+    const getUserInfo = axios.get(`/api/loggedInUser/`)
```
ramirobg94, line 23: maybe the implementation of axios can be extracted to some wrapper of the api

webapp/app/country/actions.js
```diff
@@ -24,7 +24,7 @@ export const getCountryConfig = (countryIso) => async (dispatch) => {
 export const fetchCountryInitialData = (countryIso, assessmentType, printView, printOnlyTablesView) => (dispatch) => {
   dispatch(
     batchActions([
-      { type: appCountryIsoUpdate, countryIso, assessmentType, printView, printOnlyTablesView },
+      { type: ActionTypes.appCountryIsoUpdate, countryIso, assessmentType, printView, printOnlyTablesView },
```
ramirobg94, line 27: Nice :)

webapp/store/app/utils/index.js
ramirobg94: I feel scare about TS when we like to implement this functions as 
`export const sortCountries = ({countries, i18n}) => {`
vs
`export const sortCountries = (countries, i18n) => {`

doing the params in this way is easier to add new params and at the end more scalable and easier to mock

sorja: I needed to create this file - hopefully for now - because of circular import issue with eslint.
Regarding the second comment:
I think prototyping with TypeScript is slower - but after prototyping phase the project is easier to upkeep

webapp/store/app/actions/index.js
sorja: Yes, we had separate pull request/issue for creating a service, this though did not proceed for some reason

PR #701, opened 2021-01-27
Feature/pan european entry access

### LINKED ISSUE #700: PanEuropean entry access
- [x] Header: Fao logo left / 3 logos + language right 
- [x] Remove link pan European from home page
- [x] Loader: do not show toucan, only loading text when pan European is loading

minotogna:
Resolves #700


webapp/components/loading.js
```diff
@@ -1,12 +1,21 @@
...
-    <p className="subhead">
-      <strong>Loading... </strong>
-    </p>
-  </div>
-)
+import * as PanEuropean from '@common/assessment/panEuropean'
+
+const Loading = () => {
+  // Toucan is hidden in PanEuropean pages.
+  const { pathname } = window.location
```
sorja, line 7: Should we use
`import { useLocation } from 'react-router-dom'`?

minotogna: it's the same

sorja (approved):

PR #712, opened 2021-02-05
Typescript/master

sorja:
Convert most of the code to TypeScript

Heavily influenced by: [openforis/arena/pr:445](https://github.com/openforis/arena/pull/445)

- TODO:
  - [x] Remove ramda?
  - [x] Check i18n
  - [x] Check tests

Exceptions:
- DB migrations -- It seemed a bit involved to make them use TypeScript also
- Translation files under common/i18n/resources
- tools/

All the current tests pass, but I am not 100% sure if there are any regressions or not.


package.json
```diff
@@ -12,13 +12,14 @@
     "build:prod:report": "cross-env NODE_ENV=production BUILD_REPORT=true webpack --display-error-details -p --config webpack.config.babel.js",
     "build:prod": "cross-env NODE_ENV=production webpack --display-error-details -p --config webpack.config.babel.js",
     "clean": "rm -rf dist/",
-    "start": "node server/server.js",
-    "watch": "yarn build:dev-server & yarn watch:server",
+    "start": "ts-node server/server.ts",
+    "watch": "yarn build:dev-server & yarn watch:ts",
+    "watch:ts": "ts-node server/server.ts",
```
minotogna, line 17: do we need watch:ts? isn't it the same than start?

package.json
```diff
@@ -12,13 +12,14 @@
...
     "watch-debug": "yarn build:dev-server & yarn watch:server-debug",
-    "watch:server": "NODE_ENV=development nodemon -e js --watch server server/server.js",
+    "watch:server": "ts-node server/server.ts",
```
minotogna, line 19: isn't it the same than start? no watch option here?

minotogna (changes requested):

package.json
sorja: There is: ts-node-dev
but on macos it doesnt work flawless yet, so i didnt implement it

minotogna (approved):

PR #765, opened 2021-03-29
Mergify: configuration update

minotogna:
This change has been made by @minotogna from https://mergify.io config editor.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -6,3 +6,9 @@ pull_request_rules:
       merge:
         method: squash
         strict: smart+fasttrack
+  - name: automatic branch update
+    conditions:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -6,3 +6,10 @@ pull_request_rules:
       merge:
         method: squash
         strict: smart+fasttrack
+  - name: automatic update for PR marked as “Ready-to-Go“
```
sorja: Better to use "" instead of U+201C

sorja (changes requested):

.mergify.yml
minotogna: what?

sorja (approved):

PR #809, opened 2021-06-21
Add header mobile layout

minotogna:
- [X] Introduced spacing variables
- [X] Added header mobile layout


common/lang.ts
```diff
@@ -0,0 +1 @@
+export const LanguageCodes: Array<string> = ['ar', 'zh', 'en', 'fr', 'ru', 'es']
```
sorja, line 1: This exists currently in core/common/lang.ts

minotogna: core/common should not exists. we will remove that

minotogna: plus, that's not an array of available languages. will move LanguageCodes it under core

sorja (approved):

PR #818, opened 2021-06-23
db/ssl: {rejectUnauthorized}

sorja:


server/db/db.ts
```diff
@@ -30,7 +30,9 @@ const configCommon = {
   // Max number of clients in the pool
   max: 30,
   // Whether to use ssl connections
-  ssl: ProcessEnv.pgSsl,
+  ssl: {
+    rejectUnauthorized: ProcessEnv.rejectUnauthorized,
+  },
```
minotogna, line 35: this should be added only when ProcessEnv.pgSsl is true

minotogna (changes requested):

server/utils/processEnv.ts
```diff
@@ -27,4 +27,5 @@ export const ProcessEnv = {
   pgPort: Number(pgPort),
   pgDatabase,
   pgSsl: process.env.PGSSL === 'true',
+  rejectUnauthorized: process.env.rejectUnauthorized === 'true',
```
minotogna, line 30: we don't need this. it is always true when pgssl is true

server/db/db.ts
```diff
@@ -24,13 +24,19 @@ const pgp = pgPromise(initOptions)
...
 
-const configCommon = {
+const configCommon: any = {
   // How long a client is allowed to remain idle before being closed
   idleTimeoutMillis: 30000,
   // Max number of clients in the pool
   max: 30,
-  // Whether to use ssl connections
-  ssl: ProcessEnv.pgSsl,
+}
```
minotogna, line 32: cant you add here:
```
ssl:ProcessEnv.pgSsl?{rejectUnauthorized:true}:false
```

minotogna (changes requested):

minotogna (approved):

Issue #806, opened 2021-06-18
ODPService/repositories with transactions


minotogna:
- [ ] Introduce pg-promise and add DB object 
- [ ] Refactor ODPService so that its methods call other repositories (do not import a different repository from another repository)
- [ ] Transactions always start in Service and are passed down to repositories

PR #831, opened 2021-06-25
#2: ODP: Delete Draft

sorja:
Part of #806


server/service/odp/deleteDraft.ts
```diff
@@ -1,3 +1,20 @@
-import { OdpRepository } from '@server/repository/odp'
+import { User } from '@core/auth'
+import { BaseProtocol, DB } from '@server/db'
+import { getAndCheckOdpCountryId } from '@server/repository/odp/getAndCheckOdpCountryId'
+import { OdpClassRepository, OdpRepository } from '@server/repository'
 
-export const { deleteDraft } = OdpRepository
+export const deleteDraft = async (options: { odpId: number; user: User }, client: BaseProtocol = DB) => {
```
minotogna, line 6: methods should return Promise<whatever>
It can be done later if you wish so

sorja: Separate PR

minotogna (approved):

PR #877, opened 2021-08-09
#5: ODP : Return Types: Service/repository

sorja:
Part of #806

### DIFF
--- a/server/repository/odp/getOdp.ts
+++ b/server/repository/odp/getOdp.ts
@@ -5,9 +5,10 @@ import * as camelize from 'camelize'
 import * as db from '@server/db/db_deprecated'
 
 import { OdpClassRepository } from '@server/repository'
+import { ODP } from '@core/odp'
 import { getOdpVersionId } from './getOdpVersionId'
 
-export const getOdp = async (odpId: any, schemaName = 'public') => {
+export const getOdp = async (odpId: any, schemaName = 'public'): Promise<ODP> => {
   const versionId = await getOdpVersionId({ odpId }, db.pool, schemaName)
   const tableNameOdp = `${schemaName}.odp`
   const tableNameOdpVersion = `${schemaName}.odp_version`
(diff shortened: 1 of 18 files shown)


server/repository/odp/markAsActual.ts
```diff
@@ -2,7 +2,7 @@ import { insertAudit } from '@server/repository/audit/auditRepository'
 import { OdpClassRepository } from '@server/repository'
 import { getAndCheckOdpCountryId } from './getAndCheckOdpCountryId'
 
-export const markAsActual = async (client: any, odpId: any, user: any) => {
+export const markAsActual = async (client: any, odpId: any, user: any): Promise<null | [void, void]> => {
```
sorja, line 5: Not sure about this? Thoughts?

server/repository/odp/getOdp.ts
```diff
@@ -5,9 +5,10 @@ import * as camelize from 'camelize'
 import * as db from '@server/db/db_deprecated'
 
 import { OdpClassRepository } from '@server/repository'
+import { ODP } from '@core/odp'
 import { getOdpVersionId } from './getOdpVersionId'
 
-export const getOdp = async (odpId: any, schemaName = 'public') => {
+export const getOdp = async (odpId: any, schemaName = 'public'): Promise<ODP> => {
```
minotogna, line 11: no any

server/repository/odp/markAsActual.ts
minotogna: what about Promise<void> only?
and instead of returning null, you can return Promise.resolve() ?

sorja: Can't return only promise: `TS2314: Generic type 'Promise ' requires 1 type argument(s).`

sorja: Resolved in next pr

minotogna (approved):

PR #921, opened 2021-09-17
i18n refactor

sorja:
remove i18n from store


webapp/hooks/useI18n.ts
```diff
@@ -1,5 +1,15 @@
+import i18nInstance from '@common/i18n/i18nInstance'
 import { i18n } from 'i18next'
-
+import { useEffect, useState } from 'react'
 import { useAppSelector } from '@webapp/store'
 
-export default (): i18n => useAppSelector((state) => state.app.i18n)
+export default (): Partial<i18n> => {
+  const [instance, setInstance] = useState(i18nInstance.getInstance())
```
minotogna, line 7: why adding to local components state the i18n instance ?
have you checked react i18 next? https://react.i18next.com

webapp/store/page/dataExport/reducer.ts
```diff
@@ -16,7 +14,7 @@ const initialState: DataExportState = {
 }
 
 const actionHandlers = {
-  [AppActionTypes.appCountryIsoUpdate]: () => ({ ...initialState }),
+  'app/updateCountryIso': () => ({ ...initialState }),
```
minotogna, line 17: can we avoid hardcoded strings?

sorja: In dataExport refactor will be handled, app throws error if `<action>.type` is used

sorja:
Closed in favor of react-i18next

PR #924, opened 2021-09-22
Rtk/app react i18n hook

sorja:


webapp/store/user/reducer.ts
```diff
@@ -3,12 +3,11 @@ import { exportReducer } from '@webapp/utils/reduxUtils'
...
 
-import ActionTypes from '@webapp/store/app/actions/actionTypes'
-
 const actionHandlers = {
-  [ActionTypes.appInitDone]: (state: any, { userInfo }: any) => UserState.assocUserInfo(userInfo)(state),
+  // TODO: waiting refactor
+  'app/init/fulfilled': (state: any, { userInfo }: any) => UserState.assocUserInfo(userInfo)(state),
 
-  [ActionTypes.appI18nUpdate]: (state: any, { i18n }: any) => UserState.assocUserInfoLang(i18n.language)(state),
+  'app/switchLanguage/fulfilled': (state: any, { i18n }: any) => UserState.assocUserInfoLang(i18n.language)(state),
```
minotogna, line 10: dont we have these actions to avoid the string?

sorja: Later (I think) we can use `<appAction>.type`, but currently it will throw an error 
(I think because the reducer is not a slice or done through rtk)

minotogna (approved):

PR #958, opened 2021-10-19
initial commit for odp/remove

sorja:
`$curl localhost:9001/api/originalDatapoint/813 -X DELETE`
```{"odp":{"id":"813","country_iso":"FIN","year":null,"data_source_additional_comments":null,"data_source_methods":null,"data_source_references":null,"description":null,"national_classes":null,"id_legacy":null}}```

### DIFF
--- a/server/repository/originalDataPoint/remove.ts
+++ b/server/repository/originalDataPoint/remove.ts
@@ -1,6 +1,8 @@
 import { BaseProtocol, DB } from '@server/db'
+import { Objects } from '@core/utils'
+import { ODP } from '@core/odp'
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: server/repository/originalDataPoint/remove.ts [hunk 1]
```diff
@@ -0,0 +1,14 @@
+import { BaseProtocol, DB } from '@server/db'
+
+export const remove = async (props: { id: string }, client: BaseProtocol = DB): Promise<string> => {
+  const { id } = props
+
+  const odp = await client.one(
+    `
+        delete from original_data_point where id = ($1) returning *;
+    `,
+    [id]
```
minotogna, line 10: Doesn't this return a promise of ODP?
Shouldn't it be camelized?!

PR #959, opened 2021-10-19
initial commit for odp/get

sorja:
` curl localhost:9001/api/originalDataPoint/812`
```{"odp":{"id":"812","country_iso":"FIN","year":null,"data_source_additional_comments":null,"data_source_methods":null,"data_source_references":null,"description":null,"national_classes":null,"id_legacy":null}}```

### DIFF
--- a/server/repository/originalDataPoint/remove.ts
+++ b/server/repository/originalDataPoint/remove.ts
@@ -1,6 +1,8 @@
 import { BaseProtocol, DB } from '@server/db'
+import { Objects } from '@core/utils'
+import { ODP } from '@core/odp'
 
-export const remove = async (props: { id: string }, client: BaseProtocol = DB): Promise<string> => {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: server/repository/originalDataPoint/get.ts [hunk 1]
```diff
@@ -0,0 +1,15 @@
...
+  const { id } = props
+
+  const odp = await client.one(
+    `
+        select * from original_data_point where id = $1
+    `,
+    [id]
+  )
+
+  return odp
```
minotogna, line 14: Should it be camelized?!

minotogna (changes requested):

server/repository/originalDataPoint/get.ts
sorja: Jep, missed & fixed

PR #964, opened 2021-10-20
Add audit log for dp

sorja:
- [x] Create
- [ ] Update
- [ ] Delete

### DIFF
--- a/server/api/originalDataPoint/create.ts
+++ b/server/api/originalDataPoint/create.ts
@@ -3,6 +3,7 @@ import { Requests } from '@server/utils'
 import { ApiEndPoint } from '@common/api/endpoint'
 import { OriginalDataPointService } from '@server/service/originalDataPoint'
 import { ApiAuthMiddleware } from '@server/api/middleware'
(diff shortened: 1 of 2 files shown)


server/service/originalDataPoint/remove.ts
```diff
@@ -1,8 +1,15 @@
...
 
-export const remove = async (props: { id: string }): Promise<ODP> => {
-  const { id } = props
-  const odp = await OriginalDataPointRepository.remove({ id })
-  return odp
+export const remove = async (props: { id: string; user: User }, client: BaseProtocol = DB): Promise<ODP> => {
+  const { id, user } = props
+  return client.tx(async (t) => {
+    const removedOdp = await OriginalDataPointRepository.remove({ id }, t)
+    await AuditRepository.insertAudit(t, user.id, 'deleteOdp', countryIso, 'odp', { removedOdp })
```
minotogna, line 12: the object in the target should always be called odp

minotogna (changes requested):

PR #989, opened 2021-11-02
initial commit for previous odp api

sorja:
server side implementation for previous odp

`curl localhost:9001/api/originalDataPoints/233/previous`

```
{
  "odp": {
    "id": "232",
    "year": 2000,
    "countryIso": "FIN",
    "dataSourceMethods": [
      "nationalForestInventory"
    ],
    "dataSourceAdditionalComments": "NFI9",
    "dataSourceReferences": null,
    "description": null,
    "nationalClasses": [
      ...
    ]
  }
}
```


minotogna (approved):
The double query in the repository is a bit inefficient , but it should be ok with the amount of data we have

PR #1002, opened 2021-11-04
Add initial version of Geo api

xbexbex:
Initial version of Google Earth Engine API. Contains getCountryLimits-function mainly for test purposes.
* The API route is nested as this endpoint is going to be huge.
* Requires a [Google service account configuration](https://cloud.google.com/iam/docs/creating-managing-service-account-keys) to work. 
    * Currently stored as stringified JSON in the .env file. Perhaps there is a better way to manage secret files on Heroku?
* The only reliable way to import the earthengine library seems to be via require. No support for types :(


minotogna:
The process.env is the right way to read environment variables locally and from Heroku , so what you used is already the better way.

Have you checked if there's @types/google earth engine module?
Are you sure we can't use import with import * neither ?

xbexbex:
> Have you checked if there's @types/google earth engine module? Are you sure we can't use import with import * neither ?

Yeah, couldn't find @types and I doubt there are any. Earthengine JS is a relatively unpopular library with only 400 weekly downloads.

But the * import works! Made a commit for that.

minotogna (approved):

PR #1011, opened 2021-11-12
Meta/users

sorja:


core/meta/user.ts
```diff
@@ -0,0 +1,43 @@
...
+  id: number
+  profilePictureFilename?: string
+  name: string
+  status: UserStatus
+  profilePictureFile?: string
+  position?: string
+  email: string
+}
+
+export enum Provider {
```
minotogna, line 19: call this AuthProvider

package.json
```diff
@@ -163,6 +166,7 @@
     "express-fileupload": "^1.2.1",
     "express-session": "^1.17.2",
     "i18next": "^20.3.2",
+    "jest": "^27.3.1",
```
minotogna, line 169: does jest go to devDependencies ?

server/api/assessment/export.ts
```diff
@@ -1,7 +1,7 @@
 import { Express, Response, Request } from 'express'
 import JSZip from 'jszip'
 import { ApiAuthMiddleware } from '@server/api/middleware'
-import * as ExportService from '@server/service/assessment/exportService'
+import * as ExportService from '@server/service/assessment/_legacy/exportService'
```
minotogna, line 4: why renaming the service? review might take longer

server/db/migration/migrations/sqls/20211111110342-move-to-legacy-schema-up.sql
```diff
@@ -0,0 +1,7 @@
+create schema _legacy;
+
+alter table public._legacy_assessment set schema _legacy;
+alter table public._legacy_fra_user set schema _legacy;
+alter table public._legacy_odp set schema _legacy;
+alter table public._legacy_odp_class set schema _legacy;
+alter table public._legacy_odp_version set schema _legacy;
```
minotogna, line 7: can we remove _legacy prefix from these tables now that they are in _legacy schema?

server/db/migration/migrations/sqls/20211111111648-create-users-tables-up.sql
```diff
@@ -0,0 +1,44 @@
...
+            primary key,
+    profile_picture_filename varchar(1024),
+    name                     varchar(1024),
+    status                   users_status default 'invitationPending' not null,
+    profile_picture_file     bytea,
+    position                 varchar(1024),
+    email                    varchar(255) not null
+);
+
+create type provider as enum ('local', 'google');
```
minotogna, line 18: call this auth_provider

server/db/migration/migrations/sqls/20211111111648-create-users-tables-up.sql
```diff
@@ -0,0 +1,44 @@
...
+            primary key,
+    user_id                 bigint references users (id) not null,
+    provider                provider not null,
+    props                   jsonb
+);
+
+create table users_invitation
+(
+    uuid                    uuid primary key,
+    invited_at              timestamptz,
```
minotogna, line 33: default now

server/db/migration/migrations/sqls/20211111111648-create-users-tables-up.sql
```diff
@@ -0,0 +1,44 @@
...
+    accepted_at             timestamptz,
+    user_id                 bigint references users (id) not null
+);
+
+create table users_reset_password
+(
+    uuid                    uuid primary key,
+    changed_at              timestamptz,
+    created_at              timestamptz,
```
minotogna, line 42: default now

server/db/migration/migrations/sqls/20211111111648-create-users-tables-up.sql
```diff
@@ -0,0 +1,44 @@
...
+    user_id                 bigint references users (id) not null
+);
```
minotogna, line 44: for timestamptz isn't better to use what we use in other migrations as well?

server/service/user/create.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const create = async (
+  props: { user: Pick<User, 'name' | 'email'> },
+  client: BaseProtocol = DB
+): Promise<User> => {
+  const { user } = props
+
+  return client.tx(async (t) => {
+    // insert entry in ActivityLog: Does not exist in public - should?
+    return UserRepository.create({ user }, t)
```
minotogna, line 13: we should only record when the user has been invited to a specific assessment. maybe insert an entry only when the user has been invited?

testIntegration/001-user-create.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+test('ExpectedUser === User', async () => {
+  const userParams = {
+    email: 'test@fra-platform.com',
+    name: 'Test User',
+  }
+
+  const user = await UserService.create({
+    user: userParams,
+  })
```
minotogna, line 12: when creating a user, we should also insert a user provider.
insert for now a provider 'local' with props {password:'test'}.
do not insert 'test', but the hashed version

minotogna (changes requested):

server/db/migration/migrations/sqls/20211111111648-create-users-tables-up.sql
sorja: Let's use timestamptz from now on

```
The big difference between these two data types is that timestamptz includes a timezone offset while timestamp does not. So it is important to decide for the timestamp you are storing whether the timezone is important to keep or not. 
```
https://kb.objectrocket.com/postgresql/postgresql-timestamp-vs-timestamptz-616

server/service/user/create.ts
sorja: Ok, will be handled separately

server/api/assessment/export.ts
sorja: Just moved under _legacy, no renaming!

minotogna (approved):

PR #1019, opened 2021-11-18
add settings, test, service, repository

### LINKED ISSUE #1012: Settings table
add settings table to public schema with column default_assessment_id references assessment id

test: after insert assessment, set default assessment to the one created

sorja:
Resolves #1012


server/service/assessment/create.ts
```diff
@@ -28,4 +29,12 @@ export const create = async (
     )
     return createdAssessment
   })
+
+  await SettingsRepository.update({
+    settings: {
+      defaultAssessmentId: ret.id,
+    },
+  })
```
minotogna, line 37: this shouldn't happen at assessment creation. it's something we have to trigger programmatically

server/service/settings/update.ts
```diff
@@ -0,0 +1,17 @@
...
+export const update = async (
+  props: {
+    settings: Settings
+  },
+  client: BaseProtocol = DB
+): Promise<Settings> => {
+  const { settings } = props
+
+  return client.tx(async (t) => {
+    // TODO: Add activity log entry (public schema?)
```
minotogna, line 14: nope. we assume for now is always fra. or we can think about this later? we can remove the todo?

test/integration/settings/default-assessment.ts
```diff
@@ -0,0 +1,22 @@
...
+import { SettingsService } from '@server/service'
+
+export default () =>
+  test('Expect default assessment to be updated', async () => {
+    const assessmentParams = {
+      props: {
+        name: 'fra',
+        cycles: ['1'],
+      },
+    }
```
minotogna, line 11: can we have a global common assessmentParams mock?

test/integration/settings/default-assessment.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    const assessment = await AssessmentService.read({
+      assessment: assessmentParams,
+    })
+
```
minotogna, line 16: here the test, should first set fra as default and then verify it's correct

minotogna (changes requested):

minotogna (approved):

Issue #1015, opened 2021-11-17
Home refactor 


minotogna:
Depends on #1013

Home page non logged in user 

refactor store such as :
assessment: {
 assessment:Assessment,
countries:Array<Country>
regions, regionGroups.. .....
}

app is ready is assessment state is not undefined 

on app init action, call backend endpoint /api/init 

- implement country selection

PR #1048, opened 2021-12-03
update useTheme()

sorja:
part of #1015

- add basepaths
- add hook for assessmentName
- add hooks for paths
- fix useTheme (use new hooks)


client/store/assessment/hooks/index.ts
```diff
@@ -0,0 +1,5 @@
+import { useAppSelector } from '@client/store'
+import { AssessmentName } from '@core/meta/assessment'
+
+export const useAssessmentPropsName = (): AssessmentName =>
```
minotogna, line 4: better useAssessment and that's it.
then in the component you use assessment.props.name

minotogna (changes requested):

minotogna (approved):

PR #1055, opened 2021-12-07
initial commit for userstate

sorja:
part of #1015 

- adds rootstate entry for user
- adds hook `userUser`
- adds empty state

### DIFF
--- a/client/store/RootState.ts
+++ b/client/store/RootState.ts
@@ -1,7 +1,7 @@
-import { User } from '@core/meta/user'
 import { AssessmentState } from './assessment/stateType'
+import { UserState } from './user/stateType'
 
 export type RootState = {
   assessment: AssessmentState
-  user: User
+  user: UserState
 }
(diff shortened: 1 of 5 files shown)


client/store/user/hooks/index.ts
```diff
@@ -0,0 +1,4 @@
+import { useAppSelector } from '@client/store'
+import { User } from '@core/meta/user'
+
+export const useUser = (): User => useAppSelector((state) => state.user)
```
minotogna, line 4: this should return user | undefined

client/store/RootState.ts
```diff
@@ -1,5 +1,7 @@
+import { User } from '@core/meta/user'
 import { AssessmentState } from './assessment/stateType'
 
 export type RootState = {
   assessment: AssessmentState
+  user: User
```
minotogna, line 6: user might be undefined

sorja: hmm, also assessment before initialization... should we define it as partial then?

minotogna (approved):

PR #1067, opened 2021-12-10
Meta/area selector

sorja:
part of #1015 

image
image


client/hooks/useCountryIso.ts
```diff
@@ -0,0 +1,7 @@
+import { useParams } from 'react-router-dom'
+
+export const useCountryIso = () => {
+  // TODO: get this from store if not in url
```
minotogna, line 4: Actually let's leave it as it is from url params. We'll see later on ho le it goes

minotogna (approved):

PR #1073, opened 2021-12-13
add mailservice

sorja:
adds mail service


server/controller/user/invite.ts
```diff
@@ -12,6 +13,7 @@ export const invite = async (
     email: string
     roleName: RoleName
     user: User
+    url: string // application url
```
sorja, line 16: how should we handle this?

sorja:
how should we handle testing?

minotogna (approved):

PR #1074, opened 2021-12-14
save language

sorja:
part of #1015


client/utils/url.ts
```diff
@@ -0,0 +1,9 @@
+const getRequestParam = (name: any) => {
```
minotogna, line 1: is name a string?

minotogna (approved):

Issue #1035, opened 2021-11-26
User login


minotogna:
- [x] Frontend Login with google
- [x] Backend Login with google
- [x] Frontend Login with email/password
- [ ] IT Login with email/password
- [x] Backend Login with  email/password

1. userA logins with his password success 
2. userA logins with wrong password fails

PR #1089, opened 2022-01-03
api endpoint: authorize user [merge first]

sorja:
part of #1035 
add `/api/authorize` which returns user

### DIFF
--- a/common/api/endpoint/ApiEndPoint.ts
+++ b/common/api/endpoint/ApiEndPoint.ts
@@ -7,6 +7,7 @@ export const ApiEndPoint = {
   },
   // TODO: Remove deprecated routes (still in use in _legacy)
   Auth: {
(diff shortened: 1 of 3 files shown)


common/api/endpoint/ApiEndPoint.ts
```diff
@@ -7,6 +7,7 @@ export const ApiEndPoint = {
   },
   // TODO: Remove deprecated routes (still in use in _legacy)
   Auth: {
+    authorize: () => apiPath('authorize'),
```
minotogna, line 10: can't we get the user on the init endpoint ?

sorja: We need separate endpoint for authorization
We can also return the user on init

We need later on handle token refresh

Issue #1092, opened 2022-01-05
Navigation refactor


minotogna:
Add route /:countryIso/assessments/:assessmentName/:cycleUuid/home/overview/

Add Country home page 

Add left navigation menu (sections)

PR #1102, opened 2022-01-12
Navigation

sorja:
part of #1092 
image

Todo
- [x] Navigation state (`store.navigation` vs `store.ui.navigation`)
- [x] NavigationHeader
- [ ] ReviewStatusMarker
- [ ] Header/Status
- [x] Expand on url match


client/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
```
minotogna, line 2: why not to add all old code and comment it out?

minotogna (changes requested):

client/components/Navigation/Navigation.tsx
sorja: Added files for what I need, but can add all the code too

minotogna: yes please, so we know what we'll need to do

minotogna (changes requested):
add Navigation state + expand on url match + navigation header (comment things will implement later)

client/store/rootReducer.ts
```diff
@@ -1,9 +1,11 @@
 import AssessmentSlice from './assessment/slice'
 import LoginSlice from './login/slice'
 import UserSlice from './user/slice'
+import UiSlice from './ui/slice'
 
 export default {
   assessment: AssessmentSlice,
   login: LoginSlice,
   user: UserSlice,
+  ui: UiSlice,
```
minotogna, line 10: store should be ui/navigation.

export actions/hooks/reducer etc from store/ui/navigation 
e.g. ```import {NavigationActions} from '@store/ui/navigation'```

minotogna (changes requested):

minotogna (approved):

Issue #1105, opened 2022-01-13
ODP visualization


minotogna:
- [x] move original data point table to _legacy schema
- [x] add odp tables to assessment cycle schema - exactly the same
- [x] in migrate-fra (after migrateBasicTablesData) add migrateOdps and import all old odps as they are + props:{cycles:[_**assessment cycle uuids**_]} 
- [x] add page /:countryIso/:assessmentName/:cycleName/originalDataPoint/:odpId with rendered odp from server

PR #1110, opened 2022-01-18
render path for odp

sorja:
part of #1105 

ss: Left development, right production

image

todo
- [x] fix deepscan
- [x] fix url(s)
- [x] render components
- [x] weird issue where year jumps next to references


client/basePaths.ts
```diff
@@ -18,29 +19,37 @@ export const BasePaths = {
...
     resetPassword: () => `/login/resetPassword`,
     invitation: () => '/login/invitation',
   },
+  OriginalDataPoint: {
+    root: (
+      countryIso: CountryIso | defaults.countryIso = defaults.countryIso,
+      assessmentName: AssessmentName | defaults.assessmentName = defaults.assessmentName,
+      cycleName = defaults.cycleName,
+      odpId = ':odpId'
+    ) => _generate(countryIso, assessmentName, cycleName, 'originalDataPoint', odpId),
```
sorja, line 51: should we prefix with assessment like other assessment paths?
if not, should we prefix instead with originalDataPoint?

currently looks like:
http://localhost:9000/FIN/fra/2025/originalDataPoint/72

minotogna: it should follow assessment convention. then yes.
and they should be: /:countryIso/assessments/:name/cycles/:cycleName/originalDataPoint/:odpId

minotogna: call the key, one, not root

sorja: I think we need to revisit the paths in general
For client side, I think 'root' makes more sense?
We also have other paths under OriginalDataPoint (section, tab)

Renamed for now

minotogna (changes requested):
convert less to sass

minotogna (approved):

PR #1111, opened 2022-01-18
reset password backend

rfontanarosa:


server/controller/user/sendResetPasswordRequest.ts
```diff
@@ -0,0 +1,16 @@
+import { BaseProtocol, DB } from '@server/db'
+import { User, UserResetPassword } from '@meta/user'
+import { UserResetPasswordRepository } from '@server/repository'
+
+export const sendResetPasswordRequest = async (
```
sorja, line 5: Following convention of others, this (file,method) should be called `create`

sorja: or if it's under `user` maybe `UserController.createResetPassword`

server/repository/userProvider/update.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Objects } from '@core/utils'
+import { User, UserAuthProvider } from '@meta/user'
+
+export const update = async (
+  props: {
+    user: Pick<User, 'id'>
+    password: string
+  },
+  client: BaseProtocol = DB
+): Promise<UserAuthProvider<{ password: string }>> => {
```
sorja, line 11: I wonder if the return type is correct?
Shouldn't `returning *` return the insterted row from `public.users_auth_provider`?

sorja:
can user have more than one reset password requests?

sorja (approved):

PR #1121, opened 2022-01-25
restored google login and other fixes

rfontanarosa:


client/pages/Login/LoginForm/LoginForm.tsx
```diff
@@ -52,7 +48,25 @@ const LoginForm: React.FC<Props> = (props: Props) => {
     }
   }
 
-  return (
+  return !loginLocal ? (
```
sorja, line 51: can we use convention

`if (loginLocal) return <div...>`

later this convention of `?:` will cause readability issues

sorja (changes requested):

server/api/auth/localStrategyVerifyCallback.ts
```diff
@@ -14,8 +14,8 @@ export const localStrategyVerifyCallback = async (req: Request, email: string, p
     } else if (Objects.isEmpty(password.trim())) {
       sendErr('login.noEmptyPassword')
     } else {
-      const invitationUuid = String(req.query.invitationUuid)
-      if (invitationUuid) {
+      const invitationUuid = req.query?.invitationUuid as string
+      if (invitationUuid !== undefined) {
```
sorja, line 18: i think this doesnt need `!== undefined`

sorja:
add redirect to front page on succesful login

sorja (approved):

PR #1135, opened 2022-02-02
Meta/server authorizer

sorja:


meta/user/authorizer.ts
```diff
@@ -0,0 +1,98 @@
...
+    return true
+
+  if (Users.isCollaborator(user, countryIso)) {
+    const userRole: any = Users.getCountryRole(user, countryIso)
+
+    const userSections = userRole.props?.sections ?? {}
+    if (!userSections) return true
+    if (userSections === 'none') return false
+    if (userSections === 'all') return true
+    if (userSections.sectionUuid === table.tableSectionId) return true
```
minotogna, line 85: 1. you need section uuid. pass section instead of table ?
2. the correct check is: return userSections[sectionUuid]===true

sorja (commented):
👍

minotogna (approved):

PR #1139, opened 2022-02-04
Meta/default cycle

### LINKED ISSUE #1132: Default assessment cycle
- [x] assessment props: add defaultCycleUuid string
- [x] data migration: set fist cycle as default cycle 
- [x] country select, if cycle is not in url, use default cycle

sorja:
resolves #1132


client/components/CountrySelect/CountryList/CountryListRow/CountryListRow.tsx
```diff
@@ -7,16 +7,19 @@ import { useTranslation } from 'react-i18next'
 import { BasePaths } from '@client/basePaths'
 import { Dates } from '@client/utils'
 import { CountryIso, Global, RegionCode } from '@meta/area'
+import { useAssessment } from '@client/store/assessment'
+import { useCycle } from '@client/hooks/useCycleName'
```
minotogna, line 11: let's export from @client/hooks ?

minotogna: actually that hooks should go under @client/store/assessment

minotogna (approved):

PR #1155, opened 2022-02-15
Persist node Value

minotogna:
- Added CycleDataController.persistNodeValue with calculated variables dependency graph updates
- Added NodeRepository create/update/getOneOrNone
- Added RowRepository getOne
- Added unique index node table on (country_iso, col_uuid, row_uuid) 
- Added Api new convention example (User.getInvitation)

Note: Tests have been condacted locally (as demonstrated)
a proper test hasn't been commited since we first need to implement the metadata creation part.
It will be done soonish - Added to the Backlog


server/controller/cycleData/persistNodeValue/expressionEvaluator/binary.ts
```diff
@@ -0,0 +1,40 @@
...
+      throw new Error(`Boolean ${operator} not supported`)
+    }
+
+    const leftResult = this.evaluator.evaluateNode(left, this.context)
+    const rightResult = this.evaluator.evaluateNode(right, this.context)
+    const result = fn(leftResult, rightResult)
+    if (result) return result.toString()
+    // Arithmetic operators will always return nulls for any non-numeric inputs.
+    // see Numbers.applyOperator
+    return result ?? null
```
sorja, line 38: It can never return result because of the previous if

server/controller/cycleData/persistNodeValue/expressionEvaluator/binary.ts
```diff
@@ -0,0 +1,40 @@
...
+  // '^':   (a, b) => a ^ b,
+  // '&':   (a, b) => a & b,
+  // Don't allow shifts either:
+  // '<<':  (a, b) => a << b,
+  // '>>':  (a, b) => a >> b,
+  // '>>>': (a, b) => a >>> b,
+}
+
+export class Binary extends ExpressionNodeEvaluator<Context, BinaryExpression> {
+  evaluate(expressionNode: BinaryExpression): any {
```
sorja, line 24: return type string | null?

server/repository/node/create.ts
```diff
@@ -0,0 +1,37 @@
...
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+  countryIso: string
+  colName: string
+  tableName: string
+  variableName: string
+  value: NodeValue
+}
+
```
sorja, line 14: Common props?

sorja (changes requested):

server/controller/cycleData/persistNodeValue/expressionEvaluator/binary.ts
minotogna: that's the super method

server/repository/node/create.ts
minotogna: not necessarily now

sorja (approved):

PR #1168, opened 2022-02-28
Update port env vars

xbexbex:
My ports 9000 and 9001 are mostly blocked, so...

* Added a port variable for the web frontend in .env and webpack config
* Added the existing backend port variable in webpack config


webpack.config.babel.js
```diff
@@ -78,11 +78,11 @@ const appConfig = {
       {
         // Proxy all server-served routes:
         context: ['/auth', '/img', '/css', '/ckeditor', '/video', '/api', '/definitions'],
-        target: 'http://localhost:9001',
+        target: process.env.APP_URI,
       },
     ],
     compress: false,
-    port: 9000,
+    port: process.env.WEB_APP_PORT,
```
minotogna, line 85: Can we have a default to 9000 so that we don’t necessarily need to add it to .env ?

xbexbex: Makes sense, added it.

minotogna (approved):

PR #1180, opened 2022-03-04
removed toaster default delay and fixed a bug

### LINKED ISSUE #1179: remove toaster 5s default duration
by default a toast is removed automatically after 5 seconds. remove this behaviour.

rfontanarosa:
fix #1179


minotogna (changes requested):
Only error should not be closed automatically 

Other should have default 5 seconds

minotogna (approved):

PR #1189, opened 2022-03-06
Geo/initial map

xbexbex:
* Added Google Maps JS
* Added a placeholder geo map


client/pages/Geo/Geo.tsx
```diff
@@ -1,8 +1,36 @@
...
 
 const Geo: React.FC = () => {
-  return <div />
+  const [isRenderAllowed, setIsRenderAllowed] = useState(false)
+
+  // makes the map fill remaining screen space nicely
+  useEffect(() => {
+    const main = document.getElementById('main')
+    const mainDefaultVal = main.style.minHeight
+    main.style.minHeight = '100vh'
```
minotogna, line 15: Would adding this to main css selector break something? If not , I would add it there , not in JavaScript

minotogna (approved):

Issue #1080, opened 2021-12-15
missing translations


sorja:
- common.dashboard: 'Dashboard',
- login.resetPassword: 'Reset password',
- login.noShortPassword: ‘Password too short.’,
- login.alreadyLinked: 'The selected google account is already linked to another user',
- invitationMessage: 'You have been invited to {{assessment}} {{cycle}} as {{userRole}}',
- acceptInvitation: 'Accept Invitation',
- acceptInvitationWithGoogle: 'Accept Invitation with Google',
- alreadyAcceptedInvitation: 'This invitation has already been accepted',
- logoutSuccessful: 'You have successfully logged out',
- review.loading: 'Loading',
- userManagement.userAdded: '{{email}} has been added'
- userManagement.userModified: '{{user}} has been modified'
- userManagement.invitationEmailSent: 'An invitation email has been sent'
- userManagement.invitationDeleted: 'The invitation has been deleted'
- userManagement.personalInfoRequired: 'Please complete your personal information before continuing',
- userManagement.invitationEmail.textMessage
- userManagement.invitationEmail.htmlMessage
- page.assessmentSection.dataTableHasErrors: 'Data table contains errors. Click on the red cell to see details.'
- editUser.title: 'Appellation'
- editUser.surname: 'Surname'
- editUser.activated: 'Activated'
- editUser.status: 'Status'
- editUser.demoteToUser: 'Are you sure you want to remove Admin privileges?',
- editUser.promoteToAdmin: 'Are you sure you want to grant Admin privileges?',
- fileUploaded: 'File successfully uploaded'
- fileDeleted: 'File successfully deleted'
- userManagement.editPermissions: 'Edit permissions'
- userManagement.permissions: 'Permissions'
- userManagement. permissionNames.tableData: 'Table Data'
- userManagement. permissionNames.descriptions: 'Descriptions'
- invitationExpired: 'This invitation has expired',
- noMatchingLocalUser:
    "We couldn't find any user matching these credentials.\nMake sure you have a valid FRA account or reset your password in case you forget or lose it.",
- noMatchingProvider:
    "We couldn't find any user matching these credentials.\nMake sure you have a valid FRA account or try another authentication method.",
  invitationProvidersRegistered: 'You already have an account registered with {{authProviderNames}}',
- login.noActiveAccount: 'The account is not active. Make sure you have accepted any pending invitation',

- fra.js
```javascript
  extentOfForest: {
    remainingLandArea: 'Remaining land area',
[...]

PR #1201, opened 2022-03-14
restored google login correct behaviour

### LINKED ISSUE #1194: Fix Google invitation/login
see server/repository/user/read.ts

// TODO: pass props: Pick<User, 'email'>  | Pick<User, 'id'> | {emailGoogle:string}
// TODO: rename method getOneOrNone - check if you can use getOne
// TODO: insert in user auth provider google email during invitation
// TODO: when passing emailGoogle, left join with users_auth_provider and where condition becomes props->>'email' = $1

rfontanarosa:
fix #1194


i18n/resources/en/login.js
```diff
@@ -41,6 +41,7 @@ module.exports = {
   returnHome: 'To return to the home page click',
   returnHomeClick: 'here.',
 
+  alreadyAcceptedInvitation: 'This invitation has already been accepted',
```
sorja, line 44: Was this added to #1080

server/api/auth/googleStrategyVerifyCallback.ts
```diff
@@ -19,9 +19,18 @@ export const googleStrategyVerifyCallback = async (
     const invitationUuid = req.query.state as string
     if (invitationUuid) {
       const { user: invitedUser, userRole } = await UserController.readByInvitation({ invitationUuid })
+      if (invitedUser && invitedUser.status !== 'active') {
```
sorja, line 22: invitedUser?.status !==

client/store/login/actions/localLogin.ts
```diff
@@ -1,22 +1,19 @@
...
-    history: RouteComponentProps['history']
   }
->('login/local', async ({ email, password, invitationUuid, history }) => {
+>('login/local', async ({ email, password, invitationUuid }) => {
   await axios.post(`${ApiEndPoint.Auth.Login.local()}${invitationUuid ? `?invitationUuid=${invitationUuid}` : ''}`, {
     email,
     password,
   })
-  history.push(BasePaths.Root())
+  window.location.reload()
```
minotogna, line 18: why reload?

server/controller/user/index.ts
```diff
@@ -11,7 +11,7 @@ import { changePassword } from './changePassword'
 export const UserController = {
   create,
   update,
-  read,
+  getOne,
```
minotogna, line 14: this could just be:
```
getOne: UserRepository.getOne
```

server/repository/user/getOne.ts
```diff
@@ -6,29 +6,25 @@ const fields: Array<string> = ['lang', 'id', 'name', 'status', 'position', 'emai
...
 
-// TODO: pass props: Pick<User, 'email'>  | Pick<User, 'id'> | {emailGoogle:string}
-// TODO: rename method getOneOrNone - check if you can use getOne
-// TODO: insert in user auth provider google email during invitation
-// TODO: when passing emailGoogle, left join with users_auth_provider and where condition becomes props->>'email' = $1
-
-export const read = async (
-  props: { user: Pick<User, 'email'> } | { user: Pick<User, 'id'> },
+export const getOne = async (
+  props: { user: Pick<User, 'email'> | Pick<User, 'id'>; emailGoogle?: string },
```
minotogna, line 10: this should be Pick<User, 'email'> | Pick<User, 'id'> | {emailGoogle:string}

minotogna (changes requested):

client/store/login/actions/localLogin.ts
rfontanarosa: Because after the login a refresh page was necessary to let the system know that the user is logged in, this is just a quick fix. I'm going to create an issue.

server/repository/user/getOne.ts
rfontanarosa: the problem is that using Pick the descrutrurized prop user (const { user } = props) should be a User type object

client/store/login/actions/localLogin.ts
minotogna: we have a single page webapp. reload is not necessary.
update page with user information instead

server/repository/user/getOne.ts
minotogna: then do not pass user at all!
{id:number} | {email:string} | {emailGoogle:string}

rfontanarosa: but this way is impossible to destructurize too. if props is defined this way is impossible to do props.id or { id } = props

rfontanarosa: maybe this way { id?: number, email?: string, emailGoogle?: boolean } and check if one of them is defined at the beginning

rfontanarosa: @sorja what do you think about it?

rfontanarosa: ok

rfontanarosa: https://github.com/openforis/fra-platform/pull/1201/commits/92212f47baf3399ff6b1636a14b93b41349250f6

client/store/login/actions/localLogin.ts
rfontanarosa: done

client/store/login/actions/localLogin.ts
```diff
@@ -13,10 +14,17 @@ export const localLogin = createAsyncThunk<
...
-  })
-  history.push(BasePaths.Root())
+>('login/post/local', async ({ email, password, invitationUuid, history }, { dispatch }) => {
+  const response = await axios.post(
+    `${ApiEndPoint.Auth.Login.local()}${invitationUuid ? `?invitationUuid=${invitationUuid}` : ''}`,
+    {
+      email,
+      password,
+    }
+  )
```
sorja, line 24: can we use axios params instead of concatenation?
Here is example: https://axios-http.com/docs/example

PR #1218, opened 2022-03-28
Geo/sepal auth fix

sorja:
Fixes issue with `resourceCacheControl` messing with Sepal proxy


sorja:
> maybe you can remove server/api/geo/index and its commented references

left for original author to either remove or use it :)

PR #1223, opened 2022-03-29
Geo/sepal proxy

xbexbex:
* Add a proxy with authentication header for request to Sepal
   * Uses the library http-proxy-middleware
* Authentication is basic HTTP auth that is created from Sepal username and password

### DIFF
--- a/client/components/ReviewIndicator/ReviewIndicator.scss
+++ b/client/components/ReviewIndicator/ReviewIndicator.scss
@@ -1,0 +1,18 @@
+@import 'client/style/partials';
+
+.review-indicator {
+  background: none;
+  border: none;
+  display: flex;
+  justify-content: center;
+  width: 32px;
+  svg {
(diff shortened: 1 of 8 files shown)


.env.template
```diff
@@ -31,3 +31,6 @@ FRA_MAIL_PASSWORD=bar
 
 #JWT
 TOKEN_SECRET=195...95
+
+# Basic http authorization token. Created from Sepal username and password.
+SEPAL_TOKEN="xyz1234"
```
minotogna, line 36: is this a static token ever?

xbexbex: Yes, it's just base64 encoding of `username:password`. Of course we can also put Sepal username and password in .env and base64 encode within code. That probably makes it clearer.

minotogna: please, share it separately ;)

minotogna (approved):

Issue #1292, opened 2022-04-27
Navigation sidebar review indicators


minotogna:
- [x] BE: when user opens a topic or sends a message, set last_open_time to now() in message_topic_user table
- [x] review store: ui/review/status|summary
- [x] hook: useReviewSummaryStatus(props{sectionid.subSectionid?}) 'read'|'unread'|undefined
- [x] FE: when user if logged in, fetch review summary: Show opened (green) / unread (orange) status indicators 
- [x] when user post a message, update navigation status indicator with web socket (resend all data to client)
- [x] when user resolves a topic, update navigation status indicator with web socket
(resend all data to client)

_initial sql query_
```sql

with r as (
    select r.uuid as row_uuid, s.id as sub_section_id, s.parent_id
    from assessment_fra.row r
             left join assessment_fra."table" t
                       on t.id = r.table_id
             left join assessment_fra.table_section ts
                       on ts.id = t.table_section_id
             left join assessment_fra.section s
                       on s.id = ts.section_id
    where r.props -> 'cycles' ? 'b6595ae8-c407-4ec2-9c89-c78b410513de'
),
     m as (
         select r.row_uuid,
                r.sub_section_id,
                r.parent_id,
                m.topic_id,
                m.created_time as last_message_created_time,
                row_number() over (
                    partition by r.row_uuid, r.sub_section_id
                    )          as row_number
         from r
                  left join assessment_fra_2020.message_topic mt
                            on r.row_uuid::varchar = mt.key
                  left join assessment_fra_2020.message m
                            on m.topic_id = mt.id
         where mt.status = 'opened'
           and not m.deleted)
select m.sub_section_id,
       m.parent_id,
       m.row_uuid,
       m.last_message_created_time,
       u.last_open_time,
       mt.status
from m
         left join assessment_fra_2020.message_topic_user u
                   on u.topic_id = m.topic_id
                       and u.user_id = 28457
         left join assessment_fra_2020.message_topic mt on mt.id = m.topic_id
where m.row_number = 1
;

```

PR #1322, opened 2022-05-12
status indicator sidebar BE

rfontanarosa:
part of #1292


server/controller/messageCenter/getTopic.ts
```diff
@@ -15,7 +18,21 @@ export const getTopic = async (
...
+    client
+  )
+
+  if (topic && user) {
+    const topicUser = await MessageTopicUserRepository.getOneOrNone({ assessment, cycle, topic, user })
+    if (topicUser) {
+      await MessageTopicUserRepository.update({ assessment, cycle, topic, user })
+    } else {
+      await MessageTopicUserRepository.create({ assessment, cycle, topic, user })
+    }
```
minotogna, line 34: this code is duplicated.
create a file in the same folder that export a function called updateTopicReadTime
this updateTopicReadTime will be used also in addMessage

minotogna (changes requested):

server/controller/messageCenter/getTopic.ts
rfontanarosa: ok

server/controller/messageCenter/getTopic.ts
```diff
@@ -15,7 +19,16 @@ export const getTopic = async (
...
-  const { countryIso, assessment, cycle, key, includeMessages = true } = props
+  const { countryIso, assessment, cycle, key, includeMessages = true, user } = props
+
+  const topic = await MessageTopicRepository.getOneOrNone(
+    { countryIso, assessment, cycle, key, includeMessages },
+    client
+  )
+
+  if (topic && user) {
+    await MessageCenterController.updateTopicReadTime({ assessment, cycle, topic, user }, client)
```
minotogna, line 30: you are already within messagecentercontroller . update these in all calls
```
import {updateTopicReadTime} from './updateTopicReadTime'
and then 
await updateTopicReadTime....
```

minotogna (changes requested):

minotogna (approved):

PR #1343, opened 2022-05-25
Country Select with assessment information and CSV download

### LINKED ISSUE #1098: CountrySelection
- Implement CountryListDownload (requires more data per country)
- Add missing data for countries (country informations: lastEdit, fra2020Assessment)
- Currently no linking between regions and countries => filtering for region does not show correct country
- FE: update country roles on assessment or cycle or user change

minotogna:
Resolves #1098

Screenshot 2022-05-25 at 11 33 42

### DIFF
--- a/client/store/assessment/hooks/index.ts
+++ b/client/store/assessment/hooks/index.ts
@@ -1,10 +1,12 @@
-import { useAppSelector } from '@client/store'
-import { Assessment, Section, SubSection } from '@meta/assessment'
-import { Country, CountryIso, RegionGroup } from '@meta/area'
-
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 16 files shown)


client/store/assessment/hooks/index.ts
```diff
@@ -38,6 +40,15 @@ export const useCountries = (): Array<Country> => {
   return Object.values(countries).sort((c1, c2) => compareListName(c1.countryIso, c2.countryIso))
 }
 
+export const useCountry = (countryIso?: CountryIso): Country =>
```
sorja, line 43: countryIso shouldn't be optional here?

sorja (approved):

PR #1344, opened 2022-05-25
Assessment Home blank

sorja:


client/pages/Assessment/Assessment.tsx
```diff
@@ -30,7 +31,7 @@ const Assessment: React.FC = () => {
 
       <MessageCenter />
       <Switch>
-        {/* <Route path={BasePaths.assessmentHome} component={AssessmentHome} /> */}
+        <Route path={BasePaths.Assessment.root()} component={AssessmentHome} />
```
minotogna, line 34: this should redirect to BasePaths.Assessment.root()/home 
and add that route

minotogna (changes requested):

PR #1408, opened 2022-06-28
Fix forest ownership calc formula

### LINKED ISSUE #1389: Dashboard differences vs PROD
Screenshot 2022-06-16 at 17 08 43
Screenshot 2022-06-16 at 17 13 54
Screenshot 2022-06-16 at 17 13 29

minotogna:
resolves #1389

### DIFF
--- a/tools/dataMigration/fraSpecs.ts
+++ b/tools/dataMigration/fraSpecs.ts
@@ -5262,8 +5262,9 @@ export const FraSpecs: Record<string, SectionSpec> = {
                 variableName: 'other_or_unknown',
                 migration: {
                   colNames: ['1990', '2000', '2010', '2015'],
-                  calcFormula:
-                    'extentOfForest.forestArea - (forestOwnership.private_ownership || 0) - (forestOwnership.public_ownership || 0)',
+                  calcFormula: `forestOwnership.private_ownership && forestOwnership.public_ownership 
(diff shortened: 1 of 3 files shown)


tools/dataMigration/fraSpecs.ts
```diff
@@ -5262,8 +5262,9 @@ export const FraSpecs: Record<string, SectionSpec> = {
                 variableName: 'other_or_unknown',
                 migration: {
                   colNames: ['1990', '2000', '2010', '2015'],
-                  calcFormula:
-                    'extentOfForest.forestArea - (forestOwnership.private_ownership || 0) - (forestOwnership.public_ownership || 0)',
+                  calcFormula: `forestOwnership.private_ownership && forestOwnership.public_ownership
```
rfontanarosa, line 5265: - || 0 is no more necessary
- in this case backtick is ok but not necessary, we can keep it

'forestOwnership.private_ownership && forestOwnership.public_ownership 
    ? extentOfForest.forestArea - forestOwnership.private_ownership - forestOwnership.public_ownership 
    : null'

rfontanarosa (changes requested):

tools/dataMigration/fraSpecs.ts
minotogna: yes, because one of the two can still be null, it's safer that way

minotogna: backtick is used to put the formula in multiple lines, for readability

rfontanarosa (approved):

PR #1442, opened 2022-08-08
table sections edit permissions

### LINKED ISSUE #1428: UserList edit tables access

rfontanarosa:
closes #1428

Schermata 2022-08-16 alle 07 55 36


client/pages/Assessment/Assessment.scss
```diff
@@ -62,8 +62,8 @@
   grid-area: content;
   height: 100%;
   padding: $spacing-xs;
-  overflow-x: auto;
-  overflow-y: hidden;
+  // overflow-x: auto;
+  // overflow-y: hidden;
```
sorja, line 66: Should we remove unused code comments or is it left for reference?

minotogna: this must be restored

client/store/userManagement/actions/updateTableAccess.ts
```diff
@@ -0,0 +1,16 @@
+import { ApiEndPoint } from '@common/api/endpoint'
+import { createAsyncThunk } from '@reduxjs/toolkit'
+import axios from 'axios'
+
+export const updateTableAccess = createAsyncThunk<
```
minotogna, line 5: this is in reality sections. let's call this updateSectionAuth

common/api/endpoint/ApiEndPoint.ts
```diff
@@ -117,6 +117,7 @@ export const ApiEndPoint = {
     removeInvitation: (uuid = ':uuid') => apiPath('users', 'invitation', 'remove', uuid),
     sendInvitationEmail: (uuid = ':uuid') => apiPath('users', 'invitation', 'sendEmail', uuid),
     getProfilePicture: (id = ':id') => apiPath('users', 'profilePicture', id),
+    tableAccess: () => apiPath('users', 'tableAccess'),
```
minotogna, line 120: let's call this sectionAuth

server/api/user/updateTableAccess.ts
```diff
@@ -0,0 +1,19 @@
+import { Request, Response } from 'express'
+
+import { UserController } from '@server/controller/user'
+import { Requests } from '@server/utils'
+
+export const updateTableAccess = async (req: Request, res: Response) => {
```
minotogna, line 6: updateSectionAuth

server/repository/public/userRole/updateSectionsProp.ts
```diff
@@ -0,0 +1,23 @@
+import { Objects } from '@core/utils'
+
+import { CollaboratorProps, RoleName, UserRole } from '@meta/user'
+
+import { BaseProtocol, DB } from '@server/db'
+
+export const updateSectionsProp = async (
```
minotogna, line 7: updateSectionsAuth

minotogna (changes requested):

server/repository/public/user/getMany.ts
```diff
@@ -29,5 +29,5 @@ export const getMany = async (
     `,
       [countryIso, assessment.id, cycle.uuid]
     )
-    .then(Objects.camelize)
+    .then((data) => Objects.camelize(data, { skip: ['sections'] }))
```
minotogna, line 32: let's use spread

minotogna (changes requested):

client/components/UserList/CollaboratorMultiSelect/CollaboratorMultiSelect.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  useEffect(() => {
+    document.addEventListener('click', handleClickOutside)
+
+    return () => {
+      document.removeEventListener('click', handleClickOutside)
+    }
+  }, [])
+
+  useEffect(() => {
```
minotogna, line 95: use the useOnUpdate hook

meta/user/user.ts
```diff
@@ -15,6 +15,6 @@ export interface User {
   position?: string
   profilePictureFile?: string
   profilePictureFilename?: string
-  roles: Array<UserRole<RoleName>>
+  roles: Array<UserRole<RoleName> | Collaborator>
```
minotogna, line 18: isn't Collaborator a UserRole ? you can remove this

rfontanarosa: it's a specific type of userrole without void as props. Without this OR, the application won't work. This happened because the query returns an array of users and from the type User it's impossible to specify the type of the roles which is UserRole<RoleName, void> by default.

client/components/UserList/CollaboratorMultiSelect/CollaboratorMultiSelect.tsx
rfontanarosa: wow, I didn't know about this hook. It does exactly what I implemented following an example found on the Internet.

meta/user/user.ts
sorja: I am running in similar issue(s) where I have to specifically define roles to be Collaborator in some parts, otherwise I will get typescript errors as mentioned above

I have changed the void to undefined though, as void should we used as function return type

rfontanarosa: you're right, with undefined we can cast the props field without typescript errors

minotogna (approved):

PR #1526, opened 2022-09-19
Add support for Select all for regions in country modal desktop

### LINKED ISSUE #1517: User Edit: Country Role select/unselect all by region
Add check select all / unselect all next to each region name that allows selecting or unselecting all countries in that region

OJ: When assigning countries to a user (reviewer, collaborator) ad possibility to select all countries for a region or select all regions for the world

sorja:
Resolves #1517 
- [x] Desktop
- [x] Mobile

Note:
Feature hides Select All/Unselect All when all countries of region are selected under another role

image

image

### DIFF
--- a/src/client/styles/checkbox.scss
+++ b/src/client/styles/checkbox.scss
@@ -1,7 +1,7 @@
 @import 'src/client/styles/config';
 
 .fra-checkbox {
-  width: 14px;
+  min-width: 14px;
   height: 14px;
   border-radius: 2px;
   border: 1px solid $ui-border-dark;
--- a/src/client/styles/checkbox.scss
+++ b/src/client/styles/checkbox.scss
@@ -1,7 +1,7 @@
 @import 'src/client/styles/config';
 
 .fra-checkbox {
-  min-width: 14px;
+  width: 14px;
   height: 14px;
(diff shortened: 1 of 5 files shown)


rfontanarosa (approved):

minotogna (changes requested):
Select / unselect all should be visible always and select / deselect enabled countries .

Plus , can the button be located on the right of the region label ?

minotogna (changes requested):
move  select/all unselect all next to the label with 4/8 px distance

sorja:
> move select/all unselect all next to the label with 4/8 px distance

image

src/client/styles/checkbox.scss
```diff
@@ -1,7 +1,7 @@
 @import 'src/client/styles/config';
 
 .fra-checkbox {
-  width: 14px;
+  min-width: 14px;
```
minotogna: why should be this bigger than 14px?

sorja: There was weird issue where the checkbox would clip to 12px with `width` but would work with `min-width`; seems ot be fixed now so reverted back

minotogna (approved):

PR #1529, opened 2022-09-19
Add tools: Compare bulkdownload csv

sorja:
Initial version to compare csv

## Notes:
- Use `fast-csv` to async read csv to array of objects for easier handling

## TODO
- [ ] Better output on error/mismatch
- [ ] Better comparison for rows (?)
- [x] Expect data not be exact
  - [x] ex. Region row sorted alphabetically vs other

### DIFF
--- a/tools/compareBulkdownloadCSV.ts
+++ b/tools/compareBulkdownloadCSV.ts
@@ -1,0 +1,94 @@
+import * as fs from 'fs'
+import * as fastCsv from 'fast-csv'
+
+const normalize = (string: string) =>
+  string
+    .split(',')
+    .map((x) => x.trim())
+    .sort()
+    .join(', ')
+
+const _readCSV = async (fileName: string): Promise<any> => {
+  const options = {
+    objectMode: true,
+    delimiter: ',',
+    // quote: null,
+    headers: true,
+    renameHeaders: false,
+    ignoreEmpty: true,
+  }
+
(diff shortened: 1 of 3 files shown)


tools/compareBulkdownloadCSV.ts
```diff
@@ -0,0 +1,94 @@
...
+    objectMode: true,
+    delimiter: ',',
+    // quote: null,
+    headers: true,
+    renameHeaders: false,
+    ignoreEmpty: true,
+  }
+
+  return new Promise((resolve, reject) => {
+    const data: any = []
```
minotogna, line 22: Is it possible to remove any ?

minotogna:
let's the script do the following:
- compare all files within locally downloaded bulk download (prod vs local)
- collect all diffs (per file) in an array
- write in a csv files the diffs collected

> code sample
```typescript
import * as JSON2CSV from 'json2csv'
import * as fs from 'fs/promises'

type Diff ={ countryIso,rowNo,tableName, valueProd,valueLocal,...meaningful props}
....
const diffs:Array<Diff> = []
...
diffs.push({...})
...
const csv = await JSON2CSV.parseAsync(diffs)
await fs.writeFile('./diffsFraYears.csv', csv)
...

```

sorja:
Closing this branch for now, working on separate branch

PR #1532, opened 2022-09-20
fix unpublished cycle

### LINKED ISSUE #1504: Publicly accessing non published cycle error

rfontanarosa:
close #1504

### DIFF
--- a/src/client/pages/Assessment/Assessment.tsx
+++ b/src/client/pages/Assessment/Assessment.tsx
@@ -5,14 +5,16 @@ import { Navigate, Route, Routes, useParams } from 'react-router-dom'
 import { Areas } from '@meta/area'
 import { AssessmentName } from '@meta/assessment'
 import { Sockets } from '@meta/socket'
+import { Authorizer } from '@meta/user'
 
 import { useAppDispatch } from '@client/store'
-import { AssessmentActions, useAssessment } from '@client/store/assessment'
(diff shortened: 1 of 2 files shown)


src/client/pages/Assessment/Assessment.tsx
```diff
@@ -69,6 +72,8 @@ const Assessment: React.FC = () => {
 
   if (!assessment) return null
 
+  if (!Authorizer.canView({ countryIso, assessment, cycle, user })) window.location.href = BasePaths.Root()
```
sorja, line 75: `history.replace` ?

minotogna: actually yes, but get the history from react-router

rfontanarosa: it's complicated. we can'y use useHistory at the same level of <Routes> definition (it is null).

minotogna: useNavigate ? https://reactrouter.com/en/main/hooks/use-navigate

sorja: just as info,
using `window.location.href` creates a new ajax query and leaves the SPA
while using the react-router stays inside SPA and does not create unnecessary ajax calls

minotogna: yes, that's why I'm suggesting that

sorja (approved):

minotogna (approved):

PR #1546, opened 2022-09-23
activity log invitation delete

### LINKED ISSUE #1537: ActivityLog invitationRemove
persist ActivityLog.invitationRemove when invitation is revoked

rfontanarosa:
closes #1537

### DIFF
--- a/src/server/controller/user/removeInvitation.ts
+++ b/src/server/controller/user/removeInvitation.ts
@@ -1,6 +1,6 @@
 import { CountryIso } from '@meta/area'
 import { ActivityLogMessage, Assessment, Cycle } from '@meta/assessment'
-import { User } from '@meta/user'
+import { RoleName, User, UserRole } from '@meta/user'
 
 import { BaseProtocol, DB } from '@server/db'
 import { ActivityLogRepository } from '@server/repository/assessment/activityLog'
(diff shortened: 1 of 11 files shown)


sorja (approved):

src/server/controller/user/removeInvitation.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  return client.tx(async (t) => {
+    const userRole = await UserRoleRepository.remove({ invitationUuid }, t)
+
+    const invitedUser = await UserRepository.getOne({ id: userRole.userId })
+
+    await ActivityLogRepository.insertActivityLog(
+      {
+        activityLog: {
+          target: { user: invitedUser.name, role: userRole.role },
```
minotogna: target should be: {userId:invitedUser.id, role...}

minotogna (changes requested):

src/server/controller/user/removeInvitation.ts
rfontanarosa: in the old fra_audit they were stored that way... the user information is used by the Activity section.

Schermata 2022-09-23 alle 15 18 24

rfontanarosa: I can add it but not in all the migrated entries.

minotogna (approved):

Issue #1305, opened 2022-05-10
Navigation should stay open when changing sections


sorja:

sorja:
https://fra-platform-incubator.herokuapp.com/FIN/assessments/fra/2020/specificForestCategories/
![image](https://user-images.githubusercontent.com/5508251/168090124-88d385be-c3b4-4f52-a260-aa1382bde699.png)

PR #1547, opened 2022-09-24
Persist metadata, only on cycle change, reset

### LINKED ISSUE #1538: Cache in store section metadata and data broken
When user is navigating within same cycle, section metadata should be cached and fetched only if not in store.

When user is navigating within same cycle, and same country, table data should be cached and fetched anyway every time the section page is loaded.

sorja:
resolves #1538
resolves of #1305 

- [x] separate fetching to hook
- [x] in Assessment.tsx if cycle changes (root/first level check) reset metadata
- [x]  On Mobile, when opening a page in a section, navigation menu should not be visible by default #1305
- [x] On all, when opening a page in a section, current section should be opened and highlighted #1305


minotogna:
What about data ? Does data gets cached and fetched anyway ?

sorja:
> What about data ? Does data gets cached and fetched anyway ?

Data is updated every time you change section

sorja:
Hopefully this video recording explains it a bit better


https://user-images.githubusercontent.com/5508251/192088772-31ea98c4-ba7d-40f0-8122-953d8f8440cf.mov

minotogna:
> > What about data ? Does data gets cached and fetched anyway ?
> 
> 
> 
> Data is updated every time you change section

 Actually, thinking it through better , app will be consistent and faster if:
- data is **always** fetched every time section page is accessed 
- Data is **not** removed from store when navigating in a different section or a different country (cached like metadata)  .
- data is reset when navigating to a different cycle (like metadata)

Does this make sense to you as well ?

sorja:
> > > What about data ? Does data gets cached and fetched anyway ?
> > 
> > 
> > Data is updated every time you change section
> 
> Actually, thinking it through better , app will be consistent and faster if:
> 
> * data is **always** fetched every time section page is accessed
> * Data is **not** removed from store when navigating in a different section or a different country (cached like metadata)  .
> * data is reset when navigating to a different cycle (like metadata)
> 
> Does this make sense to you as well ?

Makes sense, so we cache and fetch the data
I'll see how it works out

src/client/components/Navigation/NavigationMobile/NavigationMobile.tsx
```diff
@@ -13,6 +13,11 @@ const NavigationMobile: React.FC = () => {
   const dispatch = useAppDispatch()
   const navigationVisible = useNavigationVisible()
 
+  // Hide navigation on first mount
+  useEffect(() => {
+    dispatch(NavigationActions.updateNavigationVisible(false))
```
sorja, line 18: #1305

src/client/components/Navigation/NavAssessment/Section/NavigationSection.tsx
```diff
@@ -46,13 +47,18 @@ const NavigationSection: React.FC<Props> = (props) => {
   // // On mount check whether the location matches a child path
   useEffect(() => {
     const match = section.subSections.find((subsection) => {
-      const path = BasePaths.Assessment.section(countryIso, assessmentName, subsection.props.name)
+      const path = ClientRoutes.Assessment.Section.getLink({
```
sorja, line 50: On all, when opening a page in a section, current section should be opened and highlighted https://github.com/openforis/fra-platform/issues/1305

src/client/components/Navigation/NavigationDesktop/NavigationDesktop.tsx
```diff
@@ -13,11 +15,17 @@ import NavAssessment from '../NavAssessment'
...
   const { i18n } = useTranslation()
+  const dispatch = useAppDispatch()
   const countryIso = useCountryIso()
   const assessment = useAssessment()
   const cycle = useCycle()
   const assessmentName = assessment.props.name
 
+  // Show navigation on first mount (ex. returing from Mobile view)
+  useEffect(() => {
+    dispatch(NavigationActions.updateNavigationVisible(true))
```
sorja, line 26: #1305

minotogna: Is it visible only on desktop ?

minotogna (approved):

src/client/components/Navigation/NavigationDesktop/NavigationDesktop.tsx
sorja: This configuration now means that, when you change the browser size to mobile, the navigation is hidden, when returning to desktop size it is shown again.
On both, you can still toggle the navigation

rfontanarosa (approved):

PR #1548, opened 2022-09-25
Hide printView section if no data, always show secondary tables

### LINKED ISSUE #202: PrintViewOnlyTables - render tables/sections
- [x] Do not render section when all tables within the section are empty
- [x] Do not render **primary** tables when data is empty (always render secondary tables)

sorja:
Resolves #202

Todo
- [ ] Verify functionality as expected

Screenshot 2022-09-25 at 16-19-04 FRA platform

### DIFF
--- a/src/client/pages/AssessmentSection/AssessmentSection.tsx
+++ b/src/client/pages/AssessmentSection/AssessmentSection.tsx
@@ -7,6 +7,7 @@ import { AssessmentName, AssessmentNames } from '@meta/assessment'
 
 import { useAssessmentSection } from '@client/store/assessment'
 import { useTableSections } from '@client/store/pages/assessmentSection'
+import { useIsSectionDataEmpty } from '@client/store/pages/assessmentSection/hooks'
 import { useCanEditSection } from '@client/store/user'
 import { useIsPrint } from '@client/hooks/useIsPath'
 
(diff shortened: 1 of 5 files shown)


src/client/pages/AssessmentSection/DataTable/DataTable.tsx
```diff
@@ -80,7 +80,8 @@ const DataTable: React.FC<Props> = (props) => {
     countryIso,
   })
 
-  if (dataEmpty && onlyTables) {
+  // Always show secondary tables - unless whole section empty (handled in parent)
+  if (dataEmpty && onlyTables && !table.props.secondary) {
```
sorja, line 84: Only hide the non secondary tables as in #202 mentioned

src/client/pages/AssessmentSection/AssessmentSection.tsx
```diff
@@ -33,6 +34,12 @@ const AssessmentSection: React.FC<Props> = (props: Props) => {
 
   const { anchor, showTitle, descriptions, name: sectionName } = assessmentSection?.props ?? {}
 
+  // Hide the whole section if no tables have data
+  const isSectionDataEmpty = useIsSectionDataEmpty(tableSections)
+  if (onlyTables && isSectionDataEmpty) {
```
sorja, line 39: If we are printing only tables and the section data is empty, we return null

src/client/store/pages/assessmentSection/hooks/index.ts
```diff
@@ -43,6 +45,35 @@ export const useTableData = (props: { table: Table }): TableData => {
...
+    () =>
+      sectionTableNames.every((tableName) =>
+        TableDatas.isTableDataEmpty({
+          data,
+          tableName,
+          countryIso,
+        })
+      ),
+    [countryIso, data, sectionTableNames]
+  )
```
sorja: If we dont use `useMemo` here, we end in an infinite loop `setState <-> useEffect`

PR #1550, opened 2022-09-26
Meta/print hide section

### LINKED ISSUE #202: PrintViewOnlyTables - render tables/sections

sorja:
# Resolves #202

## Todo
- [x] Verify functionality as expected

### X09 with missing data:

image

### FIN with complete data set:

image

## PDF
Appendix:
FIN
[FRA platform.pdf](https://github.com/openforis/fra-platform/files/9653399/FRA.platform.pdf)

X09
[FRA platform.pdf](https://github.com/openforis/fra-platform/files/9653404/FRA.platform.pdf)


src/client/store/pages/assessmentSection/stateType.ts
```diff
@@ -11,4 +11,6 @@ export type AssessmentSectionState = {
   nodeValueValidation: Record<string, NodeUpdate>
   // sectionName -> name -> content
   descriptions: Record<string, Record<string, string>>
+
+  dataLoaded: boolean
```
minotogna: this is specific to print view. I wouldn't add this property.
 in the hook useIsSectionDataEmpty can u verify that if some  _**data[countryIso][tableName] is undefined then data is not loaded**_ ?

src/meta/data/tableDatas.ts
```diff
@@ -19,15 +19,23 @@ const getTableData = (props: Pick<Props, 'countryIso' | 'tableName' | 'data'>) =
...
+const isTableDataEmpty = (props: { data: TableData; tableName: string; countryIso: CountryIso; options?: Options }) => {
+  const { data, tableName, countryIso, options } = props
   const tableData = getTableData({ data, tableName, countryIso })
   if (Objects.isEmpty(tableData)) {
     return true
   }
-  // Exclude calculated nodes
+
+  const _comparator = (nodeValue: NodeValue) => {
+    return options?.excludeNull ? !nodeValue.calculated && nodeValue.raw !== null : !nodeValue.calculated
```
minotogna: why dont we exclude null by default ? it shouldn't be an option ?

src/client/store/pages/assessmentSection/stateType.ts
sorja: Makes sense, updated

minotogna (approved):

rfontanarosa (approved):

PR #1570, opened 2022-09-30
Autocomplete component for Scientific name

sorja:
Notes:
- Implements new Autocomplete semi dynamic component
- Uses [Turnstone](https://turnstone.tomsouthall.com/)

Some issues still to solve:
- [x] Saving nodeValue `code` and `scientific_name` on Cell `onChange` function called
- [x] Solve infinite loop onChange
- [x] Remove previously added taxon code if updating raw string
- [x] Solve issue onChange called on first load
- [x] Same font as input
- [x] Padding
- [x] Box shadow

https://user-images.githubusercontent.com/5508251/193575064-1c509432-d04f-482b-8a1a-760062b5ef7f.mov

image


src/client/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,67 @@
...
+  // mainly used for selecting correct key for displayName for object
+  listbox: Record<string, unknown>
+  // url: Should support query, limit params
+  url: string
+  disabled: boolean
+
+  value: string
+  name?: string
+  onChange: (value: any) => void
+}
```
sorja: Should be dynamic enough for our use case(s).
If more complexity is needed in future cases, we can alter the component

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/Taxon/Taxon.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Taxon: React.FC<PropsCell> = (props: PropsCell) => {
+  const { onChange, datum, disabled } = props
+
+  const _onChange: OnChangeTaxon = (value) => {
+    // Handle first load ajax query call of onChange
+    const isSame = value === datum || (typeof value !== 'string' && value?.scientificName === datum)
+    if (isSame || !value) return
+    onChange(value)
+  }
```
sorja: Null checking and same value checking here is required to avoid updating the datum on first load

sorja: Also, we don't want to update it if it's the same or undefined

rfontanarosa (approved):

src/client/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const Autocomplete = (props: Props) => {
```
minotogna: use React.FC

src/client/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,67 @@
...
+  // listbox: Object containing autocomplete options,
+  // mainly used for selecting correct key for displayName for object
+  listbox: Record<string, unknown>
+  // url: Should support query, limit params
+  url: string
+  disabled: boolean
+
+  value: string
+  name?: string
+  onChange: (value: any) => void
```
minotogna: do not use any. onChange: ```<T>(value: T)=>void```

src/client/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,67 @@
+import './Autocomplete.scss'
+import React from 'react'
+
+import axios from 'axios'
+import Turnstone from 'turnstone'
+
+type Props = {
+  // styles: Record of autocomplete class names
+  styles?: Record<string, string>
```
minotogna: if we do not customize style now, remove this props

src/client/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,67 @@
...
+const Autocomplete = (props: Props) => {
+  const { value, disabled, onChange, listbox, url, styles, name } = props
+
+  if (disabled) return <div className="text-input__readonly-view ">{value}</div>
+
+  // Set up listbox contents.
+  const defaultListbox = {
+    // should return array with { [listbox.displayField]: value }
+    data: (query: string) => axios.get(`${url}?query=${encodeURIComponent(query)}&limit=10`).then(({ data }) => data),
+  }
```
minotogna: this should not be added here but should be passed by the caller as prop.
we can use this autocomplete also from local data, not necessarily Ajax

minotogna: plus, use limit 20 from taxon

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/hooks/useOnChange.ts
```diff
@@ -18,30 +19,47 @@ type Props = {
   data: TableData
   sectionName: string
 }
+export type OnChangeTaxon = (value: Taxon | string) => void
+export type OnChangeDefault = (
+  event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement>
+) => void
+type OnChange = OnChangeTaxon | OnChangeDefault
```
minotogna: onChange should be generic. we should pass nodeValue here. we should refactor all other components

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/hooks/useOnChange.ts
```diff
@@ -18,30 +19,47 @@ type Props = {
...
   const countryIso = useCountryIso()
   const cycle = useCycle()
   const assessment = useAssessment()
   const assessmentSection = useAssessmentSection(sectionName)
 
-  const _persistSanitizedValue = (value: string) => {
-    const type = col.props.colType
-    if (Sanitizer.isAcceptable({ type, value })) {
+  const _persistSanitizedValue = (value: string | Taxon) => {
+    const _value = typeof value === 'string' ? value : value?.scientificName
```
minotogna: this is generic to taxon. it doesn't fit here

minotogna (changes requested):

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/hooks/useOnChange.ts
sorja: Yes, let's do separate issue for this; I saw this but I think it is slightly more complex than scope of this PR

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/hooks/useOnChange.ts
sorja: This will be removed in refactor so we can pass the nodeValue directly; we have to modify the nodeValue in the onChange, that's why it's here

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/hooks/useOnChange.ts
minotogna: I would like to start avoiding  committing to main branch code we already know we should change .
Is that really a big change ?

minotogna: The change actually can be minimum , without updating any existing components .

The hook we use to export the onChange function can export a 3rd function `onChangeNodeValue`.
The onChange will internally call onChangeNodeValue.

onChangeNodeValue accepts as param the NodeValue and it will be used only by Taxon component .

Would that work?

sorja: Sounds good, I will update the component, only thing is I have to see this piece but should be straight forward
```
Cell.tsx

  return (
    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
    <td className={className} onClick={showError} onKeyDown={showError}>
      <Component
        assessmentName={assessmentName}
        sectionName={sectionName}
        table={table}
        disabled={disabled || nodeValue?.odp}
        rowIndex={rowIndex}
        col={col}
        row={row}
        datum={datum}
        onChange={onChange} <------ 


Should we still call it onChange or should we add onChangeNodeValue and update Props?


        onPaste={onPaste}
      />
    </td>
```

src/client/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type Props = {
+  // listbox: Object containing autocomplete options,
+  // mainly used for selecting correct key for displayName for object
+  listbox: Record<string, unknown>
+  disabled: boolean
+
+  value: string
+  name?: string
+  onChange: (value: any) => void
```
minotogna: can't you use generic instead of any ?

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/props.ts
```diff
@@ -8,7 +10,9 @@ export type PropsCell = {
   rowIndex: number
   col: Col
   row: Row
+  nodeValue: NodeValue
   datum: string
```
minotogna: these two are the same. we should use only one

src/client/components/Autocomplete/Autocomplete.tsx
sorja: Issue with `.tsx` files and generics, left as `value: string | any` as user input is always `string` in input text

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/props.ts
sorja: replace with nodeValue

minotogna (approved):

PR #1582, opened 2022-10-05
#1509 Geo UI agreement level 

### LINKED ISSUE #1509: Geo UI : Agreement level
https://xd.adobe.com/view/dfbd3ca3-74e1-49e7-92da-73d5e39bf12f-8aee/screen/7faaf2aa-8adb-47e3-ac0a-7952a055bf0b

Screenshot 2022-09-15 at 17 38 18

vedran-brnjetic:
WIP for #1509
So far done some visuals, although I need to add correct icons to the iconset, but I picked a few from the existing ones

image


src/client/pages/Geo/GeoMap/GeoMapMenuData/geoMapMenuData.tsx
```diff
@@ -15,7 +15,7 @@ const GeoMapMenuData: React.FC = () => {
       <GeoMapMenuButton panel="data" text="Data Layers" icon="layers" />
       {selectedPanel === 'data' && (
         <div>
-          <GeoMenuItem title="Map Visualizer" tabIndex={-1}>
+          <GeoMenuItem title="Forest Cover" tabIndex={-1}>
```
vedran-brnjetic: This title is "Forest Cover" in the mockup

joonasnuutinen:
Implements #1740.

I hijacked the branch based on @vedran-brnjetic's work and the updated requirements in #1740.

As per the requirements, this PR implements the functionality to load the agreement layer. The UI elements are still a work in progress and will be finalized in #1741. However, I left Vedran's UI for testing purposes, as displayed below:

https://user-images.githubusercontent.com/18427577/203964824-15bd60eb-b0b4-47b1-b643-bd147dfc9ee3.mov

src/client/pages/Geo/GeoMap/GeoMapMenuData/MapVisualizerAgreementLevelsControl/mapVisualizerAgreementLevelsControl.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const map = useGeoMap()
+  const forestOptions = useForestSourceOptions()
+
+  /**
+   * Handle agreement level selection
+   * @param {number} level The agreement level
+   */
+  const handleAgreementLevelSelection = useCallback(
+    async (level: number): Promise<void> => {
+      const layerQuery = forestOptions.sources.map((name) => `&layer=${name}`).join('')
```
eraviolo: If Hansen layer is selected among the agreement layers add to the query tree cover percentage selected with 'gteHansenTreeCoverPerc'
e.g
api/geo/layers/forestAgreement/?countryIso=FIN&layer=TandemX&layer=JAXA&layer=Copernicus&layer=Hansen&gteAgreementLevel=2&gteHansenTreeCoverPerc=10

eraviolo:
> Implements #1740.
> 
> I hijacked the branch based on @vedran-brnjetic's work and the updated requirements in #1740.
> 
> As per the requirements, this PR implements the functionality to load the agreement layer. The UI elements are still a work in progress and will be finalized in #1741. However, I left Vedran's UI for testing purposes, as displayed below:
> 
>  Screen.Recording.2022-11-25.at.12.35.28.mov

Looks good!

minotogna (commented):
Great job ! 

However I would suggest following changes to better isolate single responsibilities and to better comply with our code standards.

### 1. Components and their sass file must be renamed with upper case convention

### 2. Create a MapController responsible for actions to be performed on google map instance.

**2.1** this file can be placed at the moment inside src/client/utils for now. we'll see if it needs better place later on

**2.2** 
```typescript
export class MapController {
  ... 
  constructor(map:google.maps.Map){
    this.map = map
  }

  addEarthEngineLayer(props:{mapId: string, mapLayerKey: string}):void {
  ...
  }

  ... other methods if needed
}
```

**2.3** this will be then used in MapVisualizerPanel.tsx
```typescript
const mapControllerRef = useRef<MapController>(new MapController(map))
```

### 3. move layers declared MapVisualizerPanel.tsx in a file in the same folder than MapVisualizerPanel.tsx + create enum LayerKey with JAXA, TandemX, Hansen...
these keys will be used in the layers object and everywhere they are now referenced as strings 

### 4. use redux actions/store to load/store layers from backend 
GeoActions.addForestLayer 
GeoActions.removeForestLayer 

### 5. in MapVisualizerPanel rename onCheckboxClick to toggleLayer. the action is important, not who trigger it.


How does this sound to you @joonasnuutinen ?

minotogna:
> Great job !
> 
> However I would suggest following changes to better isolate single responsibilities and to better comply with our code standards.
> 
> ### 1. Components and their sass file must be renamed with upper case convention
> ### 2. Create a MapController responsible for actions to be performed on google map instance.
> **2.1** this file can be placed at the moment inside src/client/utils for now. we'll see if it needs better place later on
> 
> **2.2**
> 
> ```ts
> export class MapController {
>   ... 
>   constructor(map:google.maps.Map){
>     this.map = map
>   }
> 
>   addEarthEngineLayer(props:{mapId: string, mapLayerKey: string}):void {
>   ...
>   }
> 
>   ... other methods if needed
> }
> ```
> 
> **2.3** this will be then used in MapVisualizerPanel.tsx
> 
> ```ts
> const mapControllerRef = useRef<MapController>(new MapController(map))
> ```
> 
> ### 3. move layers declared MapVisualizerPanel.tsx in a file in the same folder than MapVisualizerPanel.tsx + create enum LayerKey with JAXA, TandemX, Hansen...
> these keys will be used in the layers object and everywhere they are now referenced as strings
> 
> ### 4. use redux actions/store to load/store layers from backend
> GeoActions.addForestLayer GeoActions.removeForestLayer
> 
> ### 5. in MapVisualizerPanel rename onCheckboxClick to toggleLayer. the action is important, not who trigger it.
> How does this sound to you @joonasnuutinen ?

@joonasnuutinen , if you already moved to the next task, we can refactor this in a separate task.

eraviolo (changes requested):

joonasnuutinen:
@minotogna 

1. Renamed files to uppercase. There are still other files outside the scope of this PR that have lowercase naming -> future refactoring?
2. MapController created
3. Layers moved to a separate file and typed based on `@meta/geo/forest`
4. New Redux actions created and used to abstract business logic away from components
5. Renamed action

@eraviolo

Added Hansen percentage to agreement uri.

Hansen percentage selector doesn't work correctly: sometimes you need to click the checkbox twice. However, this is outside the scope of this PR so I'll create a new issue.

minotogna:
@joonasnuutinen can we have an up to date video?

joonasnuutinen:
@minotogna Sure!


https://user-images.githubusercontent.com/18427577/206452018-1a640f33-908f-4d0c-9500-9862b47e6be9.mov

eraviolo (approved):

src/client/store/ui/geo/actions/getForestLayer.ts
```diff
@@ -0,0 +1,17 @@
+import { createAsyncThunk } from '@reduxjs/toolkit'
+import axios from 'axios'
+
+interface GetForestLayerProps {
+  key: string
+  uri: string
+}
+
+export const getForestLayer = createAsyncThunk<[string, string], GetForestLayerProps>(
```
minotogna, line 9: i would prefer to make return type explicit with objects. e.g. ```createAsyncThunk<{key:string,mapId:string}, GetForestLayerProps>(```

minotogna (approved):
That's a really good PR.
I left just a minor comment that could be addressed in a later pr if we want.

PR #1597, opened 2022-10-07
user form autosave roles

### LINKED ISSUE #1520: User edit: In role edit dialog, persist updates on confirm
- [x] implement auto save for all user properties
- [x] Remove buttons 
- [ ] Auto save user roles 
- [ ] Move admin checkbox at the end of the form with window.alert confirmation 

OJ: When editing users, adding countries for reviewer or collaborator admin have to save twice, should only need to save once

rfontanarosa:
close #1520

https://user-images.githubusercontent.com/10047945/194567464-2ba92aba-a0e7-4ecc-9c22-362b7d6b8464.mov


src/client/components/EditUserForm/CountryRoles/CountryRoles.tsx
```diff
@@ -28,7 +28,7 @@ type ModalOptionsProps = {
 }
 
 type Props = {
-  onChange: (value: Array<Partial<UserRole<RoleName>>>, key: string) => void
+  onChange: (value: Array<Partial<UserRole<RoleName>>>) => void
```
minotogna: you dont need to pass onChange here. this component can dispatch the action directly

src/client/store/userManagement/actions/updateUserRoles.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { UserManagementActions } from '../slice'
+
+type Params = { roles: Array<Partial<UserRole<RoleName>>>; userId: number }
+
+const postUserRoles = Functions.debounce(async (params: Params, dispatch: Dispatch) => {
+  try {
+    const { data } = await axios.post(ApiEndPoint.User.roles(), params)
+
+    dispatch(UserManagementActions.setUserToEdit(data))
```
minotogna, line 17: dont you have already this user to edit in the store? this is not needed

src/i18n/resources/en.js
```diff
@@ -1054,6 +1054,7 @@ The FRA team
     picture1MbMax: 'Profile picture cannot exceed 1MB',
     activated: 'Activated',
     status: 'Status',
+    adminConfirm: 'Are you sure you want to promote this user to Administrator?',
```
minotogna, line 1057: we should always try to remember to add to missing translation issue

src/server/db/migration/migrations/sqls/20221007133533-user-role-on-conflict-up.sql
```diff
@@ -0,0 +1 @@
+alter table public.users_role add constraint user_country_role unique (user_id, country_iso, role);
```
minotogna: isn't user_id, country_iso unique constraint enough ?

minotogna: actually the unique constraint should be: user_id, assessment_id, cycle_uuid, country_iso .
can u verify that?

sorja (approved):
looks ok
see other comments

src/server/db/migration/migrations/sqls/20221007133533-user-role-on-conflict-up.sql
rfontanarosa: you're right

src/client/store/userManagement/actions/updateUserRoles.ts
rfontanarosa: I had some asynchronous issue with the Debounce function, so I decided to use the same logic of putOriginalDataPoint

src/server/db/migration/migrations/sqls/20221007133533-user-role-on-conflict-up.sql
```diff
@@ -0,0 +1,3 @@
+alter table public.users_role
+    add constraint user_assessment_cycle_country_role_key
+    unique (user_id, assessment_id, cycle_uuid, country_iso, role);
```
minotogna: role is not part of unique constraint . see my previous comment

minotogna (changes requested):

src/client/store/userManagement/actions/updateUserRoles.ts
rfontanarosa: I need the user roles from the response because they contain the role ids

src/server/db/migration/migrations/sqls/20221007133533-user-role-on-conflict-up.sql
rfontanarosa: you're right again. thanks.

minotogna (approved):

PR #1602, opened 2022-10-11
Allow custom combination for agreement layer

eraviolo:


minotogna (approved):

src/server/api/geo/getForestAgreementLayer.ts
```diff
@@ -1,19 +1,39 @@
...
 import { GeoController } from '@server/controller/geo'
 import Requests from '@server/utils/requests'
 
 export const getForestAgreementLayer = async (req: Request, res: Response) => {
   try {
     const countryIso: CountryIso = req.params.countryIso as CountryIso
-    const gteAgreementLevel = Number(req.params.gteAgreementLevel)
-    const gteHansenTreeCoverPerc = Number(req.params.gteHansenTreeCoverPerc)
+    const gteAgreementLevel = Number(req.query.gteAgreementLevel)
+    const gteHansenTreeCoverPerc = Number(req.query.gteHansenTreeCoverPerc)
```
minotogna, line 14: to avoid casting to all query parameters, you could type the Request.
e.g.
```typescript
export const getForestAgreementLayer = async (req: Request<{countryIso:CountryIso}, never, never, {gteAgreementLevel:number;gteHansenTreeCoverPerc:number; layer:Array<ForestSource>}>
, res: Response) => {
const {countryIso}=req.params
const {gteAgreementLevel, gteHansenTreeCoverPerc,layer}=req.query
...
```

Btw we're migrating apis to use only query, not params. we can discuss this later on. not important at all now

rfontanarosa (approved):

PR #1618, opened 2022-10-14
Type req and add opacity as query param 

eraviolo:
opacity: from 0 to 1, default 1

### DIFF
--- a/src/server/api/geo/getForestAgreementLayer.ts
+++ b/src/server/api/geo/getForestAgreementLayer.ts
@@ -6,33 +6,54 @@ import { ForestSource } from '@meta/geo'
 import { GeoController } from '@server/controller/geo'
 import Requests from '@server/utils/requests'
 
-export const getForestAgreementLayer = async (req: Request, res: Response) => {
+export const getForestAgreementLayer = async (
+  req: Request<
+    { countryIso: CountryIso },
+    never,
+    never,
+    { gteAgreementLevel: number; gteHansenTreeCoverPerc: number; layer: Array<string>; opacity: number }
+  >,
+  res: Response
+) => {
   try {
-    const countryIso: CountryIso = req.params.countryIso as CountryIso
-    const gteAgreementLevel = Number(req.query.gteAgreementLevel)
-    const gteHansenTreeCoverPerc = Number(req.query.gteHansenTreeCoverPerc)
-
+    const { countryIso } = req.params
+    const { gteAgreementLevel, gteHansenTreeCoverPerc, layer, opacity } = req.query
     const sourceOptions = Object.values(ForestSource)
 
-    if (Array.isArray(req.query.layer) && req.query.layer.length > 1) {
+    if (Array.isArray(layer) && layer.length > 1) {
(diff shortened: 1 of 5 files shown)


src/server/api/geo/getForestAgreementLayer.ts
```diff
@@ -6,33 +6,54 @@ import { ForestSource } from '@meta/geo'
 import { GeoController } from '@server/controller/geo'
 import Requests from '@server/utils/requests'
 
-export const getForestAgreementLayer = async (req: Request, res: Response) => {
+export const getForestAgreementLayer = async (
+  req: Request<
```
sorja, line 10: Not sure if it's valid for your case but you can use also CycleRequest?

sorja: Found under `meta/`

src/server/api/geo/getForestAgreementLayer.ts
```diff
@@ -6,33 +6,54 @@ import { ForestSource } from '@meta/geo'
...
+    { countryIso: CountryIso },
+    never,
+    never,
+    { gteAgreementLevel: number; gteHansenTreeCoverPerc: number; layer: Array<string>; opacity: number }
```
sorja: Is is better to defiine layer as `Array<ForestSource` ? Then you dont need to cast it later to `unknown` and `ForestSource`

src/server/api/geo/getForestAgreementLayer.ts
```diff
@@ -6,33 +6,54 @@ import { ForestSource } from '@meta/geo'
...
+      layer.forEach((layerId: string) => {
+        if (sourceOptions.includes(layerId as unknown as ForestSource)) {
+          if (
+            layerId === ForestSource.Hansen &&
+            (gteHansenTreeCoverPerc === undefined ||
+              Number.isNaN(Number(gteHansenTreeCoverPerc)) ||
+              gteHansenTreeCoverPerc < 0 ||
+              gteHansenTreeCoverPerc > 100)
+          )
+            throw Error(`Not valid Hansen tree cover percentage 0-100: ${gteHansenTreeCoverPerc}`)
```
sorja, line 35: If you `throw Error` I think the whole node/express server shuts down
Better to use `Requests.sendErr`

src/server/api/geo/getForestAgreementLayer.ts
eraviolo: I think for now we are not linking to a specific cycle, (only using ISO to clip country) but is something to be discussed.

src/server/api/geo/getForestAgreementLayer.ts
eraviolo: Yes, changed to Array<ForestSource>

src/server/api/geo/getForestAgreementLayer.ts
eraviolo: In these cases the errors are catched below and Request.sendErr is called

minotogna (approved):

sorja (approved):

PR #1626, opened 2022-10-18
new debounce logic

### LINKED ISSUE #1608: Data update: debouncing error
When entering data at fast speed, intermediate changes get lost.

The solution would be to invoke the debounced function based on the unique keys.

solution 1
add  Objects.createSingleton 
```typescript
const createSingleton = <T>(name: string, create: () => T): T => {
  const s = Symbol.for(name)
  let scope = (global as any)[s]
  if (!scope) {
    scope = { ...create() }
    ;(global as any)[s] = scope
  }
  return scope
}

and customize Functions.debounce to take an id as well 
```

solution 2
create within functions  debounced objects (see https://github.com/openforis/arena/blob/master/core/functionsDefer.js)

solution 3
See this post on how a memoized version is created :
https://github.com/lodash/lodash/issues/2403#issuecomment-290760787

see video : after refresh only last update has been persisted

https://user-images.githubusercontent.com/1138156/195519032-8e49ecc8-2fb9-4feb-967e-8bbbf4e25a62.mov

test following actions work properly

Screenshot 2022-10-13 at 08 31 08

rfontanarosa:
close #1608

- [x] updateDescriptions - `${props.countryIso}-${props.sectionName}-${props.name}`
- [x] updateNodeValues - `${props.countryIso}-${props.tableName}-${props.values[0].variableName}-${props.values[0].colName}`
- [x] updateOriginalDataPoint - 'updateOriginalDataPoint'
- [x] updateUser - 'updateUser'
- [x] updateUserRoles - 'updateUserRoles'

https://user-images.githubusercontent.com/10047945/196435654-9afd2e92-84ca-4e68-8b72-4d4a581fa4e4.mov

### DIFF
--- a/src/client/pages/OriginalDataPoint/OriginalDataPoint.tsx
+++ b/src/client/pages/OriginalDataPoint/OriginalDataPoint.tsx
@@ -62,7 +62,7 @@ const OriginalDataPoint: React.FC = () => {
         })
       )
     }
-  }, [originalDataPoint?.id, countryIso, assessmentName, cycleName, year, user, dispatch, sectionName])
+  }, [originalDataPoint?.id, countryIso, assessmentName, cycleName, user, dispatch, sectionName])
 
   if (!originalDataPoint) return null
 
--- a/src/client/pages/OriginalDataPoint/OriginalDataPoint.tsx
+++ b/src/client/pages/OriginalDataPoint/OriginalDataPoint.tsx
@@ -62,7 +62,7 @@ const OriginalDataPoint: React.FC = () => {
         })
       )
     }
-  }, [originalDataPoint, countryIso, assessmentName, cycleName, year])
+  }, [originalDataPoint?.id, countryIso, assessmentName, cycleName, year, user, dispatch, sectionName])
 
   if (!originalDataPoint) return null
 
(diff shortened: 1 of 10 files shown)


src/utils/functions.ts
```diff
@@ -1,7 +1,26 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
sorja: Should we add inline

src/client/store/pages/assessmentSection/actions/updateNodeValues.ts
```diff
@@ -7,14 +7,22 @@ import { CycleDataParams, NodesBody } from '@meta/api/request'
...
-}, 1000)
+const patchNodeValues = (id: string) =>
+  Functions.debounce(
+    async ({ tableName, values, ...params }: Props) => {
+      try {
+        await axios.patch(ApiEndPoint.CycleData.Table.nodes(), { tableName, values }, { params })
+      } catch (e) {
+        // placeholder to avoid app crash
+      }
+    },
```
sorja, line 18: Does this still crash app?

sorja (approved):

src/client/pages/OriginalDataPoint/OriginalDataPoint.tsx
```diff
@@ -62,7 +62,7 @@ const OriginalDataPoint: React.FC = () => {
         })
       )
     }
-  }, [originalDataPoint, countryIso, assessmentName, cycleName, year])
+  }, [originalDataPoint?.id, countryIso, assessmentName, cycleName, year, user, dispatch, sectionName])
```
minotogna: do we need 'year' in the dependencies of this hook? probably not

minotogna (approved):

PR #1633, opened 2022-10-19
Fix issue with unknown calculation where value was not rounded

### LINKED ISSUE #1518: Calculation results: round results to 2 digits
OJ: Table 4a, mistake generated when the unknown ownership (4a) is using values from NDP in table 1a. Albania, Azerbaijan, Kazakhstan had a problem with the decimals. The NDP in table 1a generates a number of decimals which are not reflected in table 4a resulting in a very, very tiney negative difference. Had to add zeros manually in the sheet “Data”  (SEE ALBANIA: FIX: Round final calculation results)

sorja:
resolves #1518

before
-0.00 (value: 0.0025)

image

after
image

### DIFF
--- a/src/server/controller/cycleData/persistNodeValue/calculateDependantNodes/calculateNode.ts
+++ b/src/server/controller/cycleData/persistNodeValue/calculateDependantNodes/calculateNode.ts
@@ -37,9 +37,9 @@ export const calculateNode = async (
     formula: expression,
   })
 
-  const roundedResult = Numbers.format(Numbers.toBigNumber(rawResult))
+  const raw = rawResult ? Numbers.format(Numbers.toBigNumber(rawResult)) : null
 
-  const value: NodeValue = { raw: rawResult ? roundedResult : null, calculated: true }
+  const value: NodeValue = { raw, calculated: true }
 
   return persistNode(
     {
--- a/src/server/controller/cycleData/persistNodeValue/calculateDependantNodes/calculateNode.ts
+++ b/src/server/controller/cycleData/persistNodeValue/calculateDependantNodes/calculateNode.ts
@@ -1,3 +1,5 @@
+import { Numbers } from '@utils/numbers'
+
 import { ActivityLogMessage, Node, NodeValue, Row } from '@meta/assessment'
 
 import { getTableData } from '@server/controller/cycleData/getTableData'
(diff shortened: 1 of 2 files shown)


src/test/sectionSpec/fraSpecs.ts
```diff
@@ -6025,7 +6025,7 @@ export const FraSpecs: Record<string, SectionSpec> = {
                 migration: {
                   colNames: ['1990', '2000', '2010', '2015'],
                   calcFormula: `forestOwnership.private_ownership || forestOwnership.public_ownership 
-                    ? extentOfForest.forestArea - (forestOwnership.private_ownership || 0) - (forestOwnership.public_ownership || 0)
+                    ? Math.round((extentOfForest.forestArea - (forestOwnership.private_ownership || 0) - (forestOwnership.public_ownership || 0) + Number.EPSILON) * 100) / 100
```
minotogna: all calculation results must be rounded to 2. not each single formula.
check src/server/controller/cycleData/persistNodeValue/calculateDependantNodes/calculateNode.ts line 37 where results is parsed from ExpressionEvaluator

minotogna (changes requested):

src/test/sectionSpec/fraSpecs.ts
sorja: great, wasn't 100% sure if this was what we wanted, ill fix this so

src/server/controller/cycleData/persistNodeValue/calculateDependantNodes/calculateNode.ts
```diff
@@ -34,7 +36,10 @@ export const calculateNode = async (
     row,
     formula: expression,
   })
-  const value: NodeValue = { raw: rawResult ? String(rawResult) : null, calculated: true }
+
+  const roundedResult = Numbers.format(Numbers.toBigNumber(rawResult))
+
+  const value: NodeValue = { raw: rawResult ? roundedResult : null, calculated: true }
```
minotogna: I would round if there's the rawResult , not necessarily always

sorja: Makes sense!

rfontanarosa (approved):

minotogna (approved):

PR #1639, opened 2022-10-20
Added reject function and auth. token check

eraviolo:
- Added check to do authentication only if auth. token is not present
- Added reject function to call if gee authentication fails


minotogna (approved):

minotogna:
I'm wondering if gee earth can be handled in a middleware. All other apis won't need to call it .

sorja (approved):

eraviolo:
I think it is a good idea, I will try that.

PR #1647, opened 2022-10-24
Geo gee auth middleware

eraviolo:
- Moved gee authentication from controller to middleware function.
- Moved params to query params for forest, forest agreement and boundaries endpoints.


src/server/api/geo/getForestAgreementLayer.ts
```diff
@@ -8,16 +8,21 @@ import Requests from '@server/utils/requests'
 
 export const getForestAgreementLayer = async (
   req: Request<
-    { countryIso: CountryIso },
     never,
     never,
-    { gteAgreementLevel: number; gteHansenTreeCoverPerc: number; layer: Array<ForestSource>; opacity: number }
+    never,
+    any & {
```
sorja: why do you extend any here?

eraviolo: Just to avoid 'No overload matches this call' for the calls after authentication in the chain, but I am sure missing something about TypeScript here, how should I handle this better?

src/server/api/geo/getBoundariesLayer.ts
```diff
@@ -7,7 +7,7 @@ import Requests from '@server/utils/requests'
 
 export const getBoundariesLayer = async (req: Request, res: Response) => {
   try {
-    const { countryIso } = req.params as {
+    const { countryIso } = req.query as {
```
minotogna: better to add types to Request

src/server/controller/geo/getForestLayer.ts
```diff
@@ -16,7 +16,7 @@ type Props = {
 export const getForestLayer = async (props: Props): Promise<Layer> => {
   const { countryIso, forestSource, gteHansenTreeCoverPerc, onlyProtected, opacity } = props
 
-  await authenticateToGee()
+  // await authenticateToGee()
```
minotogna: better to remove comments

src/server/api/geo/index.ts
```diff
@@ -2,14 +2,16 @@ import { Express } from 'express'
...
 import { getBoundariesLayer } from './getBoundariesLayer'
 import { getForestAgreementLayer } from './getForestAgreementLayer'
 import { getForestLayer } from './getForestLayer'
 
 export const GeoApi = {
   init: (express: Express): void => {
-    express.get(ApiEndPoint.Geo.Layers.getForest(), getForestLayer)
-    express.get(ApiEndPoint.Geo.Layers.getForestAgreement(), getForestAgreementLayer)
-    express.get(ApiEndPoint.Geo.Layers.getBoundaries(), getBoundariesLayer)
+    express.get(ApiEndPoint.Geo.Layers.getForest(), checkGeeAuthentication, getForestLayer)
```
minotogna: ```express.get(ApiEndPoint.Geo.Layers.getForest(), GeeAuthMiddleware.requireLogin, getForestLayer)```

src/server/api/geo/getForestAgreementLayer.ts
minotogna: I just checked out this branch, and I see no error if I remove any .
Screenshot 2022-10-25 at 17 01 29

minotogna (changes requested):

src/server/api/geo/getForestAgreementLayer.ts
sorja: Also here, is it possible to rename Perc => Percent?
No matter, I think just nicer to read

sorja:
For request types, see this file:
`src/meta/api/request/index.ts`

We use it to define the request type (find usages to see examples)

Maybe you could add your request types in this file, and use them where needed?

For example:
`export type { GeoRequest } from './geo/geo'` (or `'./geo'` where `./geo/index.ts`)

and

`export type GeoRequest = Request<never, never, never, { property1: type, ..., propertyn: type,  }>`

src/server/api/geo/getForestAgreementLayer.ts
eraviolo: I am changing all parameters to string as ParsedQs interface allows only string as a type. In future PR I will create a middlware function to validate number parameters.

minotogna (approved):

sorja (approved):

PR #1655, opened 2022-10-26
admin blank page

### LINKED ISSUE #1653: Admin : route with blank page
When accessing page , if user is not admin redirect to home page

rfontanarosa:
close #1653

Schermata 2022-10-26 alle 09 27 24

### DIFF
--- a/src/client/pages/Admin/Admin.tsx
+++ b/src/client/pages/Admin/Admin.tsx
@@ -1,4 +1,5 @@
-import React from 'react'
+import React, { useEffect } from 'react'
+import { useNavigate } from 'react-router-dom'
 
 import { ClientRoutes } from '@meta/app'
 import { Users } from '@meta/user'
@@ -7,10 +8,13 @@ import { useUser } from '@client/store/user'
 
 const Admin: React.FC = () => {
   const user = useUser()
+  const navigate = useNavigate()
 
-  const isAdministrator = Users.isAdministrator(user)
+  useEffect(() => {
+    if (Users.isAdministrator(user)) navigate(ClientRoutes.Root.path, { replace: true })
+  }, [navigate, user])
 
-  if (!isAdministrator) window.location.href = ClientRoutes.Root.path
+  if (!user) return null
 
   return <div className="app-view__content">Placeholder</div>
 }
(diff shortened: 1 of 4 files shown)


src/client/pages/Admin/Admin.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Users } from '@meta/user'
+
+import { useUser } from '@client/store/user'
+
+const Admin: React.FC = () => {
+  const user = useUser()
+
+  const isAdministrator = Users.isAdministrator(user)
+
+  if (!isAdministrator) window.location.href = ClientRoutes.Root.path
```
sorja: is it better to use react-router hooks

sorja (approved):

src/client/pages/Admin/Admin.tsx
minotogna: yes, try using react router in this case

rfontanarosa: there are two solutions:

```
<Route
  path={`${ClientRoutes.Admin.Root.path.absolute}/*`}
  element={isAdministrator ? <Admin /> : <Navigate replace to={ClientRoutes.Root.path} />}
/>
```

OR

```
const navigate = useNavigate()
if (!isAdministrator) navigate(ClientRoutes.Root.path)
```

src/client/pages/Admin/Admin.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const user = useUser()
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    if (!Users.isAdministrator(user)) navigate(ClientRoutes.Root.path, { replace: true })
+  }, [navigate, user])
+
+  if (!user) return null
+
+  return <div className="app-view__content">Placeholder</div>
```
minotogna: if there'a user logged in, and it's not admin, on first mount this goes through. if there's a way to avoid this, avoid it.

rfontanarosa: ok, there is no other chance than put a condition around the route

sorja: `|| !Users.isAdministrator(user)`

minotogna (approved):

PR #1657, opened 2022-10-26
WIP: Metadata 2025: metadata migration section 3d

sorja:
Adding a new subsection to fraSpecs doesn't add it to platform (db etc)

- [ ] Solve issue where table is not shown
- [ ] Solve issue where table is not in _legacy schema


minotogna:
where do you need to add a new subsection ?

sorja:
> where do you need to add a new subsection ?

image


I added a copy of `degradedForest` as a placeholder

Issue #1654, opened 2022-10-26
Admin : user list 


minotogna:
- [x] add tab bar to navigate to users page (which is the default when accessing /admin)
- [x] Add endpoint to fetch users (add limit, and offset params )
- [x] Initial mount fetch from offset 0 with limit 20
- [x] on click edit user, navigate to /admin/users/:userId
- [x] Implement pagination
- [x] Implement users filter ui 
- [x] Update backend query to fetch filtered users  
- [x] UserCount component

PR #1658, opened 2022-10-26
admin/userManagement route

rfontanarosa:
part of #1654

Schermata 2022-10-26 alle 16 16 54

### DIFF
--- a/src/client/pages/Admin/Admin.scss
+++ b/src/client/pages/Admin/Admin.scss
@@ -17,10 +17,9 @@
 }
 
 .landing__page-title {
-  &.title {
-    display: flex;
-    align-items: center;
-  }
+  display: flex;
+  align-items: center;
+  flex: 1;
 }
 
 .landing__page-menu {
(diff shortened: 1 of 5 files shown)


src/meta/app.ts
```diff
@@ -37,13 +37,19 @@ export enum AssessmentHomeRouteNames {
   links = 'links',
 }
 
+export enum AdminRouteNames {
+  userManagement = 'userManagement',
+  dataExport = 'dataExport',
```
sorja: is dataExport still used?

rfontanarosa: idk, I can comment it for now, as the section value in Admin.tsx

src/client/pages/Admin/Admin.scss
```diff
@@ -0,0 +1,68 @@
...
+
+.app-view__content {
+  grid-area: content;
+  height: 100%;
+  padding: 28px 32px 48px 32px;
+  overflow-x: auto;
+  overflow-y: hidden;
+}
+
+.landing__page-header {
```
minotogna: this is not landing. this is admin 
rename the css according to the component

minotogna (changes requested):

sorja (approved):

src/client/pages/Admin/Admin.scss
minotogna: or check if those rule have already been defined in landing page

minotogna (approved):

PR #1640, opened 2022-10-20
Update dependant nodes for ODP

sorja:
# Todo:
- [x] Merge 1a and 1b node updates to avoid double update for node

Resolves #1627

https://user-images.githubusercontent.com/5508251/198965817-238c2d22-20b0-48fc-a3f1-785ddaa3ba77.mov

Resolves #1666

https://user-images.githubusercontent.com/5508251/199221045-90919871-dc32-4a8d-9a87-0f669f74c685.mov

PR #1665, opened 2022-10-28
Forest Characteristics: Fix issue with column names

### LINKED ISSUE #1660: ODP edit: doesn't show and edit values in 1b

sorja:
Resolves #1660

Video

https://user-images.githubusercontent.com/5508251/198568831-cd06fa7f-8f3e-4192-98e3-b4bcae605024.mov

Description
Fixes issue where column names were mismatched; historically columns were prefixed with 'forest' but omitted in later changes. Renamed properties in national class to match previous

### DIFF
--- a/src/client/pages/OriginalDataPoint/components/ForestCharacteristics/ForestCharacteristics.tsx
+++ b/src/client/pages/OriginalDataPoint/components/ForestCharacteristics/ForestCharacteristics.tsx
@@ -26,7 +26,7 @@ const ForestCharacteristics: React.FC<Props> = (props) => {
   const plantationTotal = ODPs.calcTotalSubFieldArea({
     originalDataPoint,
     field: 'forestPercent',
-    subField: 'plantationPercent',
+    subField: 'forestPlantationPercent',
   })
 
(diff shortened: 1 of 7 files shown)


minotogna (approved):

minotogna:
We should also use web socket to update table 1a and 1b on odp changes .
Separate issue , you can create it 
What do you think ?

sorja:
> We should also use web socket to update table 1a and 1b on odp changes .

It should be handled here: #1640 
https://github.com/openforis/fra-platform/pull/1640/files#diff-9d286c712831435f61977bb968d28830e590b3d8aebe96663e25db74601c0e23R24

minotogna:
> > We should also use web socket to update table 1a and 1b on odp changes .
> 
> 
> 
> It should be handled here: #1640 
> 
> https://github.com/openforis/fra-platform/pull/1640/files#diff-9d286c712831435f61977bb968d28830e590b3d8aebe96663e25db74601c0e23R24

Even though it's slightly different .
The issue mentioned above will update dependent nodes, we should update odp themselves as well

rfontanarosa (approved):

PR #1668, opened 2022-11-01
admin user list and admin edit user

rfontanarosa:
part of #1654

https://user-images.githubusercontent.com/10047945/199932871-59fcc207-d378-467d-875c-0d2b6e4bc63c.mov


sorja (approved):

src/client/components/UserList/UserList.tsx
```diff
@@ -116,15 +119,46 @@ const UserRow: React.FC<{ user: User; showEmail: boolean }> = ({ user, showEmail
...
+    </tr>
+  )
+}
+
+const UsersTableHeadRow: React.FC<{ showLoginEmail: boolean; showRole: boolean; showRoles: boolean }> = ({
+  showLoginEmail,
+  showRole,
+  showRoles,
+}) => {
+  const { t } = useTranslation()
```
minotogna: Split each component in single files

src/meta/app.ts
```diff
@@ -49,7 +49,8 @@ export const ClientRoutes = {
 
   Admin: {
     Root: newInstance<void>('admin'),
-    Section: newInstance<{ sectionName: AssessmentHomeRouteNames }>(':sectionName'),
+    Section: newInstance<{ sectionName: AssessmentHomeRouteNames }>('admin', ':sectionName'),
+    User: newInstance<{ id: number }>('admin', 'users/:id'),
```
minotogna, line 53: Why both user and section ? Use only one

minotogna (changes requested):
Post a video

src/meta/app.ts
sorja: Also how come this is `'users/:id'` instead of `'users', ':id'`

rfontanarosa: this is the subrouter of the admin section:

```
      <Routes>
        {sections.map(({ name, component }) => (
          <Route key={name} path={name} element={React.createElement(component, {})} />
        ))}

        <Route path={ClientRoutes.Admin.User.path.relative} element={<User />} />

        <Route path="*" element={<Navigate to={AdminRouteNames.userManagement} replace />} />
      </Routes>
```

the admin section name is dynamic as the home sections. at the moment we have userManagement, in the future we will add dataExport.

the admin/users/:id is for the user form, as the user form in home.

about 'users/:id'... I already done this for home. the problem is that the router wants a relative path and, at the moment, splitting that address to get the relative path returns only :id:

```
    path: {
      absolute,
      relative: parts[parts.length - 1],
    },
```

src/client/components/UserList/UserList.tsx
rfontanarosa: Schermata 2022-11-02 alle 15 28 28

src/client/components/UserList/UserListElement/UserListElement.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <tr
+      className={classNames({
+        'user-list__inactive-user': status === UserStatus.inactive,
+      })}
+    >
+      <UserField user={user} field="name" />
+      <td className="user-list__cell">
+        <div className="user-list__cell--read-only">TODO</div>
```
minotogna: what is this TODO?

src/client/components/UserList/UserListElement/UserListElement.tsx
```diff
@@ -0,0 +1,131 @@
...
+      <td className="user-list__cell user-list__edit-column">
+        <Link to={ClientRoutes.Admin.User.getLink({ id })} type="button" className="link">
+          {t('userManagement.edit')}
+        </Link>
+      </td>
+    </tr>
+  )
+}
+
+const CollaboratorListElement: React.FC<{ user: User }> = ({ user }) => {
```
minotogna: esporatare I due componenti in file separati

src/client/components/UserList/UserListHeader/UserListHeader.tsx
```diff
@@ -0,0 +1,29 @@
...
+  showRoles,
+}) => {
+  const { t } = useTranslation()
+
+  return (
+    <thead>
+      <tr>
+        <th className="user-list__header-cell">{t('userManagement.name')}</th>
+        {showRole && <th className="user-list__header-cell">{t('userManagement.role')}</th>}
+        {showRoles && <th className="user-list__header-cell">TODO</th>}
```
minotogna: what's this TODO?

minotogna (changes requested):

src/meta/user/userRoles.ts
```diff
@@ -1,5 +1,16 @@
+import { RoleName } from './userRole'
+
 const noRole = { role: 'NONE', labelKey: 'user.roles.noRole' }
 
+const roles = [
```
minotogna: where is this used? is it really necessary ?

minotogna: call it roleNames if you really need it. I would prefer not to have it

rfontanarosa: renamed into roleNames. It is used by two different components: UserListHeader and AdministrationListElement

src/meta/user/userRoles.ts
```diff
@@ -1,5 +1,16 @@
+import { RoleName } from './userRole'
+
 const noRole = { role: 'NONE', labelKey: 'user.roles.noRole' }
 
+const roleNames = [
```
minotogna: This is not correct , because it doesn't include all roles .
This is not a global object .
These are these are the role names used in admin view. Create an object there if you need

minotogna (changes requested):

src/meta/user/userRoles.ts
rfontanarosa: done, the array is defined into UserList and it is passed via component props

minotogna (approved):

PR #1676, opened 2022-11-03
Initial commit invalidate token

### LINKED ISSUE #1664: Invalidate invalid token
Issue:
Invalidate invalid token (eg. user not found, expired)

Steps to reproduce:
- After migrate-fra UI shows as logged in
- Trying to edit data shows an error

- Error probably related to token persisting, token should be reset if issue fetching user (eg. writing to activity log)

Correct steps:
- Log in
- Migrate fra
- Logged out

sorja:
Resolves #1664 

Description:
Invalidate token when email and/or id fail (eg id fails when yarn migrate-fra ran)

- [x] Discuss what field against validate (example vs `user.id`)


src/server/serverInit.ts
```diff
@@ -27,6 +28,7 @@ export const serverInit = () => {
   app.use(bodyParser.json({ limit: '5000kb' }))
 
   resourceCacheControl.init(app)
+  app.use(validateToken)
```
minotogna, line 31: Can't this be part of passport middleware when we get the user ?

sorja: We only get the user when we click login, after that the user is persisted in the token

https://github.com/openforis/fra-platform/blob/0257fb3bee850c88fb55a547546d4b3dd73051fc/src/server/api/auth/login.ts#L47

sorja: This pr resolves using invalid token (eg mismatch user.id and user.email or missing token.validation which is encrypted)

minotogna: dont we get the user from passport.deserializeUser? we shouldn't store all user in the token . or what do you think?

sorja: yes, but is this not called only when logging in?
otherwise in the code we use getRequestUser
https://github.com/openforis/fra-platform/blob/39435d6a07c2734ffae5f42bb14bca3e42c6d965/src/server/utils/requests.ts#L48

lets shortly discuss this matter in daily and agree on action points?

minotogna: I think it's called at every request , have you tested ?

sorja: After testing this I can confirm it is *never* called

minotogna (approved):

rfontanarosa (approved):

PR #1677, opened 2022-11-03
Add express-validation middleware

eraviolo:
This PR is to test/review express-validation middleware:
- Added middleware/validation folder with: validateRequest function and schema to validate getForest request params  
- Chained validation call in /server/api/geo/index.ts, prior to gee auth. middleware call   
- Changed ForestLayerRequest params that are not string to any to be able to used directly the req sanitized values (e.g string params converted to numbers or booleans)


src/server/middleware/validation/geoSchemes.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { ForestSource, precalForestAgreementSources } from '@meta/geo'
+
+export const forestLayerSchema = [
+  query('countryIso')
+    .exists()
+    .withMessage('Country ISO is required')
+    .isLength({ min: 3, max: 3 })
+    .withMessage('Country ISO must be a 3 char value')
+    .bail(),
```
eraviolo: I have to add here a validation against CountryIso type valid options

src/server/middleware/validation/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  return async (req: Request, res: Response, next: NextFunction) => {
+    await Promise.all(checks.map((check: ValidationChain) => check.run(req)))
+
+    const result = validationResult(req)
+    if (result.isEmpty()) {
+      return next()
+    }
+
+    const errors = result.array()
+    return res.send(errors)
```
minotogna: this validation middleware you added is really well done 👍.
before sending error back,I would just add a 400 status code. what do you think?

eraviolo: Yes, I will add the code before sending errors back

src/meta/api/request/geo/layer.ts
```diff
@@ -10,9 +10,9 @@ export type ForestLayerRequest = Request<
   {
     countryIso: CountryIso
     forestSource: ForestSource
-    gteHansenTreeCoverPerc?: string
-    onlyProtected: string
-    opacity: string
+    gteHansenTreeCoverPerc?: any
+    onlyProtected: any
+    opacity: any
```
sorja, line 15: why was this changed to any?

src/server/controller/geo/getForestLayer.ts
```diff
@@ -27,7 +27,7 @@ export const getForestLayer = async (props: Props): Promise<Layer> => {
       .getMap(
         {
           palette: metadata.palette,
-          opacity: opacity === undefined ? 1 : Number(opacity),
+          opacity: opacity === undefined ? 1 : opacity,
```
sorja: can this be
`opacity: opacity ?? 1`

src/server/middleware/validation/geoSchemes.ts
```diff
@@ -0,0 +1,35 @@
...
+    .isIn(precalForestAgreementSources)
+    .withMessage('Not valid forest source'),
+
+  query('gteHansenTreeCoverPerc')
+    .if(query('forestSource').equals(ForestSource.Hansen))
+    .exists()
+    .withMessage('Hansen tree cover percentage is required')
+    .bail()
+    .isInt({ min: 0, max: 100 })
+    .withMessage('Not valid Hansen tree cover percentage 0-100')
```
sorja: Can we add new translation file 'validations.js' and import it in 'en.js' and add these there?

src/server/middleware/validation/index.ts
```diff
@@ -0,0 +1,20 @@
+import { NextFunction, Request, Response } from 'express'
+import { ValidationChain, validationResult } from 'express-validator'
+
+const validateRequest = (checks: Array<ValidationChain>) => {
```
sorja, line 4: This looks good, good job 👍

src/meta/api/request/geo/layer.ts
eraviolo: When request validation is performed, I convert some query params  to number. I wanted to use directly those sanitized values, but request params cannot be typed with other than string type as I understand. With any I can pass converted params (as numbers) to the controller function, maybe there is a better workaround.

src/server/controller/geo/getForestLayer.ts
eraviolo: Sure, I was not familiar with that operator :)

src/server/middleware/validation/geoSchemes.ts
eraviolo: Yes, I will change this, I wanted to ask you about how to handle localized messages.

sorja: Normally we send translation key to frontend where we use `{ t } = useTranslation()` and later in the component `t('translation.key')`
Rarely on server side we have to initialise i18n and use it there, but there are some cases (emails)

src/server/middleware/validation/index.ts
eraviolo: I removed this control in last commit because I am setting default values when validating/sanitizing request params

src/server/middleware/validation/geoSchemes.ts
eraviolo: As we are using this api only internally maybe we can keep it simple for now by no overriding the default express-validator error message, what do you think? Default error message is "Invalid value" with param name and value, something like:
{
    "value": "3",
    "msg": "Invalid value",
    "param": "opacity",
    "location": "query"
  }

src/server/controller/geo/getForestLayer.ts
eraviolo: I removed this control in last commit because I am setting default values when validating/sanitizing request params

src/server/middleware/validation/geoSchemes.ts
sorja: Looks good

minotogna: as msg pass a msgKey . like msgKey:'validation.errors.invalidValues'

src/meta/api/request/geo/layer.ts
minotogna: this is a really good pr, but i share the concern about any. is it possible to find a solution to not use any here?

minotogna: @sorja we could leave any. it requires to update ```interface ParsedQs { [key: string]: undefined | string | string[] | ParsedQs | ParsedQs[] | number }``` from the query string namespace. 
or do you think we should try to give it a shoot anyway ?

eraviolo: I get a compilation error if typing ReqQuery with other than string or any:
Screenshot from 2022-11-16 13-47-00

"No overload matches this call.
  The last overload gave the following error.
    Argument of type '(req: ForestLayerRequest, res: Response) => Promise<void>' is not assignable to parameter of type 'RequestHandlerParams<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'"

...because this is the interface:
 interface ParsedQs { [key: string]: undefined | string | string[] | ParsedQs | ParsedQs[] }

sorja: Scope of this PR any is ok

minotogna (approved):

sorja (approved):

PR #1679, opened 2022-11-06
eslint 8.22.0

sorja:
1 resolves this.liboptions.parse is not a function

https://youtrack.jetbrains.com/issue/WEB-57089/ESLint823-TypeError-thislibOptionsparse-is-not-a-function

2 resolves params type
Screenshot 2022-11-06 at 17 18 35

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "css-loader": "^6.7.1",
     "css-minimizer-webpack-plugin": "^3.0.2",
     "csv": "^6.0.5",
-    "eslint": "^8.12.0",
+    "eslint": "8.22.0",
     "eslint-config-airbnb": "^19.0.4",
     "eslint-config-prettier": "^8.5.0",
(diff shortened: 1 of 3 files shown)


minotogna (changes requested):
Why not a full dependency upgrade ?

sorja:
> Why not a full dependency upgrade ?

can do, was doing playwright stuff and everything broke so opened a separate pr only for the needed change for now

full upgrade can take time if apis change / no backwards compatibility


it can be done, but will do the playwright init first

minotogna:
> > Why not a full dependency upgrade ?
> 
> can do, was doing playwright stuff and everything broke so opened a separate pr only for the needed change for now
> 
> full upgrade can take time if apis change / no backwards compatibility
> 
> it can be done, but will do the playwright init first

sounds good : 👍

minotogna (approved):

rfontanarosa (approved):

PR #1682, opened 2022-11-07
admin user list filters

rfontanarosa:
part of #1654

https://user-images.githubusercontent.com/10047945/200548314-12ec3a8e-d165-49db-8b7a-9b060d8f0fdd.mov


src/client/components/UserListFilters/UserListFilters.scss
```diff
@@ -0,0 +1,30 @@
...
+
+.users__table-filter {
+  display: grid;
+  grid-template-columns: 100%;
+  grid-column-gap: 10px;
+  //margin: 0 25%;
+
+  border-radius: 3px;
+  margin: 10px 20% 20px;
+  padding: 20px;
```
sorja: can we use vars?

src/client/components/UserListFilters/UserListFilters.scss
```diff
@@ -0,0 +1,30 @@
...
+  background-color: $ui-bg;
+}
+
+.users__table-filter-title {
+  justify-self: center;
+  margin-bottom: 10px;
```
sorja: vars?

src/client/components/UserListFilters/UserListFilters.scss
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.users__table-filter-item {
+  display: grid;
+  grid-template-columns: 20% auto;
+  align-items: center;
+  grid-column-gap: 10px;
```
sorja: vars?

src/client/components/UserListFilters/UserListFilters.tsx
```diff
@@ -1,5 +1,91 @@
...
+  const [modalOpen, setModalOpen] = useState(false)
+
+  return (
+    <div className="users__table-filter">
+      <div className="users__table-filter-title">
+        <h3>{t('admin.filter')}</h3>
+      </div>
+
+      <div className="users__table-filter-container">
+        <div className="users__table-filter-item">
```
sorja, line 34: can we use DataGrid common component?

src/client/components/UserListFilters/UserListFilters.tsx
```diff
@@ -1,5 +1,91 @@
...
+            countries={countries}
+            excludedRegions={[
+              RegionCode.FE,
+              RegionCode.AT,
+              ...secondaryRegions.regions.map((r: Region) => r.regionCode),
+            ]}
+            headerLabel={t('common.select')}
+            onClose={() => setModalOpen(false)}
+            initialSelection={filters.countries}
+            unselectableCountries={[]}
```
sorja: should we make optional?

src/client/store/ui/userManagement/slice.ts
```diff
@@ -15,6 +15,11 @@ import { UserManagementState } from './stateType'
 const initialState: UserManagementState = {
   user: null,
   users: [],
+  filters: {
+    countries: [],
+    langs: [],
```
sorja: can we name it languages

src/client/store/ui/userManagement/stateType.ts
```diff
@@ -1,6 +1,12 @@
-import { User } from '@meta/user'
+import { CountryIso } from '@meta/area'
+import { RoleName, User } from '@meta/user'
 
 export interface UserManagementState {
   user?: User
   users: Array<User>
+  filters: {
+    countries: Array<CountryIso>
+    langs: Array<string>
```
sorja: can we name it languages

sorja (approved):

src/client/components/UserListFilters/UserListFilters.tsx
rfontanarosa: in another pr

src/client/components/UserList/UserListHeader/UserListHeader.tsx
```diff
@@ -19,7 +21,7 @@ const UserListHeader: React.FC<Props> = ({ isAdmin, roleNames }) => {
         <th className="user-list__header-cell">{t('userManagement.name')}</th>
         {!isAdmin && <th className="user-list__header-cell">{t('userManagement.role')}</th>}
         {isAdmin &&
-          roleNames.map((roleName: RoleName) => (
+          (Objects.isEmpty(filters.roles) ? roleNames : filters.roles).map((roleName: RoleName) => (
```
minotogna: create a hook, useRoleNames that will handle the filter there

src/client/store/ui/userManagement/slice.ts
```diff
@@ -24,6 +28,12 @@ export const userManagementSlice = createSlice({
     setUserToEdit: (state, { payload }) => {
       state.user = payload
     },
+    updateCountryFilter: (state, { payload }) => {
+      state.filters.countries = payload
+    },
+    updateRolesFilter: (state, { payload }) => {
+      state.filters.roles = payload
+    },
```
minotogna: ideally one action is needed, that takes a parameter which is a Partial of UserFilter

minotogna (changes requested):
also, remove those 'placeholder ' from ui. i dont see a reason of having them there

src/client/components/UserList/UserListHeader/UserListHeader.tsx
rfontanarosa: too complicated.

- that role list depends on the view, so it should not be placed into meta/users or meta/userRoles
- if we move the logic above into a hook we can use it to show the filtered list used to build the table columns but not the original one, which is used to show the filters

so this is the only logical way to avoid replicating that list.

minotogna: I dont understand the comment. dont we show the user roles only in the admin page ?

rfontanarosa: yes, that's why at the moment there is a roleNames constant defined at view level, as suggested by you.

but that list it's used to display this
Screenshot 2022-11-09 alle 12 24 26
and this
Screenshot 2022-11-09 alle 12 24 41

the first one is static and depends on the view.

the second one is dynamic and depends on the state (ui.userManagement.filters).

this means that if I move that list into the hook, I have to export it because it is used by the first case or duplicate that definition (inside the view and the hook).

minotogna: Can't you have 2 hooks ? UseRoleNames and useFilteredRileNames ?

rfontanarosa: ok

src/client/components/UserList/AdministrationListElement/AdministrationListElement.tsx
```diff
@@ -7,15 +7,14 @@ import classNames from 'classnames'
 import { ClientRoutes } from '@meta/app'
 import { RoleName, User, UserStatus } from '@meta/user'
 
+import { useFilteredRoleNames } from '@client/store/ui/userManagement/hooks'
```
minotogna: import from. '@client/store/ui/userManagement'

src/client/components/UserList/UserListHeader/UserListHeader.tsx
```diff
@@ -3,14 +3,13 @@ import { useTranslation } from 'react-i18next'
 
 import { RoleName, Users } from '@meta/user'
 
+import { useFilteredRoleNames } from '@client/store/ui/userManagement/hooks'
```
minotogna: here as well

src/client/components/UserListFilters/UserListFilters.tsx
```diff
@@ -1,5 +1,90 @@
-import React from 'react'
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
minotogna: these should not be here.

minotogna (changes requested):

minotogna (approved):

PR #1705, opened 2022-11-14
markdown previewer fix

### LINKED ISSUE #1692: Descriptions not visible in not edit mode
https://user-images.githubusercontent.com/1138156/201063016-d907c160-5e07-4438-8c18-e0687de4342e.mov

rfontanarosa:
close #1692

the problem was caused by the sanitizer plugin. it doesn't allow to put paragraph html tag (apparently without a proper configuration).

There is still an unexpected behaviour with the previewer that occurs when the html is formatted in a certain way. I've tried various plugins trying to fix it without success. Here is the plugin list:

- 'rehype-format'
- 'rehype-parse'
- 'rehype-raw'
- 'rehype-remark'
- 'rehype-sanitize'
- 'rehype-stringify'

https://user-images.githubusercontent.com/10047945/201746515-5028bba8-619a-4cfc-bded-d5225a14589b.mov


sorja:
Can you test xss?
eg.
`<script>alert()</script>`

We don't want to allow external scripts to run; we need to sanitize user input

rfontanarosa:
> 

it seems safe, but if security is the main reason why sanitizer has been used, maybe we have to find an alternative

rfontanarosa:
miro is right. it is not safe:+. the following code is executed:

```
<svg>
<script type="text/javascript">
alert("hello");
</script>
</svg>
```

minotogna (approved):

sorja (approved):

Issue #1675, opened 2022-11-03
Metadata Insert controllers


minotogna:
each method should also persist activity log
each method should be added to integration tests

- [x] AssessmentController.create / update / delete
  - [x] create
  - [x] update
  - [x] delete
- [x] SectionController.create / update / delete
- [x] TableSectionController.create  / update / delete
- [x] TableController.create  / update / delete
- [ ] RowController.create  / update / delete
- [ ] ColController.create  / update / delete

PR #1713, opened 2022-11-16
Assessment Controller for Update

persas:
AssessmentController update in #1675


src/server/controller/assessment/update.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    await ActivityLogRepository.insertActivityLog(
+      {
+        activityLog: {
+          target: updatedAssessment,
+          section: 'assessment',
+          message: ActivityLogMessage.assessmentUpdate,
+          user,
+        },
+        assessment: updatedAssessment,
```
sorja: this should be `assessment` and not `updatedAssessment`

src/server/controller/assessment/update.ts
```diff
@@ -0,0 +1,32 @@
...
+import { BaseProtocol, DB } from '@server/db'
+import { AssessmentRepository } from '@server/repository/assessment/assessment'
+import { ActivityLogRepository } from '@server/repository/public/activityLog'
+
+export const update = async (
+  props: { user: User; assessment: Assessment },
+  client: BaseProtocol = DB
+): Promise<Assessment> => {
+  const { assessment, user } = props
+  // await AssessmentRepository.updateAssessmentSchema({ assessment }, client)
```
sorja: remove unused comments

src/server/repository/assessment/assessment/updateAssessment.ts
```diff
@@ -0,0 +1,24 @@
...
+  client: BaseProtocol = DB
+): Promise<Assessment> => {
+  const { assessment } = params
+
+  const assessmentUpdated = await client.one<Assessment>(
+    `update assessment set props = $1::jsonb where id = $2 returning  *;`,
+    [JSON.stringify(assessment.props), assessment.id],
+    (assessment) => {
+      return { ...Objects.camelize(assessment), metaCache: assessment.metacache }
+    }
```
sorja: I think this is typo, should it be `metaCache: assessment.meta_cache`?
Also, I would support updating `metaCache` like you did for props`, but @minotogna can verify if this is needed

sorja:
Looks good in general, how did you test it was working?
We need to add integration test ;)

src/server/repository/assessment/assessment/updateAssessment.ts
persas: You are right. There is a typo, actually, it should be `assessment.metaCache`

src/server/controller/assessment/update.ts
persas: Are you sure? 

In line 15 the const is defined as `updatedAssessment`. Do you mean to change the line 15 as well?

sorja: Activity log is to keep track of changes
in the assessment we pass the 'old' assessment and the 'target' is the updated one

This way when you are looking at changes, it's easy to see what changed

minotogna:
> Looks good in general, how did you test it was working?
> 
> We need to add integration test ;)

Not now . We need to move to other tasks first

src/server/repository/assessment/assessment/updateAssessment.ts
minotogna: It should be the property of the database table

src/server/controller/assessment/update.ts
minotogna: The repository method name must be renamed to update , that's what @sorja meant

sorja: > The repository method name must be renamed to update , that's what @sorja meant

Thats also a good point though; what I mean is the activity log entries

src/server/repository/adapter/index.ts
```diff
@@ -0,0 +1,3 @@
+import { AssessmentAdapter } from './assessment'
+
+export const Adapter = { AssessmentAdapter }
```
minotogna: export AssessmentAdapter, do not rename it.

src/server/repository/assessment/assessment/updateAssessment.ts
```diff
@@ -0,0 +1,22 @@
...
+export const updateDefaultCycle = async (
+  params: {
+    assessment: Assessment
+  },
+  client: BaseProtocol = DB
+): Promise<Assessment> => {
+  const { assessment } = params
+
+  const assessmentUpdated = await client.one<Assessment>(
+    `update assessment set props = props || '{"defaultCycle": $1~}' where id = $2 returning  *;`,
```
minotogna: we dont want to return the meta_cache here. activity log would be huge

src/test/integration/assessment/updateAssessmentDefaultCycle.ts
```diff
@@ -0,0 +1,22 @@
...
+      cycleName: assessmentCycleName,
+    })
+
+    const user = await UserController.getOne({
+      email: userMockTest.email,
+    })
+
+    assessment.props.defaultCycle = cycle.uuid
+
+    const updatedAssessment = await AssessmentController.updateDefaultCycle({ user, assessment })
```
minotogna: AssessmentController.updateDefaultCycle  should take cycle as parameter: AssessmentController.updateDefaultCycle({ user, assessment, cycle }), where cycle is the cycle we want to update. 
do not set props above.

minotogna (changes requested):

src/server/repository/assessment/assessment/updateAssessment.ts
minotogna: so to fix this problem, the update query should not return anything.
At the end of the method, you simply ```return read({id: assessment.id})```, where read method exist in the same folder of this file

src/server/repository/adapter/index.ts
minotogna: export { AssessmentAdapter } from './assessment'

src/server/repository/adapter/assessment.ts
```diff
@@ -0,0 +1,14 @@
+import { Objects } from '@utils/objects'
+
+import { Assessment, AssessmentProps } from '@meta/assessment'
+
+interface AssessmentDB {
+  id: number
+  uuid: string
+  props: AssessmentProps
+  meta_cache: AssessmentProps
```
minotogna: meta_cache is of AssessmentMetaCache type 
import { Assessment, AssessmentMetaCache , AssessmentProps } from '@meta/assessment'

minotogna (changes requested):
good pr 👍 , just a type fix AssessmentDB

minotogna (approved):

sorja (approved):

Issue #1711, opened 2022-11-15
PanEuropean migration


minotogna:
http://localhost:9000/EU/assessments/panEuropean/2020/home/overview

- [x] setup with pan_european schema migration to _legacy_pan_european
- [x] create assessment and cycles 
- [x] metadata migration
- [x] insert countries and verify pan European entry page
- [x] data migration

PR #1716, opened 2022-11-16
Pan european/migration setup

persas:
Solves step 1 ` setup with pan_european schema migration to _legacy_pan_european` of  #1711


src/server/db/migration/migrations/sqls/20221116105034-pan-european-legacy-up.sql
```diff
@@ -0,0 +1,3 @@
+/* Replace with your SQL commands */
+
+alter schema pan_european RENAME TO _legacy_pan_european;
```
minotogna: you should use if exists

minotogna (changes requested):

minotogna (approved):

sorja (approved):

sorja:
👍

PR #1723, opened 2022-11-17
added specific getAreas permission

rfontanarosa:
anonymous user can get areas **ONLY** if assessmentName and cycleName are not specified
logged user can get areas **EVEN** if assessmentName and cycleName are not specified

### DIFF
--- a/src/server/middleware/auth.ts
+++ b/src/server/middleware/auth.ts
@@ -154,6 +154,13 @@ const requireEditAssessmentFile = async (req: Request, _res: Response, next: Nex
   }
 }
 
+const requireGetAreas = async (req: Request, _res: Response, next: NextFunction) => {
+  const { assessmentName, cycleName } = <Record<string, string>>{ ...req.params, ...req.query, ...req.body }
+  const user = Requests.getRequestUser(req)
+  if (!user) _next(!assessmentName && !cycleName, next)
+  else _next(!!user, next)
+}
+
 export const AuthMiddleware = {
   requireEditDescriptions,
   requireEditTableData,
@@ -165,4 +172,5 @@ export const AuthMiddleware = {
   requireEditUser,
   requireViewUsers,
   requireEditAssessmentFile,
+  requireGetAreas,
 }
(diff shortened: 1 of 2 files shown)


sorja:
should we use express-validation like here
https://github.com/openforis/fra-platform/pull/1677

src/server/middleware/auth.ts
```diff
@@ -154,6 +154,13 @@ const requireEditAssessmentFile = async (req: Request, _res: Response, next: Nex
   }
 }
 
+const requireGetAreas = async (req: Request, _res: Response, next: NextFunction) => {
+  const { assessmentName, cycleName } = <Record<string, string>>{ ...req.params, ...req.query, ...req.body }
+  const user = Requests.getRequestUser(req)
+  if (!user) _next(!assessmentName && !cycleName, next)
+  else _next(!!user, next)
```
minotogna, line 161: When Assessment and cycle are specified , it's needed to check that the user has read view permission to them.

The condition should be if not Assessment and not cycle then _next else requireView

minotogna (changes requested):

minotogna:
> should we use express-validation like here
> 
> https://github.com/openforis/fra-platform/pull/1677

Not really

sorja (approved):

src/server/middleware/auth.ts
rfontanarosa: but requireVIew needs CountryIso which is something not related to this.

what about adding

```
const canViewCycle = (props: { cycle: Cycle; user: User }): boolean => {
  const { cycle, user } = props
  if (cycle.published) return true
  if (Users.isAdministrator(user)) return true

  return false
}
```

and checking

```
const requireGetAreas = async (req: Request, _res: Response, next: NextFunction) => {
  const { assessmentName, cycleName } = <Record<string, string>>{ ...req.params, ...req.query, ...req.body }
  if (!assessmentName && !cycleName) next()
  const user = Requests.getRequestUser(req)
  const { cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName })
  _next(Authorizer.canViewCycle({ cycle, user }), next)
}
```

minotogna: yes, but if cycle is not published, and user is logged then user must have an active role in that cycle to be authorized to view

PR #1731, opened 2022-11-21
migrate metadata

persas:
Solves step `migrate metadata` of #1711 

NOTE: Commented lines 10050 and 10146 of `src/test/sectionSpec/PanEuropeanSpecs.ts`. Otherwise it breaks


sorja:
@minotogna 
Note: PanEuropean production table 4.4a does not load (probably related to the commented lines, rows without cols)

sorja (approved):

minotogna:
@sorja @persas can we have screenshots or video of pan European  page ?
missing tables should be fixed in separate issues then.

minotogna (approved):

PR #1749, opened 2022-11-23
Pan european/migrate data

persas:
PR To solve the area migration of #1711 

NOTE: Data shows in DB but when accessing the endpoint `http://localhost:9000/FE/assessments/paneuropean/2020/home/overview` nothing will show and the server will raise the following error:

```
[ERROR] 09:12:44 QueryResultError: No data returned from the query.
```


minotogna (commented):
I have the impression not only pan European countries/ regions are imported .

persas:
You are right, the region is being properly filtered but not the countries

persas:
Working: `http://localhost:9000/FE/assessments/pan_european/2020/home/overview`

minotogna (approved):
👍

sorja (approved):

PR #1813, opened 2022-12-05
Meta/invitation gmail fix

### LINKED ISSUE #1806: user invitation : check if user has google auth with invited email
Country home -> manage collaborators 

When inviting user , the email must be used to verify if it’s used by a user with google login as well 
If that’s the case , the invitation belongs to that user and email should be sent to his primary email

sorja:
Resolves #1806 

includes region role fix

image

### DIFF
--- a/src/server/controller/user/invite.ts
+++ b/src/server/controller/user/invite.ts
@@ -40,7 +40,6 @@ export const invite = async (
       t
     )
 
-    // TODO
     userToInvite = await UserRepository.getOne({ email }, t)
 
     const { userId, role } = userRole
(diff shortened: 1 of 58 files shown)


rfontanarosa (approved):

src/server/controller/user/invite.ts
```diff
@@ -40,7 +41,8 @@ export const invite = async (
       t
     )
 
-    userToInvite = await UserRepository.getOne({ email }, t)
+    userToInvite =
```
minotogna: is this fetch really needed? isn't the user to invite correctly initialized above?

minotogna: the thing to make sure is that UserRepository.create returns the same object than UserRepository.getOne

minotogna (approved):

PR #1825, opened 2022-12-06
Refactor Section Controller and add remove test

persas:
Refactor of Section Controllers related to [#1675](https://github.com/openforis/fra-platform/issues/1675) .

Also added Remove tests

### DIFF
--- a/src/server/controller/metadata/createSection.ts
+++ b/src/server/controller/metadata/createSection.ts
@@ -5,26 +5,27 @@ import { BaseProtocol, DB } from '@server/db'
 import { SectionRepository } from '@server/repository/assessment/section'
 import { ActivityLogRepository } from '@server/repository/public/activityLog'
 
-export const remove = async (
-  props: { assessment: Assessment; section: Section; user?: User },
+export const createSection = async (
+  props: { user: User; assessment: Assessment; section: Pick<Section, 'props'> },
   client: BaseProtocol = DB
-): Promise<void> => {
-  const { assessment, section, user } = props
+): Promise<Section> => {
+  const { user, assessment, section } = props
 
   return client.tx(async (t) => {
-    const deletedSection = await SectionRepository.remove({ assessment, section }, t)
+    const createdSection = await SectionRepository.create({ section, assessment }, t)
 
     await ActivityLogRepository.insertActivityLog(
       {
         activityLog: {
-          target: { deletedSection },
+          target: createdSection,
           section: 'section',
-          message: ActivityLogMessage.sectionDelete,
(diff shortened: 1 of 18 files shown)


minotogna (commented):
Can you verify delete and update section can return directly *?
I suspect SectionRepos.getOne mighty do something more, if that's the case, you need to return that getOne , or use the same adapter

persas:
Update and remove will work just fine because the filter is being done by `id` that is unique. However, in the getOne we are picking either by name or by id. Not sure if we want to include this functionality in the remove and update. 

There is also another issue I just identified. I think we cannot create or update child sections as it is right now, it will just work for father sections....

minotogna:
> Update and remove will work just fine because the filter is being done by `id` that is unique. However, in the getOne we are picking either by name or by id. Not sure if we want to include this functionality in the remove and update. 
> 
> 
> 
> There is also another issue I just identified. I think we cannot create or update child sections as it is right now, it will just work for father sections....

Can you update the create method by passing an optional parent id ? And test it 😅?

persas:
> > Update and remove will work just fine because the filter is being done by `id` that is unique. However, in the getOne we are picking either by name or by id. Not sure if we want to include this functionality in the remove and update.
> > There is also another issue I just identified. I think we cannot create or update child sections as it is right now, it will just work for father sections....
> 
> Can you update the create method by passing an optional parent id ? And test it 😅?

I was doing it. But given the meta definition at [meta/assessment/section](https://github.com/openforis/fra-platform/blob/7296e9dfdd934964ac997dc3570ad25f70893986/src/meta/assessment/section.ts#L1-L20) I think I would create a new `createSubSection` method instead of granting the option to the `createSection`. What do you think?

minotogna:
> > > Update and remove will work just fine because the filter is being done by `id` that is unique. However, in the getOne we are picking either by name or by id. Not sure if we want to include this functionality in the remove and update.
> 
> > > There is also another issue I just identified. I think we cannot create or update child sections as it is right now, it will just work for father sections....
> 
> > 
> 
> > Can you update the create method by passing an optional parent id ? And test it 😅?
> 
> 
> 
> I was doing it. But given the meta definition at [meta/assessment/section](https://github.com/openforis/fra-platform/blob/7296e9dfdd934964ac997dc3570ad25f70893986/src/meta/assessment/section.ts#L1-L20) I think I would create a new `createSubSection` method instead of granting the option to the `createSection`. What do you think?

Sure

minotogna (approved):

sorja (approved):
🚀

PR #1841, opened 2022-12-08
this fix possible error when sectionName or id is not passed

rfontanarosa:


src/server/repository/assessment/section/getOne.ts
```diff
@@ -12,7 +12,7 @@ export const getOne = async (
 ): Promise<Section> => {
   const schemaName = Schemas.getName(props.assessment)
   return client
-    .one<any>(
+    .oneOrNone<Section>(
```
minotogna, line 15: Is this a possible error or a real error?
This method should not return undefined. What is the specific error?

minotogna (changes requested):

PR #1843, opened 2022-12-08
Geo: Add .geojson file with GEE forest extent estimations 

eraviolo:
To import the .json data into the forest-estimations table run:
`yarn geo:forest-estimations-import "FRA_236_ISO_v9.geojson"`


tools/geo/forestEstimationsImport.ts
```diff
@@ -20,21 +20,30 @@ export const forestIndicatorsImport = async (): Promise<void> => {
   // eslint-disable-next-line no-console
   const args = process.argv.slice(2)
 
-  const path = args[0]
-  const content = await fs.readFile(path)
+  const fileName = args[0]
+
+  const content = await fs.readFile(path.resolve(__dirname, fileName))
```
minotogna: can this just be 
```
  const content = await fs.readFile(path.resolve(__dirname, 'FRA....json'))
```
or there is a need to keep it dynamic ?

eraviolo: Sure, we can omit the name, no need to keep it dynamic

minotogna: cool :D so anyone can just run the script, without worrying about file name to pass

sorja (approved):
👍

minotogna (approved):

PR #1844, opened 2022-12-08
Add 'sectionName: extentOfForest' for ODP write'

sorja:
Adds sectionName: 'extentOfForest' for OriginalDataPoint write operations

### DIFF
--- a/src/client/store/pages/originalDataPoint/actions/updateOriginalDataPoint.ts
+++ b/src/client/store/pages/originalDataPoint/actions/updateOriginalDataPoint.ts
@@ -27,6 +27,7 @@ const putOriginalDataPoint = Functions.debounce(
           countryIso,
           assessmentName,
           cycleName,
+          sectionName: 'extentOfForest',
         },
       }
     )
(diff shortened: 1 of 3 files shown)


src/client/store/pages/originalDataPoint/actions/updateOriginalDataPoint.ts
```diff
@@ -27,6 +27,7 @@ const putOriginalDataPoint = Functions.debounce(
           countryIso,
           assessmentName,
           cycleName,
+          sectionName: 'extentOfForest',
```
minotogna, line 30: don't we need to pass this to getODP as well?

sorja: We use middleware `canView` for read operations which does not take into account the sectionName

https://github.com/openforis/fra-platform/blob/703a7ffeba04efb8a7570680fb4eebfb590ac6a6/src/server/middleware/auth.ts#L46-L45

rfontanarosa (approved):

minotogna (approved):

PR #1856, opened 2022-12-09
Feature/table controller

persas:
Table controllers. Part of #1675


src/test/integration/table/updateTable.ts
```diff
@@ -0,0 +1,46 @@
...
+      assessment,
+      user,
+      table: {
+        ...table,
+        props: {
+          ...table.props,
+          name: 'updated name',
+        },
+      },
+    })
```
sorja, line 36: Tests are good, but I would prefer to have a single place for the mock data used, eg 'mocks/table' or similar

What do you think?

minotogna (approved):

sorja (approved):

PR #1870, opened 2022-12-12
Original data point: Copy previous national classes

### LINKED ISSUE #1858: ODP Edit: Copy previous classes not working (button always disabled)

sorja:
Resolves #1858 

- Fix ODP copy and add rule to enable button:
  * There is previous year
  * No national classes
  * ODP must have a year
- Moves odp reservedYears to state
- Remove `year = null` ODPs from migration

https://user-images.githubusercontent.com/5508251/207075052-4c4d5973-721b-4e21-b4f3-0621795add4e.mov


minotogna:
can we have a video with different test cases ?

minotogna (approved):

rfontanarosa (approved):

PR #1886, opened 2022-12-13
Add mailgun variables

sorja:
Leave fra variables as fallback for localhost

TODO:
Verify why sandbox used, or how to get out of the sandbox.

Data command failed: 421 Domain <censored>.mailgun.org is not allowed to send: Sandbox subdomains are for test purposes only. Please add your own domain or add the address to authorized recipients in Account Settings.


sorja:
Pending answer from support

Looks like free tier is locked into sandbox (at least on heroku)

sorja:
Response to the question:
Mailgun **cannot** be used to send emails outside sandbox on Free Tier?

> That is correct, the Heroku Starter free plan doesn't allow custom domains to be added, only the sandbox domain can be used. In order to add custom domains you will need to upgrade to one of the Heroku paid plans.

PR #1892, opened 2022-12-14
Fix ODP Edit Data sources -> Methods used

### LINKED ISSUE #1873: ODP Methods used / Camerun / 2004. Check odp migration and why if odp property dataSourceMethods is null

sorja:
Resolves #1873

Fixes a bug where multiselect not rendered if value is null (eg. not given previously) Multiselect fails on null values, fixed by defaulting to empty array on null

_**Note** Video shows first the bug and then fixed_

https://user-images.githubusercontent.com/5508251/207545695-3eaaa56f-a2ee-4ff5-b269-bf017e6623a5.mov

### DIFF
--- a/src/client/pages/OriginalDataPoint/components/DataSources/DataSources.tsx
+++ b/src/client/pages/OriginalDataPoint/components/DataSources/DataSources.tsx
@@ -44,7 +44,6 @@ const DataSources: React.FC<Props> = (props) => {
   }
 
   const isDisabled = print || !canEditData || !originalDataPoint.year
-
   return (
     <div className="odp__section">
       {!print && <h3 className="subhead">{i18n.t('nationalDataPoint.dataSources')}</h3>}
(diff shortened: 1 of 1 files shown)


minotogna:
Can u check, I think there's an issue linked to this ?

src/client/pages/OriginalDataPoint/components/DataSources/DataSources.tsx
```diff
@@ -92,7 +91,7 @@ const DataSources: React.FC<Props> = (props) => {
                     disabled={isDisabled}
                     i18n={i18n}
                     localizationPrefix="nationalDataPoint.dataSourceMethodsOptions"
-                    values={originalDataPoint.dataSourceMethods}
+                    values={originalDataPoint.dataSourceMethods ?? []}
```
minotogna, line 94: I suspect it might be safer to use default value in multi select. Other consumers  might face same issue ?

sorja: Makes sense, will default to empty array on null values
Only used in userListFilters

sorja: On closer inspection Multiselect is `deprecated` (old class component) which needs a bigger refactor.
It has fallback to  `const { values = [] } = this.props` which doesn't seem to work in this case

I would apply this fix and refactor later the component itself

minotogna (approved):

rfontanarosa (approved):

PR #1893, opened 2022-12-14
Hotfix delete national correspondant

### LINKED ISSUE #1868: Delete national correspondant as admin crashes

rfontanarosa:
close #1868

https://user-images.githubusercontent.com/10047945/207641749-d7f34d82-5ae0-4603-a7dd-2adce1235656.mov


sorja (approved):

src/i18n/resources/en.js
```diff
@@ -1062,7 +1062,8 @@ The FRA team
     picture1MbMax: 'Profile picture cannot exceed 1MB',
     activated: 'Activated',
     status: 'Status',
-    adminConfirm: 'Are you sure you want to promote this user to Administrator?',
+    demoteToUser: 'Are you sure you want to demote this user to User?',
```
minotogna: I would say:
Are you sure you want to remove Admin privileges?

src/server/api/user/index.ts
```diff
@@ -30,6 +32,9 @@ export const UserApi = {
     express.get(ApiEndPoint.User.invitationAccept(), acceptInvitation)
     express.get(ApiEndPoint.User.invitationSendEmail(), AuthMiddleware.requireEditUser, sendInvitationEmail)
 
+    express.post(ApiEndPoint.User.demote(), AuthMiddleware.requireAdmin, demoteToUser)
+    express.post(ApiEndPoint.User.promote(), AuthMiddleware.requireAdmin, promoteToAdmin)
```
minotogna: why the need of these two endpoints? couldn't you call updateUserRoles only ?

src/i18n/resources/en.js
minotogna: and below Are you sure you want to grant Admin privileges?

src/server/api/user/index.ts
rfontanarosa: because I would like to remove, in the near future, the admin role and manage this information in a different way because the management of the userroles begins to be complicated

minotogna (approved):

PR #1894, opened 2022-12-15
GEO: Add functionality to be able to filter the agreement layer

### LINKED ISSUE #1741: GEO: Add functionality to be able to filter the agreement layer
(the boxes on the mockup). Minimum level is one (only one layer says there is forest), maximum is number of layers selected (all the layers agree there is forest). This call to filter the agreement layer is the same as to load the layer but with a different agreement level as param

joonasnuutinen:
Resolves #1741 

https://user-images.githubusercontent.com/18427577/207776614-16bc164a-2cb3-4a09-a9e2-efa846bc815a.mov


src/client/pages/Geo/GeoMap/GeoMapMenuData/MapVisualizerAgreementLevelsControl/MapVisualizerAgreementLevelsControl.tsx
```diff
@@ -1,85 +1,111 @@
...
+    // Otherwise, fetch the new agreement layer and add it to the map.
+
+    const layerQuery = forestOptions.selected.map(({ key }) => `&layer=${key}`).join('')
+    const agreementLevelQuery = `&gteAgreementLevel=${forestOptions.agreementLevel}`
+    const hansenQuery = forestOptions.selected.some(({ key }) => key === ForestSource.Hansen)
+      ? `&gteHansenTreeCoverPerc=${forestOptions.hansenPercentage}`
+      : ''
+    const uri = `/api/geo/layers/forestAgreement/?countryIso=FIN${layerQuery}${agreementLevelQuery}${hansenQuery}`
+
+    // TODO: cache the mapId to reduce server calls?
```
minotogna, line 57: any chance we can use(or reuse) redux? we can discuss this in our next meeting, now we can even start merging this. what do you think?

eraviolo: Regarding layers already requested I think it is a good idea to cache the ids, [here](https://gis.stackexchange.com/questions/435252/expiration-time-of-temporary-url-returned-by-ee-data-gettileurlimage-getmapid) they say they're generally good for about 24 hours

eraviolo (approved):
Looks good! As we spoke on last meeting maybe we have to think a way of hiding all the input layers when agreement layer is checked.

src/client/pages/Geo/GeoMap/GeoMapMenuData/MapVisualizerAgreementLevelsControl/MapVisualizerAgreementLevelsControl.tsx
joonasnuutinen: @minotogna Yes, that's what I thought. We have a simple Redux-based cache for the other layers (`forestOptions.fetchedLayers`), and a similar one could be used here.

@eraviolo That's an interesting point. Could the ids be even cached server-side? Though I don't know about the reliability because the persistence is not guaranteed 🤔

But yeah, let's talk about this in a future meeting. I don't think it's part of the scope of this PR anyway.

minotogna (approved):

PR #1897, opened 2022-12-15
GEO: camelize forest-estimations properties

eraviolo:


sorja (approved):

tools/geo/forestEstimationsImport.ts
```diff
@@ -23,8 +69,8 @@ export const forestIndicatorsImport = async (): Promise<void> => {
 
   const values = response.features.map((d) => {
     const { properties } = d
-    const { iso3, ...data } = properties
-    return { country_iso: iso3, year: 2020, data }
+    const { iso3, year, ...data } = properties
+    return { country_iso: iso3, year, data }
```
minotogna: we could even cameilze the properties here, before importing them. so that in our DB the properties will be already of correct type. what do you think?

eraviolo: Sure, do you mean jsonb data column with already camelized properties names right?
`return { country_iso: iso3, year, data: Objects.camelize(data) }`

minotogna: yes ☺️

minotogna (approved):

PR #1899, opened 2022-12-15
PR - Update assessment status as NC error

### LINKED ISSUE #1913: Update assessment status as NC error
To replicate this: try to change an Assessment status with a no-administrator user.

rfontanarosa:
close #1913


src/meta/user/authorizer.ts
```diff
@@ -113,8 +113,53 @@ const canEdit = (props: {
...
+ * @param props.assessment
+ * @param props.countryIso
+ * @param props.cycle
+ * @param props.table
+ * @param props.status
+ * @param props.user
+ * @returns boolean
+ */
+const canEditAssessmentStatus = (props: {
+  countryIso: CountryIso
```
minotogna: Herr countryiso can be removed and taken from country

src/meta/user/authorizer.ts
```diff
@@ -113,8 +113,53 @@ const canEdit = (props: {
...
+
+  return false
+}
+
 export const Authorizer = {
   canView,
   canViewUsers,
-  canEdit,
+  canEditAssessmentStatus,
+  canEditSections,
```
minotogna: Can this be rolled back to canEdit ?

rfontanarosa: ok

src/meta/user/authorizer.ts
rfontanarosa: done

minotogna:
Can you rename the Pr please?

src/meta/user/authorizer.ts
```diff
@@ -113,8 +112,46 @@ const canEdit = (props: {
...
+    (Users.isNationalCorrespondent(user, countryIso, cycle) ||
+      Users.isAlternateNationalCorrespondent(user, countryIso, cycle)) &&
+    status === AssessmentStatus.editing
+  )
+    return true
+
+  if (Users.isReviewer(user, countryIso, cycle)) {
+    return [AssessmentStatus.editing, AssessmentStatus.review].includes(status)
+  }
+
```
minotogna: Don't we have A centralized place
To check status allowance?

src/server/api/area/index.ts
```diff
@@ -10,7 +10,7 @@ import { postCountry } from './postCountry'
 export const AreaApi = {
   init: (express: Express): void => {
     // Country
-    express.post(ApiEndPoint.Area.country(), AuthMiddleware.requireEditTableData, postCountry)
+    express.post(ApiEndPoint.Area.country(), AuthMiddleware.requireEditAssessmentStatus, postCountry)
```
minotogna: This end point should also be specific to edit assessment status.
It should be something like countrySstatus

rfontanarosa: that endpoint allows to edit all the country properties, not only the status, so probably requireEditAssessmentStatus should be renamed into requireEditCountryAssessment

src/meta/user/authorizer.ts
rfontanarosa: nope, we don't :( we have getRolesAllowedToEdit but it's not related to status

sorja (approved):

src/meta/user/authorizer.ts
minotogna: Can't we get status allowed to edit and then check if the status passed to the middleware is included in those ?

src/server/api/area/index.ts
minotogna: canEditCountryProps and requireEditCountryProps ?
What do you think ?

src/meta/user/authorizer.ts
rfontanarosa: mmmm, can we do that in another pr? at the moment the status check is used by canEditData and canEditCountryProps and the logic is "check the role then check the status"
we have to reverse that logic "get the roles allowed by the status then check the role"

minotogna (approved):

src/meta/user/authorizer.ts
minotogna: sure, let's talk in the next daily

PR #1933, opened 2022-12-29
BUG - missing checks (solves reload page)

rfontanarosa:


src/client/pages/Country/Country.tsx
```diff
@@ -68,6 +70,8 @@ const Country: React.FC = () => {
     }
   }, [countryIso, assessmentName, cycleName, user, dispatch])
 
+  if (!country || !sections) return null
```
minotogna: Why should country page not be rendered with no sections?

minotogna (changes requested):

minotogna (approved):

sorja (approved):

PR #1944, opened 2023-01-03
PR - Errors in message board

### LINKED ISSUE #1916: Errors in message board
- [x] can't send chat messages
- [x] collaborator can't see chat messages

https://user-images.githubusercontent.com/99216009/208472329-114b8dff-5f9f-484e-86d4-af8319dfc874.mp4

Issue

Hi Mino @minotogna! 
There are two issues regarding the message board for now. One as shown in the video NC cannot send one-to-one messages to others although the message to all is good. Another presented in the pic is the collaborator(please double check the code for the viewer) cannot see others in this case NC in the one-to-one message session. Please help to solve these two issues, thanks!

rfontanarosa:
close #1916

https://user-images.githubusercontent.com/10047945/210337687-7bd9aad7-4ebb-4e2a-80d8-48b45f79d080.mov


src/server/repository/public/activityLog/getCycleDataActivities.ts
```diff
@@ -21,6 +21,8 @@ const acceptedMessages = [
   .map((s) => `'${s}'`)
   .join(', ')
 
+const unacceptedSections = ['chat'].map((s) => `'${s}'`).join(', ')
```
sorja: can we avoid this

rfontanarosa: no, the requirement is that that activity must be persisted, so it should be filtered backend side and this query return only the cycledata activities.

sorja (approved):

src/client/components/MessageCenter/Topic/Topic.tsx
```diff
@@ -36,7 +36,7 @@ const Topic: React.FC<TopicProps> = (props) => {
   const user = useUser()
 
   let sectionName = useParams<{ sectionName: string }>()?.sectionName
-  if (topic.type === MessageTopicType.messageBoard) sectionName = 'messageBoard'
+  if (topic.type !== MessageTopicType.review) sectionName = topic.type
```
minotogna, line 39: Why do we set sectionName here ? Isn't it added to the topic instance ?

src/client/pages/AssessmentHome/FraHome/FraHome.tsx
```diff
@@ -33,7 +33,7 @@ const FraHome: React.FC = () => {
   const displayTabs = sections.length > 1 && Areas.isISOCountry(countryIso)
 
   useEffect(() => {
-    if (Users.getRolesAllowedToEdit({ user, countryIso, cycle }).length > 0) {
+    if (Authorizer.canViewUsers({ countryIso, assessment, cycle, user })) {
```
minotogna: Better to have a common hook in home . useGetUsers to use in fra and pan European home

src/client/components/MessageCenter/Topic/Topic.tsx
rfontanarosa: no, it is not. the topic instance doesn't contain the sectionName.
sectionName is used only to save the activity in the log. When it is not specified it is based on MessageTopicType.

src/client/pages/AssessmentHome/PanEuropeanHome/PanEuropeanHome.tsx
```diff
@@ -30,7 +30,7 @@ const PanEuropeanHome = () => {
   const displayTabs = sections.length > 1 && Areas.isISOCountry(countryIso)
 
   useEffect(() => {
```
minotogna: since you are calling the hook in AssessmentHome, this fetch is redundant then ? can u remove it from here and test pan European ?

rfontanarosa: done

minotogna (approved):

PR #2001, opened 2023-01-12
Resolve redirect issue when user is logged

### LINKED ISSUE #1986: Open 2025 section page with direct link blank page or redirect
User is admin.

- [x] Open page https://fra-data.review.fao.org/assessments/fra/2025/FIN/home/overview, system redirects to home
- [x] https://fra-data.review.fao.org/assessments/fra/2025/FIN/extentOfForest/ gives error 
image

sorja:
Resolves #1986

- Fix issue where we didn't wait for the user to be populated
- Fixed by adding a hook to check if the app is initialised (currently only user is loaded on app init)

https://user-images.githubusercontent.com/5508251/212030142-ba01b233-c046-434c-a068-0041aecfd410.mov


rfontanarosa (approved):

src/client/pages/Country/Country.tsx
```diff
@@ -71,7 +72,8 @@ const Country: React.FC = () => {
 
   if (!countryIso) return null
 
-  if (!Authorizer.canView({ countryIso, assessment, cycle, user })) window.location.href = ClientRoutes.Root.path
+  if (isAppInitialized && !Authorizer.canView({ countryIso, assessment, cycle, user }))
```
minotogna: I think that if App is not initialized we shouldn't even reach the rendering at this point .

Better to return null in PageRoutes if app is not initialized .

minotogna (approved):

PR #2010, opened 2023-01-13
Table 4a calculation issues

### LINKED ISSUE #1988: Issues- Table 4
Table 4a:

1. Add warnings if a subcategory surpasses the parent category in case only one or two subcategory data are available (currently **only** these three rows are filled and sum exceeds the parent the warning will pop up)
2. _Unknow(d)_ is auto-filled with negative values, double check
3. _Total (a+b+c+d)_, if I understand right, should equal _Total forest area_(table 1a). 

Ps:  _Unknow(d)_  and   _Total (a+b+c+d)_ can be updated if re-enter data in one of the three parent categories

sorja:
Resolves #1988 

Fix issues in table 4a:
- Unknown is 0 if smaller than 0 from calculation (Math.max(0, calculation)
- Check if 1 or more are larger than parent in private ownership

Fix generic:
- calculations failed if result was 0, fixed (returned null)

No changes to table 4b


minotogna (dismissed):

sorja:
Moved to draft until discussion resolved for 4b

rfontanarosa (dismissed):

minotogna (approved):

rfontanarosa (approved):

Issue #2024, opened 2023-01-16
User Edit Form


minotogna:
[Contact_info_module_23Dec2022.xlsx](https://github.com/openforis/fra-platform/files/10426363/Contact_info_module_23Dec2022.xlsx)

- [x] BE structure and migration (user)
- [x] BE structure and migration (user role)
- [x] invitation (prefill role props - country and primary email)
- [x] accept invitation and update Users.isPersonalInfoRequired
- [x] edit basic info
- [ ] edit specific info

PR #2076, opened 2023-01-24
PR - User Edit Form (Incomplete but working)

rfontanarosa:
part of #2024

What have been done:

- Changed User type
- User migration
- Moved name, lang, institution and position under user.props
- used user.props.name and user.props.lang where necessary
- Full working UserList
- Full working UserForm (now supports user property edit)


src/meta/user/user.ts
```diff
@@ -6,15 +6,23 @@ export enum UserStatus {
   inactive = 'inactive',
 }
 
+export type UserProps = {
+  // old fields
+  name: string
+  lang: string
+  institution?: string
+  position?: string
```
minotogna, line 14: to remove. in migration they will become role props (organization, position)

src/meta/user/userRole.ts
```diff
@@ -9,7 +9,30 @@ export enum RoleName {
   VIEWER = 'VIEWER',
 }
 
-export interface UserRole<N extends RoleName, P = undefined> {
+export type UserRoleProps = {
+  userInfo: {
```
minotogna, line 13: remove userInfo object. create hierarchy

src/meta/user/user.ts
rfontanarosa: ok, so they will be replicated for every role of the user. but position will be Professional title / Titre professionnelle / Título professional

rfontanarosa: next pr

minotogna (commented):
Pr too big. Can we split first BE changes and then FE ?

PR #2096, opened 2023-01-26
BE structure and migration (user)

rfontanarosa:
part of #2024


src/server/controller/user/invite.ts
```diff
@@ -28,7 +28,8 @@ export const invite = async (
     // If user with primary email not found, check if user has active google login
     if (!userToInvite) userToInvite = await UserRepository.getOne({ emailGoogle: email }, t)
     // If neither of above, create new user
-    if (!userToInvite) userToInvite = await UserRepository.create({ user: { email, name: name ?? '' } }, client)
+    if (!userToInvite)
+      userToInvite = await UserRepository.create({ user: { email, props: { name: name ?? '' } } }, client)
```
minotogna, line 32: image

tools/dataMigration/migrateUsers.ts
```diff
@@ -42,19 +42,15 @@ export const migrateUsers = async (props: { client: BaseProtocol }): Promise<voi
...
                u_l.profile_picture_filename,
-               u_l.name,
                case
                    when u_l.active then 'active'::users_status
                    else 'invitationPending'::users_status
                    end as status,
-               u_l.profile_picture_file,
-               u_l.position,
-               u_l.email
+               jsonb_build_object('name', u_l.name, 'institution', u_l.institution, 'lang', u_l.lang, 'position', u_l.position) as props
```
minotogna: do not include institution and position

tools/dataMigration/migrateUsersRole.ts
```diff
@@ -76,6 +76,7 @@ export const migrateUsersRole = async (props: Props): Promise<void> => {
                    else '{}'::jsonb
                    end
                                                                                                           as roles,
+               '{}'::jsonb                                                                                as info,
```
minotogna: this will be needed in the next pr, and use column names props and permissions

src/meta/user/users.ts
```diff
@@ -73,18 +73,18 @@ export const profilePictureUri = (userId: number) => ApiEndPoint.User.profilePic
 export const validProfilePicture = (file: File) => !file || file.size <= 1000000
 
 // validation methods
-export const validName = (user: Partial<User>) => !Objects.isEmpty(user.name)
-export const validRole = (user: Partial<User>) => !Objects.isEmpty(user.roles)
+export const validName = (value: string) => !Objects.isEmpty(value)
+export const validRole = (value: string) => !Objects.isEmpty(value)
```
minotogna: are they used elsewhere ? they should take as input the user.

tools/dataMigration/migrateUsersRole.ts
rfontanarosa: my bad, wrong file

tools/dataMigration/migrateUsers.ts
rfontanarosa: done

src/server/controller/user/invite.ts
rfontanarosa: solved. lang is mandatory and 'en' by default

src/meta/user/users.ts
rfontanarosa: reverted, that way was useful for the front-end part but we can handle it later

rfontanarosa: I have also seen that validEmail is replicated somewhere else. that should be optimized and centralized in another pr. I've noted it.

minotogna (approved):

PR #2102, opened 2023-01-27
Feature/2089 geo preset recipes

### LINKED ISSUE #2089: Geo: recipes
Add select above layers selection with pre calculated recipes

joonasnuutinen:
Resolves #2089 

https://user-images.githubusercontent.com/18427577/215020133-b0c7a6c1-6a2c-461f-a896-2a1a558412dd.mov


minotogna (commented):
I've seen you dispatch several actions at the same time .
This will cause x times store update, therefore x times ui rerenders .
To avoid multiple rerenders you could dispatch one action , and update several keys to any slice of the reducer , causing only one rerender.
Would that work ? Or what do you think ?

sorja:
> I've seen you dispatch several actions at the same time . This will cause x times store update, therefore x times ui rerenders . To avoid multiple rerenders you could dispatch one action , and update several keys to any slice of the reducer , causing only one rerender. Would that work ? Or what do you think ?

This is preferable solution.
In cases where you can't avoid dispatching multiple actions, you can try to use batch which is directly supported by redux toolkit:
https://react-redux.js.org/api/batch

joonasnuutinen:
Thanks @minotogna and @sorja for the comments. Yes, I see your point and agree. I'll find a way to combine the actions so that they don't trigger multiple rerenders.

joonasnuutinen:
@minotogna I combined the dispatches now such that they only render once. Some were combined to slices and others managed with batches.

minotogna (approved):

eraviolo (approved):

PR #2104, opened 2023-01-27
GEO: country bounds to center map

eraviolo:
- Added .geojson file with bounds and centroid info (processed on  gee) and import task
- Added bounds table on geo scheme to store the info
- Added endpoint to retrieve bounds and center per country as required in google map component, ie:
req: /api/geo/bounds/?countryIso=ITA
res: 
{"countryIso":"ITA","data":{"bounds":{"east":18.520439271605508,"west":6.62683055960604,"north":47.091576289560685,"south":35.49361944228319},"centroid":{"lat":42.71019681665934,"lng":12.20598871165744}}}
- To create table and import data run:
yarn run-migrations
yarn geo:bounds-import


src/meta/api/endpoint/ApiEndPoint.ts
```diff
@@ -108,6 +108,7 @@ export const ApiEndPoint = {
       forest: () => apiPath('geo', 'estimations', 'forest'),
       forestAgreement: () => apiPath('geo', 'estimation', 'forestAgreement'),
     },
+    bounds: () => apiPath('geo', 'bounds'),
```
sorja: better to move it before sepalProxy, so we can keep it tidy :)
(Have Geo.<path> before Geo.<Key>.<path>)

src/server/api/geo/getBounds.ts
```diff
@@ -0,0 +1,15 @@
+import { Request, Response } from 'express'
+
+import { CountryIso } from '@meta/area'
+
+import { GeoController } from '@server/controller/geo'
+import Requests from '@server/utils/requests'
+
+export const getBounds = async (req: Request<never, never, never, { countryIso: CountryIso }>, res: Response) => {
```
sorja, line 8: Do we have a request type already for this?
I don't remember, if you could check

(Request with only countryIso)

sorja (dismissed):
Left some minor comments, but looks good!

src/meta/api/endpoint/ApiEndPoint.ts
eraviolo: Sure :)

src/server/api/geo/getBounds.ts
eraviolo: I didn't see one with only countryIso, should I add it under geo request types?

minotogna:
@yaguzmang can you post a video pls ?

src/server/api/geo/getBounds.ts
sorja: Sure, we can move it to higher level if it is used elsewhere 👍

sorja (approved):

minotogna (approved):

PR #2108, opened 2023-01-30
Feat/pan eu migration

persas:
Closes to #2009


src/client/pages/Geo/GeoMap/GeoMapMenuRecipes/index.ts
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-line import/no-unresolved
```
minotogna: This doesn’t look good , maybe better to rename the component itself in upper case ?

src/i18n/resources/en.js
```diff
@@ -2010,5 +2010,257 @@ The FRA team
       no19_area_available_for_public_recreation: 'Area available for public recreation 19',
       no20_area_available_for_public_recreation: 'Area available for public recreation 20',
     },
+
+    // Country Comments
+    countryComments: {
```
minotogna: Better to add country comments translations under panEuropean file with prefix panEuropean and import them from countryComments file

src/client/pages/Geo/GeoMap/GeoMapMenuRecipes/index.ts
persas: Yes, I don't know why the precommit hook was complaining about this. I haven't created this, I didn't want to modify it. But will do it.

minotogna: Yes I know you didn't write this component, but let's fix it anyway

src/i18n/resources/en.js
persas: I didn't catch what you exactly mean. Will see now in the daily

tools/dataMigration/migrateData/_repos.ts
```diff
@@ -72,4 +72,59 @@ export const isBasicTable = (tableName: string): boolean =>
     'biomassStockTotal',
     'carbonStockAvg',
     'carbonStockTotal',
+    'country_comments_1_1_1',
```
sorja, line 75: can this be moved to 'panEuropeanTables' and extended here like '...panEuropeanTables'

minotogna: This is just temporary

sorja (dismissed):

sorja (approved):

minotogna (approved):

PR #2114, opened 2023-02-01
User Edit Form (prefill props)

rfontanarosa:
part of #2024


src/server/controller/user/invite.ts
```diff
@@ -42,6 +42,14 @@ export const invite = async (
       t
     )
 
+    await UserRoleRepository.updateProps(
```
minotogna: insert the props within the insert query

minotogna (approved):

PR #2122, opened 2023-02-02
postDataMigration: disable validate nodes

minotogna:
part of #2120

### DIFF
--- a/src/server/repository/adapter/col.ts
+++ b/src/server/repository/adapter/col.ts
@@ -9,11 +9,12 @@ export interface ColDB {
   row_id: number
 }
 
-export const ColAdapter = (
-  { props: { labels, style, variableNo, calculateFn, validateFns, ...rest }, ...col }: ColDB,
-  _?: unknown,
-  __?: unknown
-): Col => {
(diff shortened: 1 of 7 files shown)


src/server/repository/adapter/col.ts
```diff
@@ -9,11 +9,12 @@ export interface ColDB {
...
-  { props: { labels, style, variableNo, calculateFn, validateFns, ...rest }, ...col }: ColDB,
-  _?: unknown,
-  __?: unknown
-): Col => {
+export const ColAdapter = (colDB: ColDB): Col => {
+  const {
+    props: { labels, style, variableNo, calculateFn, validateFns, ...rest },
+    ...col
+  } = colDB
+
```
sorja, line 17: _ and __ was added to support `client.map` and `arr.map`; they had different second param type , does it work without?
this problem was first found in usage of `Objects.map`

sorja (approved):

src/server/repository/adapter/col.ts
minotogna: this change doesn't give any problem, from what I saw

rfontanarosa (approved):

PR #2125, opened 2023-02-03
Fix ui validations

sorja:


src/client/pages/AssessmentSection/DataTable/Table/TableBody/TableBody.tsx
```diff
@@ -13,10 +30,115 @@ type Props = {
   data: TableData
 }
 
+type UseValidateProps = {
+  table: Table
+  rowsData: TypeRow[]
+}
+// ===================
+
+const useValidations = (props: UseValidateProps) => {
```
minotogna: this hook looks nice. I would move it in a separate file, under subfolder  hooks.
question, can we use this hook in Table, not TableBody?

src/client/store/pages/assessmentSection/slice.ts
```diff
@@ -43,6 +43,14 @@ export const assessmentSectionSlice = createSlice({
...
     },
+    setNodeValidations: (state, action: PayloadAction<{ nodeUpdates: NodeUpdates }>) => {
+      const { nodeUpdates } = action.payload
+      const { countryIso, nodes } = nodeUpdates
+
+      nodes.forEach(({ tableName, variableName, colName, value }) => {
+        state.data = TableDatas.updateDatum({ data: state.data, countryIso, tableName, variableName, colName, value })
+      })
+    },
     setNodeValidation: (
```
minotogna: we should find better name for the setNodeValidation... maybe setNodeValidationToDisplay ?

src/client/pages/AssessmentSection/DataTable/Table/TableBody/TableBody.tsx
```diff
@@ -13,10 +30,115 @@ type Props = {
...
+
+          rowsData.forEach((row) => {
+            const { variableName } = row.props
+            const { cols } = row
+            cols.forEach((col) => {
+              const { colName } = col.props
+              if (!colName) return
+              const validateFns = col.props.validateFns?.[cycle.uuid] ?? row.props.validateFns?.[cycle.uuid]
+              if (!validateFns) return
+              const metaCache: AssessmentMetaCache = {
```
minotogna: this is created every time and for each row.
Maybe add this to a local state variable or memo, that gets updated when table prop changes? what do you think?

minotogna (approved):

PR #2145, opened 2023-02-05
Biomass stock calculator per cycle

sorja:
Added files for: en, es, fr

Change url from query to params:
- (?assessmentName=fra...) to .../fra/2025/FIN/...

Original files (for translation reference)

[BiomassCalculatorFRA2025 (1).zip](https://github.com/openforis/fra-platform/files/10660571/BiomassCalculatorFRA2025.1.zip)


rfontanarosa (approved):

src/meta/api/endpoint/ApiEndPoint.ts
```diff
@@ -62,7 +62,15 @@ export const ApiEndPoint = {
...
-    biomassStock: () => apiPath('file', 'biomass-stock'),
+    biomassStock: ({
+      assessmentName = ':assessmentName',
+      cycleName = ':cycleName',
+      countryIso = ':countryIso',
+      sectionName = ':sectionName',
+      language = ':language',
+      selectedDomain = ':selectedDomain',
+    }) =>
+      apiPath('file', 'biomass-stock', assessmentName, cycleName, countryIso, sectionName, selectedDomain, language),
```
minotogna, line 73: why this?

sorja: so the url can be of format
`/file/biomass-stock/fra/2025/FIN/carbonStock/tropical/en`
instead of 
`/file/biomass-stock?assessmentName=fra&cycleName=2025&countryIso=FIN...............................`

when referring to files I think it looks cleaner in URL and not query params, thoughts?

minotogna: but I thought you proposed to use query params as convention.. that's why I'm confused .. why for files different convention ?

sorja: This is a 'link' the user can see (comparable to platform links)

API calls are invisibile for users

minotogna (approved):

PR #2147, opened 2023-02-06
Missing information toaster

### LINKED ISSUE #2129: Login alert
@minotogna 
People for the first login would subconsciously click the home without filling in the info for INSTITUTION and POSITION. Raising an alert in case it happens.

a

sorja:
Resolves #2129 

Screenshot 2023-02-07 at 11 28 58


rfontanarosa (dismissed):

minotogna (changes requested):
the risk is that user can also click somewhere else.
to fix this, we need the redirect to work in whatever page user clicks.

sorja:
> the risk is that user can also click somewhere else. to fix this, we need the redirect to work in whatever page user clicks.

I am unable to redirect anywhere else, what are the options?
Whatever page I open, I get redirected here

rfontanarosa (approved):

minotogna (approved):

PR #2164, opened 2023-02-07
Start user ids from 1 million

sorja:
After migration, user ids start from 1 million to avoid conflicts

### DIFF
--- a/tools/dataMigration/migrateUsers.ts
+++ b/tools/dataMigration/migrateUsers.ts
@@ -42,8 +42,7 @@ export const migrateUsers = async (props: { client: BaseProtocol }): Promise<voi
         delete
         from users;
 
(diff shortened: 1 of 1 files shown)


tools/dataMigration/migrateUsers.ts
```diff
@@ -54,7 +53,9 @@ export const migrateUsers = async (props: { client: BaseProtocol }): Promise<voi
                    end as status,
                u_l.profile_picture_file,
                u_l.position,
-               u_l.email
+               u_l.email,
+               -- Avoid possible conflicts with ids from legacy database
+               (u_l.id + 1000000) as id
```
minotogna, line 58: I really don't like this .. sorry 😣

sorja: we don't have to do it
should we still include the id in the migration though?

PR #2205, opened 2023-02-12
2137 reload layers when country changes

yaguzmang:
Implements #2137 by:

- Migrating the logic that handles changes on the layers from the MapVisualizerPanel and MapVisualizerAgreementsLevelsControl. Reason: we will now move away from the hardcoded 'FIN' countryISO and will support countryISO changes. Those changes can happen at any point and are independent of the mentioned components being rendered or not, hence the migration of the logic to its own hook. The new hook is called at the GeoMap level, where the map is rendered. 
- Adding two more states in the ForestOptions: we previously had the state "fetchedLayers", which was used to determine whether a given layer was cached or not. Now, there are two more states "pendingLayers" and "failedLayers". Those are necessary to avoid infinite calls to the API (e.g a layer is being fetched, so it is not in "fetchedLayers" and then it will be attempted to fetch it again until it is in "fetchedLayers". Problem: the layer could be loading or the previous attempt has failed). 
The previous system worked because there weren't multiple layers being fetched at once, but now that's not the case, when the countryISO changes, all selected layers are fetched for the new country.
The two new states also open the possibility to add loading or failed status in the UI, giving more feedback to the user.

Bug fix: 

- Added a default center to the configuration of the map. This is because the map will not load unless it has a default center or the function 'panTo' is called to a location. It was the case that sometimes the API call that fetches the center of a country would take a couple of seconds, and since the map did not have a center, it resulted in the UI just showing a grey box.

iso change


joonasnuutinen (approved):
I tested it with various countries and layer configurations and tried to break it but couldn't. Everything seems to be working as expected. Good job!

minotogna (approved):

PR #2249, opened 2023-02-16
Use for loop instead of Promise.all for calculations

sorja:


minotogna (dismissed):

minotogna:
@sorja updating calculations client side after estimation fulfilled will be done in a separate pr ?

sorja:
> @sorja updating calculations client side after estimation fulfilled will be done in a separate pr ?

Added here

minotogna (approved):

PR #2252, opened 2023-02-17
fixed metadata migration

rfontanarosa:
metadata are not migrated, something breaks the migration.

after investigation we found that this line causes the trouble.

this is just a temporary fix.


sorja (approved):

minotogna:
can u explain this?

minotogna (approved):

PR #2253, opened 2023-02-17
Update validations 5b

### LINKED ISSUE #2221: 5b validation
image

`Total land area affected by fire` should be checked against country total land area (see table 1a 

Total land area (c))

image

`…of which on forest` should be checked against country Forest area (see table 1a)

sorja:


minotogna (commented):
this changes applies to both cycles. is that what we want?

minotogna (approved):

rfontanarosa (approved):

PR #2279, opened 2023-02-20
Fix calculations for columns

sorja:
Fix calculations when there is column level dependancy
- [x] FE
- [x] BE

### DIFF
--- a/src/server/controller/cycleData/updateDependencies/updateCalculationDependencies/updateCalculationDependencies.ts
+++ b/src/server/controller/cycleData/updateDependencies/updateCalculationDependencies/updateCalculationDependencies.ts
@@ -73,22 +73,23 @@ export const updateCalculationDependencies = async (
           )
         }
       } else {
-        for (let i = 0; i < row.cols.length; i += 1) {
-          const col = row.cols[i]
-          if (col.props.calculateFn?.[cycle.uuid]) {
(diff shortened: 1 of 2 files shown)


src/server/controller/cycleData/updateDependencies/updateCalculationDependencies/updateCalculationDependencies.ts
```diff
@@ -21,7 +21,7 @@ export const updateCalculationDependencies = async (
     { assessment, cycle, variableName, tableName, colName, countryIso, isODP, type: 'calculations' },
     client
   )
-  const visitedVariables: Array<VariableCache> = [{ variableName, tableName }]
+  const visitedVariables: Array<VariableCache> = []
```
sorja, line 24: # The problem
```
const visitedVariables: Array<VariableCache> = [{ variableName, tableName }]
```

In the video you can see that the change "lags" behind
The current row is not evaluated, but all other rows are.

https://user-images.githubusercontent.com/5508251/220076753-0a981d2a-f5df-4804-b1b0-060b04621e09.mov

sorja: # The resolution
```
const visitedVariables: Array<VariableCache> = []
```

Still WIP

PR #2327, opened 2023-02-27
Feat/pan eu validations

### LINKED ISSUE #2008: PanEuropean: data validations

persas:
Closes #2008


src/test/migrations/steps/template.ts
```diff
@@ -1,5 +1,5 @@
-import { BaseProtocol } from '@server/db'
-
-export default async (client: BaseProtocol) => {
-  // DB Step
-}
+// import { BaseProtocol } from '@server/db'
+//
```
sorja: this file doesnt need to be changed

persas: Pre-commit hooks were failing because of "never used". That's why I commented out everything.

sorja: In merge commits you can use --no-verify
👍

src/test/migrations/index.ts
```diff
@@ -25,6 +25,7 @@ describe(`Migrations:`, () => {
         .filter((migration: string) => !ranMigrations.includes(migration))
         .sort()
         .map(async (file: string) => {
+          // eslint-disable-next-line no-console
```
minotogna: Can we use Logger here instead of console.log ?

tools/dataMigration/panEuropean.ts
```diff
@@ -47,7 +47,7 @@ export const migrate = async (props: {
   console.log('========== START ', new Date().getTime())
   const { assessmentName, legacySchemaName, assessmentLegacy, cycleNames, spec } = props
 
-  console.log('========== 1. delete old assessment')
+  // console.log('========== 1. delete old assessment')
```
minotogna: Can we leave console.log active in this file and use ignore comment ?

sorja (approved):

src/test/migrations/index.ts
sorja: Does logger work in this context? Can you check?

minotogna (approved):

PR #2366, opened 2023-03-07
2350 - Table 2b: update most recent years options

### LINKED ISSUE #2350: Table 2b: update most recent years options
ÖJ: Hi, in table 2b you can only select the yrs 2011 to 2025, shold be from 1990 to 2023. Thanks

rfontanarosa:
close #2350

Screenshot 2023-03-07 alle 14 29 48


sorja (approved):

sorja:
This is ok as is and should work as is

I would still prefer to cross match always for cell (variableName and columnName)

minotogna (approved):

PR #2376, opened 2023-03-10
2369 update translations ar

sorja:
import new translations from excel


src/i18n/resources/ar.js
```diff
@@ -117,21 +125,15 @@ module.exports.translation = {
...
+      VIEWER: 'مشاهد',
+      reviewer_plural: 'المراجعين',
+      nationalCorrespondent_plural: 'المراسلون الوطنيون',
+      alternateNationalCorrespondent_plural: 'المراسلون الوطنيون المناوبون',
+      collaborator_plural: 'المتعاونين',
     },
 
     resetPasswordEmail: {
-      subject: 'منصة تقييم الموارد الحرجية، إعادة ضبط كلمة المرور',
+      subject: 'منصة فرا - إعادة تعيين كلمة المرور',
```
samym392, line 136: "منصة تقييم الموارد الحرجية، إعادة ضبط كلمة المرور" this one make more sense

samym392 (changes requested):
when you copy and past the paragraphs has **full stop** at the end be carful because if you are not change the language of the keyboard the full stop will goes automatically to the **end right** as in **English** language not at the **end left** like in **Arabic** language.

minotogna (approved):

PR #2388, opened 2023-03-13
2299  - Table 1d additional row

### LINKED ISSUE #2299: Table 1d - additional row
Table 1d It would have been nice to have an additional row at the end of the table showing net change according to T1a and prefilled (validation check). That would let countries see how data they fill in correspond to what they have reported in T1a. As it is now, no validation against T1a is done and the table can be filled with data that completely contradicts T1a without any warning message

rfontanarosa:
close #2299

https://user-images.githubusercontent.com/10047945/225066048-2823b7e8-3e13-465f-8f7e-a1c2a4610635.mov

Screenshot 2023-03-14 alle 17 05 42


src/meta/assessment/row.ts
```diff
@@ -43,7 +43,7 @@ export interface RowProps {
   }
   index: number | string
   label?: RowLabel // TODO: remove? (check if used - probably not)
-  linkToSection?: string
+  linkToSection?: Record<CycleUuid, string> | string
```
minotogna: I think should be only Record<CycleUuid, string>

tools/dataMigration/migrateData/_repos.ts
```diff
@@ -17,6 +17,7 @@ export const getRows = (client: BaseProtocol, schema: string, table: Table): Pro
         props: {
           ...Objects.camelize(row.props),
           calculateFn: row.props.calculateFn,
+          linkToSection: row.props.linkToSection,
           validateFns: row.props.validateFns,
           chart: row.props.chart,
```
minotogna, line 22: maybe a RowAdapter would help reduce all these duplicated code?

sorja (dismissed):

tools/dataMigration/migrateData/_repos.ts
rfontanarosa: https://github.com/openforis/fra-platform/issues/2407

minotogna (approved):

sorja (approved):

Issue #2415, opened 2023-03-16
GEO: FE allow user to calculate on the fly area estimation for custom agreement recipe


eraviolo:
Endpoint to calculate on the fly forest agreement area:

POST to api/geo/estimations/forest-agreement
```
{
    "countryIso": "URY",
    "layers": [
        {
            "key": "ESAWorldCover"
        },
        {
            "key": "ESRI"
        },
        {
            "key": "Hansen",
            "options": {
                "gteTreeCoverPercent": 10
            }
        }
    ],
    "gteAgreementLevel": 2,
    "scale": 100
}
```

PR #2482, opened 2023-03-30
2415 allow user to calculate agreement area

yaguzmang:
Solves #2415 by:

- Refactoring:
    - Migrated MapContext to use global mapController instead (this allows to manipulate the map from outside of the Map Component)
    - Moved the mosaic properties to the mosaic options in redux (this is better to divide the logic).
    - Moved the loading spinner to its own file.
    - Deleted the duplicated enum 'LayerStatus'
    - Refactored the Geo statistics to use redux. The data is updated using the same hook (renamed to 'useGeoStatisticsHandler').
    - Refactored the statistics graph to update when its data changes.

New code:
   - Fetch a forest estimation based on the selected agreement layers, and then add the data to the statistics graph. 

Large GIF (802x904)


eraviolo (approved):
Functionality looks good to me, just some minor comments:
- The area estimation on the fly can take a while or timeout so we should add some status indicator or message indicating that the estimation for the area is being processed 
- Add unit (m) after the scale number or maybe label with 'Reducer scale' before select menu with 10m/30/100m as options

minotogna (approved):

PR #2484, opened 2023-03-31
2479 - Fixed One-To-One Message and Country Message Board

### LINKED ISSUE #2479: Message Board: add one-to-one or all members message error 'No data returned from the query'
when adding a one-to-one or to all members message, backend gives error (see screenshot)

Image

rfontanarosa:
close #2479

https://user-images.githubusercontent.com/10047945/229069796-30900554-c4ca-40f0-86e5-3da81d429b5c.mov

https://user-images.githubusercontent.com/10047945/229069928-cdcde439-c460-4eb8-9aea-6f3c2eb97e48.mov

### DIFF
--- a/src/server/repository/assessment/section/getOne.ts
+++ b/src/server/repository/assessment/section/getOne.ts
@@ -10,7 +10,8 @@ export const getOne = async (
   client: BaseProtocol = DB
 ): Promise<Section> => {
   const schemaName = Schemas.getName(props.assessment)
-  return client.one<Section>(
+
+  return client.oneOrNone<Section>(
     `
         select s.*
         from ${schemaName}.section s
@@ -18,6 +19,6 @@ export const getOne = async (
           and props -> 'cycles' ? $1;
     `,
     [props.cycle.uuid, 'sectionName' in props ? props.sectionName : props.id],
-    SectionAdapter
+    (section) => section && SectionAdapter(section)
   )
 }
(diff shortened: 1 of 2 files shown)


src/server/repository/assessment/section/getOne.ts
```diff
@@ -10,14 +10,15 @@ export const getOne = async (
...
+  return client.oneOrNone<Section>(
     `
         select s.*
         from ${schemaName}.section s
         where ${'sectionName' in props ? `s.props->>'name' = $2` : `s.id = $2`}
           and props -> 'cycles' ? $1;
     `,
     [props.cycle.uuid, 'sectionName' in props ? props.sectionName : props.id],
-    SectionAdapter
+    (section) => section && SectionAdapter(section)
```
sorja: does it call this if nothing is returned

minotogna: the method should be getOneOrNone and use client.oneOrNone in case that's what needed. but is it ?

minotogna: the return type of the method should be updated as well. but why do we want to allow this? 
shouldn't the caller know that it doesn't need to call this method ?

rfontanarosa: the error was caused by the SectionAdapter.

I had two options: add an exception in case the sectionName is 'messageBoard' or allow that the api can return null in case a section has not been found.

minotogna: Section is not mandatory to create a message . Then you can avoid to call this method in such case?

rfontanarosa: ok, fixed

sorja (approved):

minotogna (approved):

Issue #2455, opened 2023-03-24
GEO: Add protected areas section


eraviolo:
Layers keys:
- WDPA (this will require a new type of overlay on Google Map because it is a FeatureView asset from GEE)
- FilteredWDPA
- CustomPA

E.g. CustomPA: POST endpoint api/geo/layers/protected-area/
Body:

```
{
    "countryIso": "ITA",
    "layer": {
        "key": "CustomPA",
        "options": {
            "assetId": "users/path_to_gee_custom_image_asset"
        }        
    }
}
```

PR #2490, opened 2023-04-01
2455 add protected areas section

yaguzmang:
Solves #2455 by:

Refactoring:
-  Dividing the logic to handle map layers: separating the layers handlers in different files, so they are more modular and easier to maintain.
- Renaming "layers" to "forestLayers" since there will be new non-forest layers (this is also done in the redux reducers).

Adding new logic and UI:
- Added a new hook, redux state management, and UI components for the protected areas.

TODO: Find a way to render the WDPA layer. After spending quite some time trying to get the WDPA layers to work, it seems like it is not possible the way it is set up now.

As per the [documentation](https://developers.google.com/earth-engine/guides/featureview_overview): "The process for visualizing a FeatureView asset using the Google Maps API is a four part process: obtain a tiles key using ee.data.getFeatureViewTilesKey, pass it to an instance of FeatureViewTileSource, construct an ee.layer.ImageOverlay [...]" at the moment we call `ee.data.getFeatureViewTilesKey` in the backend and then we send the response to the frontend. In the mapController we then try to pass it to a new `FeatureViewTileSource`, but due to the typing that is lost when sending the object from backend to frontend, the `FeatureViewTileSource` constructor doesn't recognise the object and throws an error. Unfortunately, there is no casting available for this kind of object.  

Large GIF (862x484)


eraviolo (dismissed):

src/client/pages/Geo/GeoMap/GeoMapMenuData/MapVisualizerPanel/ProtectedAreaLayerSection/CustomAssetControl/CustomAssetControl.scss
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    &:disabled {
+      background-color: #f8f7f7;
+      border-color: #dedfdf;
+      box-shadow: 0 0 10px 2px #efeeee inset;
+      cursor: default;
+    }
+  }
+
```
minotogna: Any chance to use the global color variables ?

src/client/store/ui/geo/slice.ts
```diff
@@ -174,9 +182,57 @@ export const geoSlice = createSlice({
...
+      state.protectedAreasOptions.failedLayers = initialState.protectedAreasOptions.failedLayers
+    },
+    resetSingleProtectedAreaLayerStates: (state, { payload }: PayloadAction<ProtectedAreaKey>) => {
+      delete state.protectedAreasOptions.fetchedLayers[payload]
+      delete state.protectedAreasOptions.pendingLayers[payload]
+      delete state.protectedAreasOptions.failedLayers[payload]
+    },
+    setCustomProtectedAreaAssetId: (state, { payload }: PayloadAction<string>) => {
+      state.protectedAreasOptions.customAssetId = payload
+    },
```
minotogna, line 235: This file is becoming too big , that means there's either common logic to refactor or we should split this into several files ... what do you think ?

src/client/pages/Geo/GeoMap/GeoMapMenuData/MapVisualizerPanel/ProtectedAreaLayerSection/CustomAssetControl/CustomAssetControl.scss
yaguzmang: Yes, fixed this and a bug in the new commit.

src/client/store/ui/geo/slice.ts
yaguzmang: Yes, I agree it is getting too big. The logic for the layers is similar but not always the same, so generalizing that logic would end up in a lot of extra code handling edge cases. I think it should better be split. Is it okay to do it in the upcoming Burned Areas PR?

minotogna: let's think how to refactor this later on?

yaguzmang: Yes, splitting it into files is a good start, but there should be a better way to handle the layers. We can take a look into that later.

minotogna: well, the best way would be to avoid duplicated code.

minotogna (approved):

eraviolo (approved):

PR #2543, opened 2023-04-14
2510 - Plantation forest: enable empty value

### LINKED ISSUE #2510: High priority-Table 1b-NDP-…of which introduced(cannot be empty)
@minotogna 

Want to keep percentage of **…of which introduced** empty, but not be possible.

rfontanarosa:
close #2510

https://user-images.githubusercontent.com/10047945/232052158-4962cf27-49ac-4fbe-bbd7-be2f71648ae8.mov

### DIFF
--- a/src/meta/assessment/originalDataPoint/odps/updateNationalClass.ts
+++ b/src/meta/assessment/originalDataPoint/odps/updateNationalClass.ts
@@ -40,7 +40,7 @@ const calculateValues = (nationalClass: ODPNationalClass) => {
     otherPlantedForestPercent: getValueOrNull(otherPlantedForestPercent, rowIsMaxedForestCharacteristics),
     forestPercent: getValueOrNull(forestPercent, rowIsMaxedExtentOfForest),
     otherWoodedLandPercent: getValueOrNull(otherWoodedLandPercent, rowIsMaxedExtentOfForest),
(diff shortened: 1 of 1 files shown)


minotogna (approved):

sorja (approved):

src/meta/assessment/originalDataPoint/odps/updateNationalClass.ts
```diff
@@ -40,7 +40,7 @@ const calculateValues = (nationalClass: ODPNationalClass) => {
     otherPlantedForestPercent: getValueOrNull(otherPlantedForestPercent, rowIsMaxedForestCharacteristics),
     forestPercent: getValueOrNull(forestPercent, rowIsMaxedExtentOfForest),
     otherWoodedLandPercent: getValueOrNull(otherWoodedLandPercent, rowIsMaxedExtentOfForest),
-    forestPlantationIntroducedPercent: forestPlantationIntroducedPercent ?? '0',
+    forestPlantationIntroducedPercent,
```
sorja, line 43: should this be getValueOrNull

rfontanarosa: no, it shouldn't. I think that getValueOrNull is used to check if a in-row sum is 100% or not. the case of forestPlantation is different because we have only one value per row.

PR #2558, opened 2023-04-20
Validation: sdg indicator 3 and 4 not greatet than 101

### LINKED ISSUE #2532: SDG  15.2.1-Sub-Indicator 3 & Sub-Indicator 4
@minotogna 
Values which slightly higher than 100(below 101%) should **not be highlighted** since it is confirmed that 2015 forest area is used as baseline for the calculation of these two sub-indicators.

rfontanarosa:
close #2532

Screenshot 2023-04-20 alle 22 26 25


minotogna (commented):
@rfontanarosa are you really sure that the code in this pr addresses the issue ?

minotogna (approved):

PR #2562, opened 2023-04-24
Geo module refactoring - Add new types

yaguzmang:
This PR is the first part of the major refactoring of the way we handle map layers in the Geo Module. 

Before adding the new types, some renaming took place to improve consistency: 

- Renamed `ForestSource` to `ForestKey`, to be consistent with 'BurnedAreaKey' and 'ProtectedAreaKey'
- Renamed `sourcesMetadata` to `forestLayersMetadata`, since a source can be non-forest related.
- Renamed `burnedAreaSourcesMetadata` to `burnedAreaLayersMetadata` and `protectedAreaSourcesMetadata` to `protectedAreaLayersMetadata` to be more specific and consistent with `forestLayersMetadata`.
- Renamed 'Layer' to 'LayerConfig', since 'Layer' didn't actually fully describe a Layer.

Also, moved some types that were in the wrong place:

- Moved UI-related types from `forest.ts` to `ui.ts`, since the contents in `forest.ts` should be agnostic of UI typing as it is also used for the backend.
- Moved general Layer types from `forest.ts` to their own file. This is because `forest.ts` should only have types and contents specific to the forest layers. The Layer types and constants should have their own file since they are used for different kinds of layers. 

Finally, the new Layer types are added to `layer.ts`, without removing or deprecating other types. The reason for that is that it makes this PR a 'No Breaking Changes' PR, so it is easier to test and verify. The following PRs for the refactoring will include breaking changes, so this is a good checkpoint.


yaguzmang:
@mergifyio refresh

src/meta/geo/layer.ts
```diff
@@ -0,0 +1,63 @@
...
+export type LayerMetadata = {
+  scale?: number
+  palette?: Array<string>
+  citation?: string
+  forestAreaDataProperty?: string
+}
+
+export type LayerOptions = {
+  minTreeCoverPercentage?: Array<number>
+  agreementLayerOptions?: { agreementLevels: Array<number>; reducerScales: Array<number> }
```
minotogna: I would call this agreementLayer without options postfix. it's implicit by the type definitions

src/meta/geo/layer.ts
```diff
@@ -0,0 +1,63 @@
...
+    assetId?: string
+    year?: number
+    agreement?: {
+      layers: Array<LayerSource>
+      gteAgreementLevel: number
+    }
+  }
+}
+
+export interface Recipe {
```
minotogna, line 26: not sure what this interface does. however let's talk in the next meeting about this

minotogna (commented):
what about one Options interface ?

eraviolo (approved):

src/meta/geo/layer.ts
yaguzmang: Yes, good point. I updated it.

src/meta/geo/layer.ts
yaguzmang: Okay 👍

PR #2568, opened 2023-04-26
Refactor datasource

### LINKED ISSUE #2557: NationalData->DataSources: use only one “variables” property. fraVariables and variable —> become variables
```typescript
dataSources: {
    linkedVariables?: ... // keep linkedVariables as they are 
    // if text exists, render text. it can be readOnly if specified in prop
    text?: { readOnly?:boolean }
    // if table exists, render table
    table?: {
       // variables can be array of variable name, or single text input
        variables: Array<{variableName: VariableName; labelPrefix?:Label}> | 'text'
    }
}

```

```label: fra.2020.description.dataSources.variables```

- [ ] migrate fra metadata
- [ ] migrate panEuropean metadata

### PR DESCRIPTION
**Pull request by sorja**
Resolves #2557 

firefox_NGQgf4JhJS

- [x] metadata
  - [x] remove old code and references (eg. metadata)
- [x] frontend
  - [x] support new metadata
  - [x] refactor
  - [x] remove old code ansd clean up
- [x] migrations
- [x] i18n
- [x] final test (manual)

Known bug:
- [x] ~~When inserting a new row, the cursor jumps to the next one. Not sure what is causing this~~
firefox_SOa356WDah

### CODE REVIEW

File: src/meta/assessment/description/nationalDataDataSourceDescription.ts [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+  prefixLabel?: Label
+}
+
+export interface DataSourceDescription {
+  table?: {
+    variables: Array<DataSourceVariable>
+    typeOfDataSourceText?: boolean
+  }
+  text?: { readOnly?: boolean }
+}
```
sorja: this is what matters currently

src/client/pages/AssessmentSection/Descriptions/CommentableDescription/Description/DataSources/DataSourceColumn/ColumnComments.tsx
```diff
@@ -6,24 +6,25 @@ import { DataSource } from '@meta/assessment'
...
 import DataColumn from '@client/components/DataGrid/DataColumn'
 import VerticallyGrowingTextField from '@client/components/VerticallyGrowingTextField'
-import { datasourceValidators } from '@client/pages/AssessmentSection/Descriptions/Description/DataSources/DataSourceColumn/DataSourceColumn'
+
+import { datasourceValidators } from './datasourceValidators'
 
 type Props = {
   disabled: boolean
-  dataSource: DataSource
+  dataSourceValue: DataSource
```
sorja, line 14: Note: Added more descriptive names for dataSourceValue (= eg. assessment_fra_2025.descriptions entity) and dataSourceMetadata (= eg. assessment_fra.section.props.descriptions... entity)

src/client/pages/AssessmentSection/Descriptions/CommentableDescription/Description/DataSources/DataSourceColumn/ColumnTypeOfDataSource.tsx
```diff
@@ -0,0 +1,69 @@
...
+import DataColumn from '@client/components/DataGrid/DataColumn'
+import VerticallyGrowingTextField from '@client/components/VerticallyGrowingTextField'
+
+type Props = {
+  disabled: boolean
+  dataSourceValue: DataSource
+  onChange: (key: string, value: string) => void
+}
+
+const TextInput: React.FC<Props> = (props: Props) => {
```
sorja, line 19: Subcomponents kept in same file, only used here and only used to tidy and separate logic from parent

src/test/migrations/steps/20230427094458-step-datasource-include-variable-label.ts
```diff
@@ -0,0 +1,207 @@
...
+import { MetadataController } from '@server/controller/metadata'
+import { BaseProtocol, Schemas } from '@server/db'
+
+interface DataSourceVariablesDeprecated {
+  // Custom i18n keys to include
+  include: Array<string>
+  prefixes?: Record<VariableName, Label>
+}
+
+type DataSourceColumnDeprecated =
```
sorja, line 26: Old types

src/meta/assessment/description/nationalDataDataSourceDescription.ts
```diff
@@ -0,0 +1,17 @@
+import { Label } from '@meta/assessment'
+
+export interface DataSourceVariable {
+  value: string
```
minotogna: better to call this variableName

src/meta/assessment/description/nationalDataDataSourceDescription.ts
```diff
@@ -0,0 +1,17 @@
...
+  label: Label
+  prefixLabel?: Label
+}
+
+export interface DataSourceDescriptionTable {
+  variables: Array<DataSourceVariable>
+  typeOfDataSourceText?: boolean
```
minotogna: here it should be intuitive to understand that the type of input can depend on a predefined array of variables, or text input if no variables are provided. 

To make it easier, i would suggest to make variables an optional property.
The component will behave this way: if array of variables is provided, then values change based on the select; 
otherwise, a text input.

What do you think?

sorja: This makes sense

What about typeOfDataSourceText? I was struggling to find a better name, so left it open for discussion

This boolean, when set to true, renders a text input field, otherwise a dropdown select of the following:

( https://github.com/openforis/fra-platform/blob/7a2d1f9744cd4060d13fe6a2d24159af6dbc9ad4/src/meta/assessment/description/dataSourceType.ts )

minotogna: I don't think we need that typeOfDataSourceText  property at all.

we could render dropdown if array of variables is provided, otherwise render text input.

what do you think?

sorja: This is for types, not variables (second column, refer to video)

minotogna: 😅I missed that sorry 😅
now I see it. let's leave it as you wrote it, if in any future refactor of that property name we might possibly want to do, it will be easy.
what do you think?

sorja: sure, sounds good

ill handle the first comment

rfontanarosa (dismissed):

minotogna (approved):

rfontanarosa (approved):

PR #2590, opened 2023-05-05
Geo refactor/add redux management to layers control

yaguzmang:
Adding redux state management to toggle layers, change opacity, and set the custom assets id.

Large GIF (1074x478)

### DIFF
--- a/src/client/store/ui/geo/slice.ts
+++ b/src/client/store/ui/geo/slice.ts
@@ -141,12 +141,12 @@ export const geoSlice = createSlice({
       state.sections[sectionKey][layerKey] = { ...layerState, opacity }
     },
     setAssetId: (
-      state,
-      {
-        payload: { sectionKey, layerKey, assetId },
-      }: PayloadAction<{ sectionKey: LayerSectionKey; layerKey: LayerKey; assetId: string }>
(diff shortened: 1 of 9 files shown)


src/client/store/ui/geo/slice.ts
```diff
@@ -33,6 +33,20 @@ const initialState: GeoState = {
...
 
+const getSectionState = (state: Draft<GeoState>, sectionKey: LayerSectionKey): LayersSectionState => {
+  // Default the states to an empty object if they don't exist yet
+  state.sections[sectionKey] ??= {} as LayersSectionState
+  return state.sections[sectionKey]
+}
+
+const getLayerState = (state: Draft<GeoState>, sectionKey: LayerSectionKey, layerKey: LayerKey): LayerState => {
+  // Default the states to an empty object if they don't exist yet
+  state.sections[sectionKey] ??= {} as LayersSectionState
```
minotogna: You can call getSectionState here

yaguzmang: I kept it like this for readability, but now that you mention, if we are ever to change the way we initialize sections, it's better to call the getSectionState. Thanks, this is now updated.

minotogna (approved):

PR #2633, opened 2023-05-15
Geo refactor/handle agreement layer

yaguzmang:
Added stage management and UI components for handling the agreement layer. Additionally, added a custom hook to re-fetch the agreement layer when the level changes. The styles and form of the UI components are taken from the previous implementation. 

Large GIF (846x762)


src/client/store/ui/geo/actions/postLayer.ts
```diff
@@ -1,27 +1,27 @@
...
   layerState: LayerState
+  sectionState?: LayersSectionState
 }
 
-export const postLayer = createAsyncThunk<[LayerSectionKey, LayerKey, string], PostLayerProps>(
+export const postLayer = createAsyncThunk<[LayerSectionKey, LayerKey, LayerResponseData], PostLayerProps>(
   'geo/post/layer',
-  async ({ countryIso, sectionKey, layerKey, layerState }) => {
-    const body = _getLayerRequestBody(countryIso, layerKey, layerState)
+  async ({ countryIso, sectionKey, layerKey, layerState, sectionState }) => {
```
minotogna: Better not to pass the state . Here you have access to getState directly .

src/client/store/ui/geo/hooks/useAgreementLevel.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { LayerKey, LayerSectionKey } from '@meta/geo'
+
+import { useAppDispatch } from '@client/store'
+import { useCountryIso, usePrevious } from '@client/hooks'
+
+import { GeoActions } from '../slice'
+import { useGeoLayerSection } from '.'
+
+export const useAgreementLevel = (sectionKey: LayerSectionKey, layerKey: LayerKey) => {
```
minotogna: This hooks doesn't belong to store. It dispatches an action . If it's used in one please , better to put in a hook folder where used. If common hook , under common hook folder

src/client/store/ui/geo/hooks/useAgreementLevel.ts
```diff
@@ -0,0 +1,26 @@
...
+  const sectionState = useGeoLayerSection(sectionKey)
+  const layerState = sectionState?.[layerKey]
+  const agreementLevel = layerState?.options?.agreementLayer?.level
+  const prevAgreementLevel = usePrevious(agreementLevel)
+
+  useEffect(() => {
+    if (prevAgreementLevel === agreementLevel) return // Skip if no change
+    if (agreementLevel === undefined) return // Skip when the property is not set
+    if (prevAgreementLevel === null) return // Skip with initial prev. value
+
```
minotogna: Usually you can avoid getting the prev state , since a hook can run when a value changes . Check the logic .

src/client/store/ui/geo/hooks/useAgreementLevel.ts
yaguzmang: Okay, in that case, I'll move this one and useGteTreeCoverPercent (I based it off `useGetTableSections`).

src/client/store/ui/geo/actions/postLayer.ts
yaguzmang: Good one 👍 Thanks!

src/client/store/ui/geo/hooks/useAgreementLevel.ts
yaguzmang: Done, thanks!

src/client/store/ui/geo/hooks/useAgreementLevel.ts
yaguzmang: Yes, after getting the state from the postLayer action, the usePrevious became unnecessary. Thanks!

src/client/pages/Geo/GeoMap/GeoMapMenuData/MapVisualizerPanel/LayersSectionPanel/components/AgreementLevelControl/AgreementLevelControl.tsx
```diff
@@ -0,0 +1,128 @@
...
+                  currentAgreementPaletteLength > 0
+                    ? {
+                        backgroundColor: `${layerState?.options?.agreementLayer?.palette?.[agreementLevelOffset]}`,
+                        boxShadow: `0px 0px 3px 3px ${layerState?.options?.agreementLayer?.palette?.[agreementLevelOffset]}30`,
+                      }
+                    : {}
+
+                const style = level === currentSelectedLevel ? selectedStyle : genericStyle
+
+                return (
```
minotogna: Exporting this as separate component would make the code easier to maintain

src/client/pages/Geo/GeoMap/hooks/useFetchGteTreeCoverPercentLayer.ts
```diff
@@ -0,0 +1,19 @@
...
+export const useFetchGteTreeCoverPercentLayer = (sectionKey: LayerSectionKey, layerKey: LayerKey) => {
+  const dispatch = useAppDispatch()
+  const countryIso = useCountryIso()
+  const layerState = useGeoLayer(sectionKey, layerKey)
+  const gteTreeCoverPercent = layerState?.options?.gteTreeCoverPercent
+
+  useEffect(() => {
+    if (gteTreeCoverPercent === undefined) return // Skip when the property is not set
+    dispatch(GeoActions.postLayer({ countryIso, sectionKey, layerKey }))
+  }, [countryIso, layerKey, gteTreeCoverPercent, sectionKey, dispatch])
```
minotogna, line 18: This is the same than useFetch agreement level ... can't it be one and only one fetch ? Then based on the layer , you parse the params ?

yaguzmang: Wouldn't that make it harder to implement logic specific to those layers in the future? like caching (which would need to be handled differently for agreement).

src/client/pages/Geo/GeoMap/GeoMapMenuData/MapVisualizerPanel/LayersSectionPanel/components/AgreementLevelControl/AgreementLevelControl.tsx
yaguzmang: True. I added an `AgreementLevelSelector` component to the UI commit 👍

src/client/pages/Geo/GeoMap/hooks/useFetchGteTreeCoverPercentLayer.ts
yaguzmang: Also, we need to add logic to re-fetch the agreement layer when the selected layers change. My initial thought was to do that from `toggleLayer` in the slice, but that is not possible since that would be a side effect. I suppose then it would need to go in this hook, making it different from the `useFetchGte...`. Any other suggestions of where that logic could go?

minotogna (approved):

PR #2711, opened 2023-06-01
Jsep assessment cycle member

sorja:
notes
- *parseExpression* must match depth and not type! we only have this formula support and they cannot overlap:
  * extentOfForest.forestArea (depth 2)
  * extentOfForest.forestArea['2025'] (depth 3)
  * fra['2025'].extentOfForest.forestArea (depth 4)
  * fra['2025'].extentOfForest.forestArea['2005'] (depth 5)

 **Depth matching required to avoid issues if we use non numeral col names (eg. growing stock)**

- [x] meta cache seems to be ok
- [x] update calculation and validation function parsing (using parseExpression)
- [x] parse expression needs to return col
- [x] depth matching vs type matching
- [x] Compare vs prod meta cache


package.json
```diff
@@ -24,7 +24,7 @@
     "run:prod": "run-s build start",
     "run:prod:server": "node dist/server/start",
     "start": "run-s migrations:run run:prod:server",
-    "generate-metacache": "yarn jest --verbose false --no-cache --detectOpenHandles --setupFiles dotenv/config --config src/test/generateMetaCache/jest.config.js",
+    "generate-metacache": "ts-node src/tools/generateMetaCache/",
```
minotogna: let's use the new convention? metacache:generate ?

src/server/controller/assessment/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
```diff
@@ -12,35 +13,47 @@ const excludeDependant = (row: Row, tableName: string, variableName: string): bo
...
 
-    // @ts-ignore
-    const objectName = object?.object?.name ?? object.name
-    // @ts-ignore
-    const propertyName = object?.property?.name ?? property.name
+    // Example of: <calcualteFn> = [...result]
+    // Case 1: extentOfForest.forestArea (2) [extentOfForest, forestArea]
+    // Case 2: extentOfForest.forestArea['2025'] (3) [extentOfForest, forestArea]
+    // Case 3: fra['2025'].extentOfForest.forestArea (4) [fra, '2025', extentOfForest, forestArea]
+    // Case 4: fra['2025'].extentOfForest.forestArea['2025'] (4) [fra, '2025', extentOfForest, forestArea]
```
minotogna: can they be removed from here ?

src/server/controller/cycleData/getOriginalDataPointVariables.ts
```diff
@@ -61,7 +61,7 @@ export const getOriginalDataPointVariables = (cycle: Cycle) => {
 // Find given variable from ODPVariables
 export const isODPVariable = (cycle: Cycle, variable: VariableCache): boolean => {
   return getOriginalDataPointVariables(cycle).some((odpVariable) => {
-    const keys: Array<keyof VariableCache> = ['tableName', 'variableName']
+    const keys: ['tableName', 'variableName'] = ['tableName', 'variableName']
```
minotogna: why defining a type exactly like the value ?

src/server/controller/assessment/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
```diff
@@ -12,35 +13,47 @@ const excludeDependant = (row: Row, tableName: string, variableName: string): bo
...
+      tableName: _tableName,
+      variableName: _variableName,
+      assessmentName,
+      cycleName,
+    } = ExpressionEvaluator.parseExpression(expressionNode)
 
-    if (assessmentMetaCache.variablesByTable[objectName]) {
-      const dependantTable = assessmentMetaCache[type].dependants?.[objectName] ?? {}
-      const dependants = dependantTable[propertyName] ?? []
+    if (assessmentMetaCache.variablesByTable[_tableName]) {
```
minotogna: 1. it's a but unclear reading _tableName and tableName. it's difficult to understand what is what.
2. parseExpression sounds like a public method that does something different from what it really does. call this parseMemberVariable and don't expose it from ExpressionEvaluator. is it just an internal method used only here, right?

the result of my 2 comments above could be:

```
const memberVariable  = parseMemberVariable(expressionNode)
```

sorja: Used in 2 places:
https://github.com/openforis/fra-platform/blob/89786310e0db4026c01525d239ab7f267a2c52bb/src/meta/expressionEvaluator/evaluators/member.ts#L8

https://github.com/openforis/fra-platform/blob/89786310e0db4026c01525d239ab7f267a2c52bb/src/meta/expressionEvaluator/evaluators/member.ts#L14

https://github.com/openforis/fra-platform/blob/89786310e0db4026c01525d239ab7f267a2c52bb/src/server/controller/assessment/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts#L20-L32

minotogna: I only see one usage, and one import

sorja: image
image

sorja: should i still export only the function?

minotogna (approved):

rfontanarosa (approved):

Issue #2604, opened 2023-05-09
Persist and Visualize Estimation method applied 


pengchenglai:
#### Part 1 (Persist Estimation method parameters: Backend)  ( @minotogna )
- [x] when estimating values, persist in a new table node_values_estimation (id, uuid, props:{type,parmas...}) (to check table format) estimation applied. + add an entry to the activity log of type 'nodeEstimationApplied' with target the created nodeValuesEstimation .
- [x] when marking node as estimated, add to node value prop estimationUuid.

#### Part 2 (Visualize Estimation method  applied: Frontend) ( @rfontanarosa )
- [x] in ui, to cell which node values have an estimationUuid, add on top right a squared box with letter E inside.
- [x] on data table effect, fetch estimation if a node is encountered with an estimationUuid
- [x] when hovering E box, show tooltip with estimation information applied (add loader within tooltip if server still fetching) 

#### Part 3 (Show estimated cell invalid: Backend/Frontend)
Objective: An estimated cell must be 'invalidated' if an ODP has been updated after estimation was applied
Note: last edit odp timestamp must concern only and only values (not other metadata information)
- [x] UI/BE: on Datatable effect, and table has odp, add endpoint and fetch lastOdpEditTime, or add last edit time to odp already fetched?
- [x] UI: when a cell is estimated and estimation engine has been applied before last odp edit timestamp, on the left of box E, add warning icon (orange) with tooltip: `This value might not longer be consistent with the estimation method applied. One or original data point has changed after estimation.`

_________________________________________________________________________________________________________________________ 

**Original Issue:**

@minotogna 
New values estimated by estimation methods such as linear, repeat last or annual change should be highlighted.

Without estimation(how it looks like):
<img width="738" alt="1" src="https://github.com/openforis/fra-platform/assets/99216009/b398d80f-d83c-4821-90ba-37b562d48c10">

 The cell values using estimation method should be highlighted:
<img width="798" alt="2" src="https://github.com/openforis/fra-platform/assets/99216009/00ceb197-3d58-4751-91ac-b0f6e1db714d">

pengchenglai:
One possible idea could be to put the capital letter of the method in the corner of the cell as an example of using _Repeat last_ below:
<img width="798" alt="3" src="https://github.com/openforis/fra-platform/assets/99216009/6d818b6e-66b4-4942-ac76-f351937c1b1e">

Or any other way if feasible and suitable.

minotogna:
[...]

PR #2748, opened 2023-06-07
2604 - Visualize Estimation method applied: Frontend (BIS)

rfontanarosa:
part of #2604

https://github.com/openforis/fra-platform/assets/10047945/469f6bad-fa84-4840-8540-88a490aa728b


sorja (dismissed):

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/hooks/useOnChange.ts
```diff
@@ -38,12 +38,15 @@ export default (props: Props): UseOnChange => {
   const assessment = useAssessment()
   const assessmentSection = useAssessmentSection(sectionName)
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { estimated, estimationUuid, ...nodeValue } = _nodeValue ?? ({} as NodeValue)
```
minotogna, line 42: why this estlint disable? I think it's unnecessary. what about using the ?. to access nullable properties ?

src/i18n/resources/en/common.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   clearTable: 'Clear table',
   column: 'Column',
   continue: 'Continue',
+  method: 'Method',
```
minotogna: please, ask for translations . or we just remove it from the tooltip ? we only show Linear/etc... ?

src/server/controller/cycleData/getNodeValuesEstimations.ts
```diff
@@ -0,0 +1,27 @@
...
+    assessment: Assessment
+    countryIso: CountryIso
+    cycle: Cycle
+    tableName: string
+  },
+  client: BaseProtocol = DB
+): Promise<Record<string, NodeValuesEstimation>> => {
+  const { assessment, cycle, countryIso, tableName } = props
+
+  const table = await TableRepository.getOne({ assessment, cycle, tableName })
```
minotogna, line 19: getOne should return the table. table?.uuid should be table.uuid. can u please verity tableRepository.getOne ?

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/hooks/useOnChange.ts
rfontanarosa: because we want to exclude estimated and estimationUuid, keeping the others, I don't know if there is a prettier way to do that. the other options were: use delete (I don't want to change the original object), iterate over the keys.

src/i18n/resources/en/common.js
rfontanarosa: I removed the label Method

src/server/controller/cycleData/getNodeValuesEstimations.ts
rfontanarosa: you're right. at some point I modified the tableRepository.getOne because it can't return null (it returns an error if table doesn't exist). the null check is useless there and I removed it.

src/client/store/data/actions/postEstimate.ts
```diff
@@ -3,12 +3,14 @@ import axios from 'axios'
...
 
-export const postEstimate = createAsyncThunk<RecordAssessmentData, CycleDataParams & EstimateBody>(
-  'section/post/estimate',
-  async ({ fields, method, tableName, ...params }) => {
-    const { data } = await axios.post(ApiEndPoint.CycleData.Table.estimate(), { fields, method, tableName }, { params })
-    return data
-  }
-)
+export const postEstimate = createAsyncThunk<
+  { nodes: RecordAssessmentData; nodeValueEstimations: Record<string, NodeValuesEstimation> },
```
minotogna, line 10: Does the ui pass this new NodeValuesEstimation as param now?

minotogna (approved):

sorja (approved):

PR #2760, opened 2023-06-12
DataSources error messages

### LINKED ISSUE #2729: Add notifications of errors for the National Data(Data sources + type of data source eg NFI, etc)
@minotogna 

There are no error notifications currently available in the _Data sources + type of data source eg NFI, etc_ of the National Data.

sorja:
image

### DIFF
--- a/src/client/components/DataGrid/DataColumn.tsx
+++ b/src/client/components/DataGrid/DataColumn.tsx
@@ -6,8 +6,13 @@ import classNames from 'classnames'
 type Props = React.PropsWithChildren<{ className?: string; head?: boolean }>
 
 const DataColumn: React.FC<Props> = (props: Props) => {
-  const { children, className, head } = props
-  return <div className={classNames('data-grid-column', className, { head })}>{children}</div>
+  const { children, className, head, ...rest } = props
+  return (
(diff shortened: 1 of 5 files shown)


src/client/components/DataGrid/DataColumn.tsx
```diff
@@ -6,8 +6,13 @@ import classNames from 'classnames'
 type Props = React.PropsWithChildren<{ className?: string; head?: boolean }>
 
 const DataColumn: React.FC<Props> = (props: Props) => {
-  const { children, className, head } = props
-  return <div className={classNames('data-grid-column', className, { head })}>{children}</div>
+  const { children, className, head, ...rest } = props
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
```
minotogna, line 11: I think it's better to have explicit props , plus better to avoid eslint ignore as much as possible

minotogna (approved):

src/client/components/DataGrid/DataColumn.tsx
sorja: Yes, I agree.

In this case though, this is used as a 'dumb' component therefore added the object spreading to support eg. the tooltip on parent level

rfontanarosa (approved):

PR #2766, opened 2023-06-13
Updated CSV files

### LINKED ISSUE #2765: [PanEuropean] Update CSVs for Country Comments
Update de CSV for the issues 6.X

persas:
Closes #2765


minotogna (commented):
The csv files include country names , not anymore country iso . Does the import work correctly?

persas:
> The csv files include country names , not anymore country iso . Does the import work correctly?

Oh God... I cannot test the imports anymore because something changed. Just slacked that a while ago. But they won't work for sure, I'm asking through email now...

PR #2769, opened 2023-06-14
PanEuropean: use only English

### LINKED ISSUE #2679: panEuropean: use only English

sorja:
# Select any language in FRA
image

# Change to panEuropean with same user/without user
image

Language is changed but not persisted.
Changing back to FRA assessment shows previously selected language

### DIFF
--- a/src/client/components/PageLayout/Header/LanguageSelector/LanguageSelector.tsx
+++ b/src/client/components/PageLayout/Header/LanguageSelector/LanguageSelector.tsx
@@ -5,20 +5,14 @@ import { Lang, LanguageCodes } from 'meta/lang'
 
 import Icon from 'client/components/Icon'
 import PopoverControl from 'client/components/PopoverControl'
+import { updateLanguage } from 'client/utils/updateLanguage'
 
 const LanguageSelector: React.FC = () => {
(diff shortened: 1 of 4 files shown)


src/client/components/PageLayout/Header/LanguageSelector/LanguageSelector.tsx
```diff
@@ -1,26 +1,54 @@
...
+  )
 
   const languageSelectionItems = LanguageCodes.map((lang: Lang) => ({
     content: i18n.t(`language.${lang}`),
     onClick: () => updateLanguage(lang),
   }))
 
+  // If the assessment is panEuropean, we use the english language,
+  // and we don't show the language selector
+  useEffect(() => {
```
minotogna: Can this be done at initialization level ? The language menu should only be hidden and not contain any logic . What do you think ?

sorja: Introduced utils/updateLanguage
Moved logic to assessment initialisation (highest possible level)

minotogna (approved):

rfontanarosa (approved):

PR #2831, opened 2023-06-30
2723 - Used StatusTransition to differentiate notification based on direction

### LINKED ISSUE #2723: Highest priority User notification list(Report sent back to the country)
@minotogna @rfontanarosa @sorja 
Ideally, the reviewer should be able to see a list of **all users** when sending back a national report, not just the reviewers and NC.

By default, the list should include all users, including **alternate NC** and **collaborators**. The option to select people for notification should also be available.

For example, in the Vanuatu notification list, the alternate NC should be included when Monica was trying to send back the report. Similarly, in the case of China, the collaborator should be notified as they are the main person working on the report.

rfontanarosa:
close #2723

https://github.com/openforis/fra-platform/assets/10047945/44d9ba2f-8769-48c8-b86e-c0fb6fad4cb8

### DIFF
--- a/src/meta/user/userRoles.ts
+++ b/src/meta/user/userRoles.ts
@@ -5,6 +5,8 @@ import { AssessmentStatus } from 'meta/area/country'
 import { Assessment } from 'meta/assessment'
 import { User } from 'meta/user/user'
 
+import { StatusTransition } from 'client/components/PageLayout/Toolbar/Status/types'
+
 import { RoleName, UserRole } from './userRole'
 
 export const isInvitationExpired = (userRole: UserRole<RoleName>, expiryPeriod?: number) =>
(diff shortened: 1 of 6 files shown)


src/meta/user/userRoles.ts
```diff
@@ -13,11 +15,18 @@ export const isInvitationExpired = (userRole: UserRole<RoleName>, expiryPeriod?:
...
 
 // Return roles to receive email on country assessment status change
-const getRecipientRoles = (props: { status: AssessmentStatus }) => {
+const getRecipientRoles = (props: { status: StatusTransition }) => {
   const { status } = props
 
-  switch (status) {
+  switch (status.status) {
     case AssessmentStatus.editing:
+      if (status.direction === 'previous')
```
minotogna: What's this previous ? The roles are purely based on the target status

minotogna (changes requested):

minotogna (approved):

minotogna (approved):

PR #2833, opened 2023-07-03
enable free text entry in Year for data source

### LINKED ISSUE #2813: Year for data source - Data sources
@minotogna 

We will keep the current options for year selection regarding the **Year for data source** in the Data sources. Meanwhile, we also need to **enable free text entry**. 

Reason for this:
Discussed it with colleagues: it could include several reference years, such as 1999, 2003, and 2017, for a single reference. Alternatively, it could also refer to a period of time, like table 1d.

minotogna:


sorja:
Would Multiselect be preferable?
Approved (checked referenced issue)

sorja (approved):

PR #2855, opened 2023-07-12
Feat/paneu updates

persas:
All the fixes we saw in the meeting


src/test/dataMigration/stepsPanEuropean/migrateForestServiceCategoryValues.ts
```diff
@@ -15,23 +15,22 @@ export const migrateForestServiceCategoryValues = async (props: Props, client: B
...
-                            left join ${schemaAssessment}.col c on n.col_uuid = c.uuid
-                            left join ${schemaAssessment}.row r on r.id = c.row_id
-                            left join ${schemaAssessment}."table" t on r.table_id = t.id
-                   where c.props ->> 'colName' = 'forest_service_category'
-                     and n.value ->> 'raw' is not null),
-             b as (select a.id, a.value, a.rep[1], replace(initcap(replace(a.rep[1], '_', ' ')), ' ', '') as val
-                   from a)
-        update ${schemaCycle}.node n
-        set value = jsonb_set(n.value, '{raw}', to_jsonb(val.key))
+      with a as (select distinct n.id,
```
minotogna: why has this changed ? difficult to see diffs with different formatting

src/server/service/socket/socketService.ts
```diff
@@ -20,7 +20,7 @@ const init = (server: http.Server): void => {
   //   console.log('==== NEW CONNECTION', socket)
   // })
 
-  io.engine.on('connection_error', (err) => {
+  io.engine.on('connection_error', (err: any) => {
```
minotogna: is this changes needed ? I'd rather not use any. but check if there's a type in case you want to add it.

src/test/dataMigration/stepsPanEuropean/migrateForestServiceCategoryValues.ts
persas: It was failing, sorry for the formatting, was automatic. Basically, the only change was adding `as` here:

image

src/server/service/socket/socketService.ts
persas: I rather not use any either. But type Error didn't have `.code` and it's used below, and it was failing the migration if nothing changes.

persas: I rather not use any either. But type Error didn't have .code and it's used below, and it was failing the migration if nothing changes.

src/test/dataMigration/stepsPanEuropean/migrateForestServiceCategoryValues.ts
minotogna: what version of Postgres and node are you using ?

minotogna (approved):

PR #2868, opened 2023-07-18
Feat/pan eu fixes

persas:
All fixes commented in today's meeting


minotogna (changes requested):
please, use label keys in validators

minotogna (approved):

PR #2898, opened 2023-08-02
2897 geo add statistics sections for protected area burned area layers

### LINKED ISSUE #2897: GEO: Add statistics sections for Protected Area /Burned Area Layers
verify with and without tree cover

sorja:
https://github.com/openforis/fra-platform/assets/5508251/d73742da-d07d-43d6-99b2-776b01b49d6b


src/client/components/Navigation/Navigation.tsx
```diff
@@ -15,9 +15,12 @@ const Navigation: React.FC = () => {
   const countryIso = useCountryIso()
   const dispatch = useAppDispatch()
   const { print } = useIsPrint()
+  const isInGeoPage = useIsGeoPage()
+
   useEffect(() => {
-    if (print) dispatch(NavigationActions.updateNavigationVisible(false))
-  }, [print, dispatch])
+    if (print || isInGeoPage) dispatch(NavigationActions.updateNavigationVisible(false))
```
minotogna: better not to render at all, to avoid the animation. is that ok?

src/client/pages/Geo/GeoMap/GeoMapMenuStatistics/ProtectedAreaPanel/ProtectedAreaPanel.scss
```diff
@@ -0,0 +1,12 @@
...
+.geo-map-menu-statistics-btn-download {
+  float: right;
+  width: 150px;
+  margin: $spacing-s $spacing-l $spacing-s;
+}
+
+.table-title {
+  padding: 0 0 $spacing-s $spacing-m;
+  margin: auto;
+}
```
minotogna: this is duplicated from BarnedAreaPanel. any chance to avoid that ?

src/client/components/Navigation/Navigation.tsx
sorja: If this is not rendered, for some reason the map doesn’t render

This doesn’t cause an animation though, so I would leave as is if this is ok, otherwise I can also investigate why the map doesn’t render when this is hidden


Which is preferred?

minotogna: that's because a css grid. please have a look there , it should be easy.

sorja: Great, the issue was only that there was no min-width.

sorja: Screenshot 2023-08-02 at 15 35 36

src/client/components/Navigation/Navigation.tsx
```diff
@@ -15,10 +15,13 @@ const Navigation: React.FC = () => {
   const countryIso = useCountryIso()
   const dispatch = useAppDispatch()
   const { print } = useIsPrint()
+  const isInGeoPage = useIsGeoPage()
+
   useEffect(() => {
     if (print) dispatch(NavigationActions.updateNavigationVisible(false))
-  }, [print, dispatch])
-  if (!countryIso || print) return <div />
+  }, [print, dispatch, isInGeoPage])
```
minotogna: remove isInGeo from dependency hook

src/client/pages/Geo/geo.scss
```diff
@@ -4,4 +4,6 @@
   display: flex;
   flex: 1;
   width: 100%;
+  // min-height: 100vh - (headerHeight + toolbarHeight);
+  min-height: calc(100vh - (64px + 50px));
```
minotogna: why not the vars ?

src/client/store/ui/navigation/hooks/index.ts
```diff
@@ -1,11 +1,13 @@
...
+import { useCountryIso, useIsGeoPage } from 'client/hooks'
 
 export const useNavigationVisible = (): boolean => {
   const countryIso = useCountryIso()
+  const isInGeoPage = useIsGeoPage()
+
   const navigationVisible = useAppSelector((state) => state.ui.navigation.visible)
 
-  if (!countryIso) {
+  if (!countryIso || isInGeoPage) {
```
minotogna, line 10: why should it matter here ?

sorja: empty div otherwise rendered (eg. `!countryIso` handles this for landing page)

minotogna (approved):

PR #2909, opened 2023-08-04
Update growingStockTotal validations

### LINKED ISSUE #2836: Missing validation check - table 2a Total growing stock
@minotogna 

No validation check against **Total Forest** when one of these two classes is missing **Naturally regenerating forest** and **Planted forest**.

### LINKED ISSUE #2851: Table 2a total growing stock
@minotogna 

**Planted forest** should  = **…of which plantation forest** + **…of which other planted forest** for **the total growing stock**. This rule will be applied to calculate the total growing stock when values for _all three classes exist_, which also should be flexibly matched, allowing for a **±1 unit difference**, considering that some NCs would use the first table, Growing stock, and there is a conversion process involved.

sorja:
Resolves #2851 
Resolves #2836 

firefox_EZUM3S8dQm

image

image


src/meta/expressionEvaluator/functions/index.ts
```diff
@@ -69,6 +70,7 @@ export const functions: Array<ExpressionFunction<Context>> = [
   validatorPrivateOwnership,
   validatorRemainingLandWithTreeCoverTotal,
   validatorSubCategory,
+  validatorSubCategoryEqualToParent,
```
minotogna: can we use similar approach used in validatorSumSubCategoriesNotGreaterThanParent (showing values, labels,) ? and call this validatorSumSubCategoriesNotEqualToParent ?
Please, extract common code if needed.

minotogna: also, can u close both issues related to tables 2a in one pr ?

sorja: Sure! Sounds good

sorja: > also, can u close both issues related to tables 2a in one pr ?

It is done so, I will add references

minotogna (approved):
👍

PR #2914, opened 2023-08-07
fix crash when lang null

### LINKED ISSUE #2912: newly invited users have in the database the lang of the user that has invited them
**Describe the bug**
Crash

**To Reproduce**
1. Log in with newly invited

**Expected behavior**
No crash

sorja:
Resolves #2912 

- Cleaned up code and removed onMount

### DIFF
--- a/src/client/pages/PageRoutes/hooks/useInitLanguage.tsx
+++ b/src/client/pages/PageRoutes/hooks/useInitLanguage.tsx
@@ -2,33 +2,29 @@ import { useEffect } from 'react'
 
 import { AssessmentNames } from 'meta/assessment'
 import { Lang } from 'meta/lang'
-import { User } from 'meta/user'
 
 import { useAssessment } from 'client/store/assessment'
 import { useUser } from 'client/store/user'
 import { useLanguage, useUpdateLanguage } from 'client/hooks/useLanguage'
-import { useOnMount } from 'client/hooks/useOnMount'
 import { Urls } from 'client/utils'
-
-const getLang = (user: User) => {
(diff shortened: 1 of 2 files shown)


src/client/pages/PageRoutes/hooks/useInitLanguage.tsx
```diff
@@ -2,28 +2,33 @@ import { useEffect } from 'react'
...
 export const useInitLanguage = (): void => {
+  const user = useUser()
   const language = useLanguage()
   const updateLanguage = useUpdateLanguage()
   const assessment = useAssessment()
+  const lang = getLang(user)
 
   useOnMount(() => {
-    const lang = (Urls.getRequestParam('lang') || localStorage.getItem('i18n/lang') || Lang.en) as Lang
     if (lang !== language) updateLanguage({ lang }).then(() => ({}))
```
minotogna: I'm thinking, do we still need this onMount? I suspect not, can you please verify ?

src/client/pages/PageRoutes/hooks/useInitLanguage.tsx
```diff
@@ -2,28 +2,33 @@ import { useEffect } from 'react'
...
+const getLang = (user: User) => {
+  return (Urls.getRequestParam('lang') || localStorage.getItem('i18n/lang') || user?.props?.lang || Lang.en) as Lang
+}
+
 export const useInitLanguage = (): void => {
+  const user = useUser()
   const language = useLanguage()
   const updateLanguage = useUpdateLanguage()
   const assessment = useAssessment()
+  const lang = getLang(user)
```
minotogna: also, this getLang might be useless, and better to derive it in the useEffect, so that user is a dependency of the useEffect

minotogna (approved):

PR #2933, opened 2023-08-11
Sub category validator: Sub categories sum must be equal to parent

### LINKED ISSUE #2923: Validation check(tab 2a)
@minotogna 
MicrosoftTeams-image (3)

**Expected behavior**
The validation check(above) should not be applied when one of **…of which plantation forest** and **…of which other planted forest**  is missing.

**Priority**
Medium

sorja:
firefox_4zirqsPIkK

### DIFF
--- a/src/meta/expressionEvaluator/functions/subcategories/validatorSumSubCategoriesNotEqualToParent.ts
+++ b/src/meta/expressionEvaluator/functions/subcategories/validatorSumSubCategoriesNotEqualToParent.ts
@@ -24,7 +24,7 @@ export const validatorSumSubCategoriesNotEqualToParent: ExpressionFunction<Conte
       const valid =
         Objects.isEmpty(parentValue) ||
         Objects.isEmpty(categoryValues[categoryIndex]) ||
-        categoryValues.length !== categoryLabelKeys.length ||
+        categoryValues.some((value) => Objects.isEmpty(value)) ||
         Numbers.eqWithTolerance(parentValue, categoriesSum)
 
       if (valid) {


src/meta/expressionEvaluator/functions/subcategories/validatorSumSubCategoriesNotEqualToParent.ts
```diff
@@ -24,7 +24,7 @@ export const validatorSumSubCategoriesNotEqualToParent: ExpressionFunction<Conte
       const valid =
         Objects.isEmpty(parentValue) ||
         Objects.isEmpty(categoryValues[categoryIndex]) ||
-        categoryValues.length !== categoryLabelKeys.length ||
+        categoryValues.some((value) => Objects.isEmpty(value)) ||
```
minotogna, line 27: Why is this valid if some is empty ?

sorja: From the issue
'The validation check(above) should not be applied when one [..] is empty'

I think it makes sense, we cannot test equality if it's empty?

sorja: Or should we error also the empty one?

minotogna: no. I think that if a cell is empty then the validator should not highlight in red only the empty cell. 
If all cells are empty then all cells are valid.
But the cells with values should be highlighted in red if their sum is not equal to the parent category. 
 Otherwise I don't get the logic of this validator. 
 this must be discussed with ÖJ. or what do you think?

minotogna (approved):

PR #2935, opened 2023-08-11
Added validator: validatorSumEqualTo

### LINKED ISSUE #2922: Sum of Override value should equal 100(tab 1a)
@minotogna 
**Describe the bug**
The Sum of the **Override value** should equal 100.

**Additional expected behavior**
If the value for one Climatic domain is missing, the rule does not apply.

**Priority**
Low

sorja:
firefox_XUlluuRtL8

### DIFF
--- a/src/i18n/resources/en.js
+++ b/src/i18n/resources/en.js
@@ -1049,6 +1049,7 @@ The FRA team
     countryReportYearGreaterThanCurrentYear: 'Value should be greater or equal than {{minValue}}',
     valueNotGreaterThan: 'Value should be not greater than {{maxValue}}',
     sumNotGreaterThan: 'Sum should not exceed {{maxValue}}',
+    sumEqualTo: 'Sum ({{value}}) should equal to {{maxValue}}',
     notEmpty: 'Value cannot be empty',
     valuesAreInconsistentWithNetChange: 'Values are inconsistent with Forest area net change',
     valuesAreInconsistent1aOr1b: 'Values are inconsistent with Areas reported in tables 1a or 1b',
--- a/src/i18n/resources/en.js
+++ b/src/i18n/resources/en.js
@@ -1049,7 +1049,7 @@ The FRA team
     countryReportYearGreaterThanCurrentYear: 'Value should be greater or equal than {{minValue}}',
     valueNotGreaterThan: 'Value should be not greater than {{maxValue}}',
     sumNotGreaterThan: 'Sum should not exceed {{maxValue}}',
-    sumEqualTo: 'The sum of {{categoryLabels}} ({{categoriesSum}}) should equal to {{maxValue}}',
+    sumEqualTo: 'The sum of {{categoryLabels}} ({{categoriesSum}}) must equal to {{maxValue}}',
     notEmpty: 'Value cannot be empty',
     valuesAreInconsistentWithNetChange: 'Values are inconsistent with Forest area net change',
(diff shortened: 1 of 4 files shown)


src/meta/expressionEvaluator/functions/validatorSumEqualTo.ts
```diff
@@ -0,0 +1,31 @@
...
+      const sum = Numbers.sum(values)
+      const valid = values.some((v) => Objects.isEmpty(v)) || Objects.isEmpty(maxValue) || Numbers.eq(sum, maxValue)
+
+      if (valid) {
+        return { valid }
+      }
+
+      const messages: Array<NodeValueValidationMessage> = [
+        {
+          key: 'generalValidation.sumEqualTo',
```
minotogna, line 29: can we start making better error messages like validatorSumSubCategoriesNotEqualToParent and    validatorSumSubCategoriesNotGreaterThanParent?

sorja: I was looking at this case more closely, how should we change it?
I don't think it makes sense to differentiate each, because we are just checking that the sum === 100

minotogna: the sum of varA, varB ... must be equal to 100. that's what I meant be better message.

src/i18n/resources/en.js
```diff
@@ -1049,6 +1049,7 @@ The FRA team
     countryReportYearGreaterThanCurrentYear: 'Value should be greater or equal than {{minValue}}',
     valueNotGreaterThan: 'Value should be not greater than {{maxValue}}',
     sumNotGreaterThan: 'Sum should not exceed {{maxValue}}',
+    sumEqualTo: 'The sum of {{categoryLabels}} ({{categoriesSum}}) should equal to {{maxValue}}',
```
minotogna: I would say , **must be equal to**

src/meta/expressionEvaluator/functions/validatorSumEqualTo.ts
```diff
@@ -0,0 +1,41 @@
...
+export const validatorSumEqualTo: ExpressionFunction<Context> = {
+  name: 'validatorSumEqualTo',
+  minArity: 2,
+  executor: ({ t }) => {
+    return (
+      categoryValues: Array<string | undefined>,
+      categoryLabelKeys: Array<string>,
+      maxValue?: string
+    ): NodeValueValidation => {
+      const sum = Numbers.sum(categoryValues)
```
minotogna: Does this validator excludes empty cells ? If not , it should

sorja: image

src/i18n/resources/en.js
```diff
@@ -1049,6 +1049,7 @@ The FRA team
     countryReportYearGreaterThanCurrentYear: 'Value should be greater or equal than {{minValue}}',
     valueNotGreaterThan: 'Value should be not greater than {{maxValue}}',
     sumNotGreaterThan: 'Sum should not exceed {{maxValue}}',
+    sumEqualTo: 'The sum of {{categoryLabels}} ({{categoriesSum}}) must equal to {{maxValue}}',
```
minotogna: must be equal to

minotogna: also, please ask for translations

minotogna (approved):

PR #2966, opened 2023-08-21
Calculated cell: Fix zero value rendering issue

### LINKED ISSUE #2895: Enhancement- unknown should show as 0.00 - tab 4a
@minotogna 
When the **unknown**(d) is 0.00, it should show the value 0.00 in tab 4a instead of being empty.

Tab 1a:

Tab 4a:

sorja:
image


src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/Calculated/Calculated.tsx
```diff
@@ -7,7 +7,7 @@ import { PropsCell } from '../props'
 const Calculated: React.FC<PropsCell> = (props) => {
   const { nodeValue, row } = props
 
-  const value = nodeValue?.raw ? Numbers.format(Number(nodeValue.raw), row.props?.format?.integer ? 0 : 2) : ''
+  const value = Numbers.format(nodeValue.raw) ? Numbers.format(nodeValue.raw, row.props?.format?.integer ? 0 : 2) : ''
```
minotogna: This is strange . Why is this happening ?

sorja: I tested it:

### solution
Numbers.format(0) => 0.00

Numbers.format(null) => null

Numbers.format(undefined) => null

### issue
` const value = nodeValue?.raw ? Numbers.format(Number(nodeValue.raw), row.props?.format?.integer ? 0 : 2) : ''`

When calculation formula returns 0 ( eg. here: unknown = `forestArea - A - B` where `forestArea = 15` and `A = 5` and `B = 10` nodeValue.

( unknown = 0 = 15 - 10 - 5 )

therefore

` const value = nodeValue?.raw ( here 0 = false ) ? Numbers.format(Number(nodeValue.raw), row.props?.format?.integer ? 0 : 2) : ''`

evaluated to

`const value = ''`

even though we want

`const value = 0.00`

minotogna: ok, but the double format is strange to read. can u find a better way? Like Number.isNan ?

src/client/pages/AssessmentSection/DataTable/Table/Row/RowData/Cell/Calculated/Calculated.tsx
```diff
@@ -1,13 +1,14 @@
...
 import { Numbers } from 'utils/numbers'
+import { Objects } from 'utils/objects'
 
 import { PropsCell } from '../props'
 
 const Calculated: React.FC<PropsCell> = (props) => {
   const { nodeValue, row } = props
 
-  const value = nodeValue?.raw ? Numbers.format(Number(nodeValue.raw), row.props?.format?.integer ? 0 : 2) : ''
+  const value = Objects.isEmpty(nodeValue.raw) ? Numbers.format(nodeValue.raw, row.props?.format?.integer ? 0 : 2) : ''
```
minotogna: are you sure if it's empty you format it ?

sorja: Good catch, typo

Now it works as expected!

image

minotogna (dismissed):

minotogna (approved):

PR #2967, opened 2023-08-21
remove most of client routes + fix deep scan

minotogna:


src/client/pages/CountryHome/PanEuropeanHome/hooks/useSections.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 
+import { SectionNames } from 'meta/routes'
```
sorja, line 3: Should it be HomeSectionNames in case we have AdminSectionNames or similar?

sorja (approved):

src/client/pages/CountryHome/PanEuropeanHome/hooks/useSections.tsx
minotogna: not sure I understand. SectionNames has properties, SectionNames.Country.Home... is that what you mean? see SectionNames

PR #3045, opened 2023-09-07
Original Data Point: Fix primary forest and plantation forest

### LINKED ISSUE #3037: High priority -  NDP cell value cannot be 0
@minotogna 
**Describe the bug**
This only happens when all the area of national classes for that category are 0 (and no way to enter 0% to update the value), for example:
  NDP cell value(**... of which introduced species**) cannot be 0 when using NDP( for Saint Lucia) in tab 1b

**To Reproduce**
1. Create NDP
2.  All the national classes for that category are 0(example plantation forest)
3.  The overall shows as **empty(Dash)** instead of 0

**Expected behavior**
The value should be calculated as **zero** when all percentages for that catergory is 0%(not empty neither dash).

**Screenshots**

**Priority**
High

sorja:


src/meta/assessment/originalDataPoint/odps/calc.ts
```diff
@@ -46,20 +46,56 @@ export const calcTotalSubFieldArea = (props: {
...
+  }
+
   const nationalClasses = originalDataPoint.nationalClasses.filter(
     (nationalClass) =>
       !Objects.isNil(nationalClass.area) &&
       !Objects.isNil(nationalClass[field]) &&
       !Objects.isNil(nationalClass[subField]) &&
       !Objects.isNil(nationalClass[subSubField])
   )
+
```
sorja, line 98: Note to self:
- This is now resulting in correct results in ODP
- Revisit this code and clean it up if needed
- Important parts:
  - allowed class (is the class included in the calculation of total)
  - forestPercent: it is required for all calculations and must be larger than zero

src/meta/assessment/originalDataPoint/odps/updateNationalClass.ts
```diff
@@ -33,14 +33,21 @@ const calculateValues = (nationalClass: ODPNationalClass) => {
 
   const rowIsMaxedExtentOfForest = Numbers.eq(Numbers.sum([forestPercent, otherWoodedLandPercent]), 100)
 
+  // if forestPlantationPercent is 0, set forestPlantationIntroducedPercent to 0
+  let _forestPlantationIntroducedPercent = forestPlantationIntroducedPercent
```
minotogna, line 37: This can be ? : 
It will shorten the code

minotogna (approved):

PR #3046, opened 2023-09-07
3016 admin countries page sort columns

minotogna:
https://github.com/openforis/fra-platform/assets/1138156/380f9b59-9af9-4aae-ace8-6a2973aaa26b


sorja (approved):
👍

PR #3073, opened 2023-09-18
add DataRedisRepository.getODPYears / remove server side validations code / introduce calculation dependencies Context

minotogna:


src/server/controller/cycleData/updateDependencies/updateCalculationDependencies/context.ts
```diff
@@ -0,0 +1,57 @@
...
+import { NodeUpdate, NodeUpdates } from 'meta/data'
+
+type ConstructorProps = {
+  assessment: Assessment
+  cycle: Cycle
+  countryIso: CountryIso
+}
+
+export class Context {
+  private readonly _assessment: Assessment
```
sorja, line 12: would it make sense to use private class fields?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

sorja (approved):
👍

src/server/controller/cycleData/updateDependencies/updateCalculationDependencies/context.ts
minotogna: Sure @sorja . Thanks for this .Will update it in next pr

PR #3092, opened 2023-09-26
fix estimation years / update data table cache on clear node values 

minotogna:


src/server/controller/cycleData/index.ts
```diff
@@ -21,7 +20,6 @@ export const CycleDataController = {
   persistNodeValues,
   persistNodeValuesEstimated,
   // table data
-  getOriginalDataPointData,
```
minotogna, line 24: @sorja please keep this file in its original position in your pr to avoid conflict later on

sorja (approved):

src/server/controller/cycleData/index.ts
sorja: Added note

PR #3120, opened 2023-10-06
Fix imported values for 1b: Primary forest by climatic domain

### LINKED ISSUE #3119: 1b fix migrated data in cycle 2025 fra
**Describe the bug**
Node is missing property: calculated: true

Note:
This is true also for null values, we have imported data from Finland 2020 eg.
```
2000	"{""raw"": null}"
1990	"{""raw"": null}"
```

**To Reproduce**
1. Edit 1b second table and check total

sorja:


src/test/migrations/steps/20231006141043-step-1b-fix-migrated-values-2025-fra.ts
```diff
@@ -0,0 +1,26 @@
...
+      set value = jsonb_set(value, '{calculated}', 'true'::jsonb)
+      from ${schemaName}.table t
+               left join ${schemaName}.row r on t.id = r.table_id
+               left join ${schemaName}.col c on r.id = c.row_id
+      where t.props ->> 'name' = 'primaryForestByClimaticDomain'
+        and r.props ->> 'variableName' = 'totalPrimaryForest'
+        and c.props ->> 'colType' = 'calculated'
+        and n.row_uuid = r.uuid
+        and n.col_uuid = c.uuid
+        and n.value ->> 'calculated' is null
```
minotogna, line 24: Should this be value is not null ?

sorja: I added in issue also why we include nulls

In short, we have migrated values of format { raw: null } so it wouldn't update those in the UI otherwise
This can be seen in review finland 2025 when editing eg. 2000

https://fra-data.review.fao.org/assessments/fra/2025/FIN/sections/forestCharacteristics

minotogna (approved):

PR #3136, opened 2023-10-12
Fix context queue issue and extract abstract redis commands

sorja:
Opened separate PR to highlight changes made in this commit


src/server/controller/cycleData/updateDependencies/context/contextFactory.ts
```diff
@@ -115,7 +115,8 @@ export class ContextFactory {
     )
   }
 
-  async #initQueue(): Promise<void> {
+  async #initQueue(props: Props): Promise<void> {
+    const { includeSourceNodes } = props
```
minotogna: includeSourceNodes can be access from instance, to avoid passing props. ```const { assessment, cycle, nodeUpdates, includeSourceNodes } = this.#props```

src/test/migrations/steps/utils/updateCalculatedVariable.ts
```diff
@@ -20,18 +19,19 @@ type Props = {
...
     tableName,
     variableName,
   })
 
-  const r = (await RowRepository.getManyCache({ assessment })).find(
-    (r) => r.tableName === tableName && r.props.variableName === variableName
-  )
-
-  await redis.hset(key, rowKey, JSON.stringify(r))
+  await RowRedisRepository.updateRow({
```
minotogna: I wonder if it's more convenient (to shorten the code), to update RowRedisRepository.getRows as 

```RowRedisRepository.getRows({assessment,rowKeys:[RowCaches.getKey({tableName,variableName})], force:true})```

Note the new force param, that when true, will force updating the cache.
This also, implies to modify internal _cacheRows to accept rowKeys as optional param.

(Instead of having RowRedisRepository.updateRow and RowRepository.getOneCache)
What do you think?

sorja: Sure, I will implement this

minotogna (approved):

PR #3138, opened 2023-10-12
API for ODP Year

sorja:
Note:
API only returns given params currently

Next PR implements controller

- [x] add new action to dispatch to edit the ODP year
- [x] add new endpoint with body prevYear, nextYear, odpId (and odp if consistent with other endpoints? maybe not needed for this endpoint)

### DIFF
--- a/src/client/store/ui/originalDataPoint/actions/updateOriginalDataPointOriginalYear.ts
+++ b/src/client/store/ui/originalDataPoint/actions/updateOriginalDataPointOriginalYear.ts
@@ -1,0 +1,35 @@
+import { createAsyncThunk } from '@reduxjs/toolkit'
+import axios from 'axios'
+import { Functions } from 'utils/functions'
+
+import { ApiEndPoint } from 'meta/api/endpoint'
+import { CycleParams } from 'meta/api/request'
+import { OriginalDataPoint } from 'meta/assessment'
+
+type Props = CycleParams & {
+  originalDataPoint: OriginalDataPoint
+  id: string
+  year: string
+  targetYear: string
+}
+
+const putOriginalDataPointYear = Functions.debounce(
+  async (props: Props) => {
+    const { countryIso, assessmentName, cycleName, year, targetYear, id } = props
+
+    const params = { countryIso, assessmentName, cycleName, sectionName: 'extentOfForest' }
+    const config = { params }
+    const data = { id, year, targetYear }
+    await axios.put(ApiEndPoint.CycleData.OriginalDataPoint.year(), data, config)
+  },
+  1000,
+  'updateOriginalDataPointYear'
+)
+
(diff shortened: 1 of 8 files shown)


src/client/store/ui/originalDataPoint/actions/updateOriginalDataPointOriginalYear.ts
```diff
@@ -0,0 +1,35 @@
...
+  },
+  1000,
+  'updateOriginalDataPointYear'
+)
+
+export const updateOriginalDataPointOriginalYear = createAsyncThunk<OriginalDataPoint, Props>(
+  'originalDataPoint/year/update',
+  async (props) => {
+    putOriginalDataPointYear(props)
+    return props.originalDataPoint
```
minotogna: should update year with target year here? so that the store will be updated immediately ?

sorja: It is updated when calling the action:

```
        originalDataPoint: {
          ...originalDataPoint,
          year: Number(year),
        },

```

It was already implemented like this, so it didn't show up in the PR

https://github.com/openforis/fra-platform/pull/3138/files#diff-5750cf1e94118494411ca7dd7e3fcd20658bbd4a23eed7f077d31eee5f9b5b67R29

minotogna: not what I see in the action:
```
...         
year: originalDataPoint.year,
targetYear: year,
```

sorja: Here is the whole code, would you prefer we change the odp year in action?

```
useOnchange.tsx
---

      const commonProps = {
        assessmentName,
        cycleName,
        countryIso,
        originalDataPoint: {
          ...originalDataPoint,
          year: Number(year), <-- here we set the year to the changed year
        },
      }

      // When creating ODP, year is -1
      if (originalDataPoint.year === -1) {
        dispatch(OriginalDataPointActions.createOriginalDataPoint(commonProps))
        // Otherwise year already exists
      } else {
        const propsUpdateYear = {
          ...commonProps,
          id: originalDataPoint.id,
          year: originalDataPoint.year, <-- current odp year
          targetYear: year, <-- target odp year
        }

        dispatch(OriginalDataPointActions.updateOriginalDataPointOriginalYear(propsUpdateYear))
      }

```

```
actions/updateOriginalDataPointYear.ts:
---


export const updateOriginalDataPointOriginalYear = createAsyncThunk<OriginalDataPoint, Props>(
  'originalDataPoint/year/update',
  async (props) => {
    ..backend..
    return props.originalDataPoint <-- return the passed odp (with updated year) directly to the store
  }
)


```

minotogna (approved):

PR #3164, opened 2023-10-20
Remove unused references to status: inactive

sorja:


src/client/components/UserList/AdministrationListElement/AdministrationListElement.tsx
```diff
@@ -24,11 +22,7 @@ const AdministrationListElement: React.FC<{ user: User }> = ({ user }) => {
   const { id } = user
 
   return (
-    <tr
```
minotogna, line 27: I would keep this , for disabled users . Why remove it ?

src/client/components/UserList/CollaboratorListElement/CollaboratorListElement.tsx
```diff
@@ -37,7 +37,6 @@ const CollaboratorListElement: React.FC<{ user: User; readOnly: boolean }> = ({
       className={classNames({
         'user-list__invitation-row': isInvitation,
         expired: isInvitation && invitationExpired,
-        'user-list__inactive-user': user.status === UserStatus.inactive,
       })}
```
minotogna, line 40: Also this , won't it be part of admin page ?

sorja: This is only used in country user list and can be safely removed

PR #3182, opened 2023-10-24
Update node_ext def

### LINKED ISSUE #3181: update node_ext definition

sorja:
- [x] For each assessment and cycle: create a table node_ext (id, uuid, country_iso, type, props)


minotogna (commented):
it looks good 👍 

but tests do not pass

src/test/migrations/steps/20231024170224-step-update-node_ext-table.ts
```diff
@@ -0,0 +1,84 @@
...
+        migrateOldData`${schemaCycle}`,
+        dropOldTable`${schemaCycle}`,
+      ])};
+      ${getCreateSchemaCycleOriginalDataPointViewDDL(schemaCycle)}`
+
+      const query = wrapWithIf(schemaCycle, queries)
+
+      await client.query(query)
+    })
+  })
```
minotogna, line 83: should we recreate odp view and relative cache ?

sorja: at row 77 we recreate odp view but cache is not needed to recreate because no data changes at all
(this was also tested by comparing data before and after running generateCache)

minotogna (approved):

PR #3188, opened 2023-10-25
Contact meta

sorja:
Note:
The meta might change during upcoming PRs

This is more like a base for future and opening for discussion if needed


minotogna:
I think it's better to incrementally add what you are needing at the moment . This as no context .
What do you think ?

sorja:
> I think it's better to incrementally add what you are needing at the moment . This as no context . What do you think ?

Sure, I will open _slightly_ larger pr next

PR #3191, opened 2023-10-26
Contacts UI

sorja:
Note:
Add and remove contact will be added later

[screen-recorder-thu-oct-26-2023-15-29-26.webm](https://github.com/openforis/fra-platform/assets/5508251/81848416-63d0-4ba8-a841-086492532108)


src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }, [t])
+
+  return (
+    <DataGrid className="contacts-datagrid">
+      <DataColumn head>Appellation</DataColumn>
+      <DataColumn head>Name</DataColumn>
+      <DataColumn head>Surname</DataColumn>
+      <DataColumn head>Role</DataColumn>
+      <DataColumn head>Institution</DataColumn>
+      <DataColumn head>Contribution</DataColumn>
```
minotogna: I18n

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,101 @@
...
+                value={institution}
+                onChange={(e: React.ChangeEvent<HTMLInputElement>) => onChange(uuid, 'institution', e.target.value)}
+              />
+            </DataColumn>
+            <DataColumn> {contribution.join(', ')}</DataColumn>
+          </React.Fragment>
+        )
+      })}
+    </DataGrid>
+  )
```
minotogna: So we go for static grid ? 
We'll handle generic node ext table editing later if needed ?

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { useTranslation } from 'react-i18next'
+
+import { RoleName, Users } from 'meta/user'
+
+import { useContacts } from 'client/store/data'
+import DataGrid from 'client/components/DataGrid'
+import DataColumn from 'client/components/DataGrid/DataColumn'
+import TextInput from 'client/components/TextInput'
+import { useGetContacts } from 'client/pages/Section/Contacts/hooks/useGetContacts'
+import { useOnChange } from 'client/pages/Section/Contacts/hooks/useOnChange'
```
minotogna: For internal hooks , should we use relative imports ?

src/meta/user/contact.ts
```diff
@@ -0,0 +1,21 @@
...
+  contribution: Array<SectionName>
+  readOnly?: boolean
+}
+
+export interface Contact {
+  readonly id: number
+  readonly uuid: string
+  countryIso: CountryIso
+  props: ContactProps
+}
```
minotogna: Should these two be exported from user index ? What do you think ?

src/client/pages/Section/Contacts/Contacts.tsx
sorja: I was also thinking about this, but, let's do it when we have another use case for node ext table

(Eg. when working on datasources)

I was also thinking we could introduce types of cell for datagrid, eg. select, etc

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useGetContacts } from './hooks/useGetContacts'
+import { useOnChange } from './hooks/useOnChange'
+import Select from './Select'
+
+type Props = {
+  disabled: boolean
+}
+
+const allowedRoles = [RoleName.NATIONAL_CORRESPONDENT, RoleName.ALTERNATE_NATIONAL_CORRESPONDENT, RoleName.COLLABORATOR]
+const appellations = ['Mr.', 'Mrs.', 'Ms.', 'Other']
```
minotogna: Don't we have these translated ?

src/client/pages/Section/Contacts/Select/Select.tsx
```diff
@@ -0,0 +1,24 @@
...
+import Autocomplete from 'client/components/Autocomplete'
+
+type Props = {
+  value: string
+  onChange: (params: { label: string; value: string }) => void
+  disabled: boolean
+  items: Array<{ label: string; value: string }>
+}
+
+const Select: React.FC<Props> = (props) => {
```
minotogna: Don't we have already some generic Select / multi select component ?

src/i18n/resources/en.js
```diff
@@ -1054,43 +1054,44 @@ The FRA team
...
-    demoteToUser: 'Are you sure you want to remove Admin privileges?',
-    promoteToAdmin: 'Are you sure you want to grant Admin privileges?',
-    mandatoryFields: '* are mandatory fields',
-    disabled: 'Disabled',
+    street: 'Street address',
+    surname: 'Surname(s)',
+    title: 'Appellation',
+    whatsapp: 'Whatsapp',
+    zipCode: 'Zip code',
   },
```
minotogna, line 1100: Let's ask for translations now ?

src/client/pages/Section/Contacts/Contacts.tsx
sorja: Unfortunately not.

Other usage:

https://github.com/openforis/fra-platform/blob/7ceac138fdd847d702656a9e938dbfe2a42b109a/src/client/components/EditUserForm/EditUserForm.tsx#L97

I will translate and add an issue to migrate these values for user
I think lowercase without dot (eg. `mr`) is ok for value/key


[issue](https://github.com/orgs/openforis/projects/8/views/1?pane=issue&itemId=42860151)

src/i18n/resources/en.js
sorja: I have added missing translation (`Contribution`, other are only sorted) to the main PR todo
Hopefully it's ok I will ask it as last step, in case we need to add other (eg. Mr, Mrs...)

src/client/pages/Section/Contacts/Select/Select.tsx
sorja: We only have multiselect component, I can move this under Components/ and we can make it more generic when needed

src/client/pages/Section/Contacts/Contacts.tsx
minotogna: can u add to i18n now and ask for translations together with others?

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <DataColumn head>{t('editUser.contribution')}</DataColumn>
+
+      {contacts.map((contact) => {
+        const { id, uuid, props } = contact
+        const { role, appellation, name, surname, institution, contribution, readOnly } = props
+        const _disabled = disabled || readOnly
+
+        return (
+          <React.Fragment key={`contact_${String(id)}`}>
+            <DataColumn>
```
minotogna: ca new have an array of ColumnNodeExt with properties uuid, field, type, 
and then iterate over those to render a CellNodeExt ?

minotogna: see Cell.tsx

minotogna: here will be something like

```typescript
...
import {CellNodeExt} from 'components/common/TableNodeExt'
...
...

// somewhere:
const columns:Array<ColumnNodeExt> = [
{uuid:'---', type:'text', colName:'title'}
...
]

...
{contacts.map((contact) => {
....
    {columns.map(column=><CellNodeExt column={column} datum={??} value={??} />)} //pass what's useful, datum or value or ??
...    
```

sorja: Introduced `TableNodeExt` to handle the looping - I think it looks a bit cleaner like this and is slightly more dynamic
Let me know if you prefer to loop in `Contacts.tsx` instead

src/client/components/Select/Select.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useMemo } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import Autocomplete from 'client/components/Autocomplete'
+
+type Props = {
+  value: string
+  onChange: (params: { label: string; value: string }) => void
```
minotogna: extract common type Option ? {label, value}

minotogna: I would also leave the responsibility of i18n messages to the caller

src/client/components/TableNodeExt/TableNodeExt.tsx
```diff
@@ -0,0 +1,58 @@
...
+      })}
+
+      {data.map(({ uuid, props: row }) => {
+        return (
+          <React.Fragment key={uuid}>
+            {columns.map((column) => {
+              const { type, colName, items } = column
+              const Component = Components[type]
+              return (
+                <DataColumn key={`${uuid}_${colName}_data`}>
```
minotogna: I'd rather have a single common CellNodeExt that also wraps the DataColumn. and inside have a record of type, component (see Cell.tsx)

src/client/components/TableNodeExt/types.ts
```diff
@@ -0,0 +1,12 @@
+export enum ColumnNodeExtType {
```
minotogna: do this belong to meta/assessment (nodeExt) ?

src/meta/user/user.ts
```diff
@@ -2,14 +2,21 @@ import { Lang } from 'meta/lang'
 
 import { RoleName, UserRole } from './index'
 
+export enum UserTitles {
```
minotogna: I would use singular for enum. UserTitle

src/meta/nodeExt/nodeExt.ts
```diff
@@ -0,0 +1,4 @@
+export type NodeExt<Datum = any> = {
```
minotogna: also in meta/assessment (nodeExt). or better meta/nodeExt ?

sorja: I am working with the backend currently, and I would like to separate `nodeExt` from currently existing `cycleData` or any other, what do you think?
This would be:
`meta/nodeExt`
`api/nodeExt`
`api/nodeExt/createContact`
`controller/nodeExt/createContact` (need to specify contact: activitylog, etc)
`repository/nodeExt/create` (can be generic, repository doesn't need to know what kind of entity we create)

Initially I started working under cycle data but it is very populated, thoughts?

minotogna: it is part of cycle data though. isn't it in cycle schema ?

minotogna: we can improve the CycleDataController later on to be something like:
CycleDataController.OriginalDataPoint.getOne() or  CycleDataController.Node.getTableData().. something like this?

src/client/components/Select/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from 'client/components/Select/Select'
```
minotogna: better local path

src/client/components/Select/Select.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Option } from 'client/components/Select/types'
+
+type Props = {
+  value: string
+  onChange: (params: Option) => void
+  disabled: boolean
+  items: Array<Option>
+}
+
+const Select: React.FC<Props> = (props) => {
```
minotogna, line 24: I think simple select should not be autocomplete . we can talk about this

src/client/components/Select/Select.tsx
```diff
@@ -0,0 +1,45 @@
...
+  onChange: (newValue: Option) => void
+  disabled: boolean
+  options: Array<Option>
+}
+
+const Select: React.FC<Props> = (props) => {
+  const { value, onChange, disabled, options: _options } = props
+  const { t } = useTranslation()
+
+  const options = _options.map(
```
minotogna: I would also memoize options

src/client/components/TableNodeExt/CellNodeExt/Components.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+
+import Select from 'client/components/Select'
+import TextInput from 'client/components/TextInput'
+
+export const Components: Record<string, React.FC<any>> = {
```
minotogna: better to be Record<ColumnNodeExtType, React.FC<CellNodeExtProps>>. any is not suggested here.
plus this can be inline in CellNodeExt directly.  see Cell.tsx

src/client/components/TableNodeExt/CellNodeExt/CellNodeExt.tsx
```diff
@@ -0,0 +1,33 @@
...
+  column: ColumnNodeExt
+  onChange: (uuid: string, colName: string, value: any) => void
+  disabled: boolean
+}
+
+const CellNodeExt: React.FC<Props> = (props: Props) => {
+  const { datum: row, column, onChange, disabled } = props
+  const { uuid } = row
+
+  const { type, colName, options } = column
```
minotogna: pass the column to Component directly. options are only in Select

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,55 @@
...
+    })
+  }, [])
+
+  const columns: Array<ColumnNodeExt> = [
+    { type: ColumnNodeExtType.select, colName: 'appellation', header: 'editUser.title', options: optionsAppellation },
+    { type: ColumnNodeExtType.text, colName: 'name', header: 'editUser.name' },
+    { type: ColumnNodeExtType.text, colName: 'surname', header: 'editUser.surname' },
+    { type: ColumnNodeExtType.select, colName: 'role', header: 'editUser.role', options: optionsRole },
+    { type: ColumnNodeExtType.text, colName: 'institution', header: 'editUser.institution' },
+    { type: ColumnNodeExtType.multiselect, colName: 'contributions', header: 'editUser.contributions' },
```
minotogna: maybe better to have ColumnNodeExt.props and add options only to column type select.

src/meta/nodeExt/nodeExt.ts
```diff
@@ -0,0 +1,19 @@
...
+  select = 'select',
+  multiselect = 'multiselect',
+}
+
+export enum NodeExtType {
+  contact = 'contact',
+  node_ext = 'node_ext',
+}
+
+export type NodeExt<Datum = any> = {
```
minotogna: not any here. maybe never or {} ?

src/client/components/Select/Select.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { value, onChange, disabled, column } = props
+  const { options: _options } = column.props
+  const { t } = useTranslation()
+
+  const options = useMemo(
+    () =>
+      _options.map(
+        (item: Option): Option => ({
+          ...item,
+          label: t(item.label),
```
minotogna: better to use label format (see metadata)

src/client/components/TableNodeExt/CellNodeExt/CellNodeExt.tsx
```diff
@@ -0,0 +1,60 @@
...
+type CellProps = {
+  value: string | Array<string>
+  onChange: (newValue: string | Option) => void
+  disabled: boolean
+  options?: Array<any>
+  column: ColumnNodeExt
+}
+
+const components: Record<string, React.FC<CellProps>> = {
+  text: (props) => (
```
minotogna: again, this should not be a function, see Cell.tsx

src/client/components/TableNodeExt/CellNodeExt/CellNodeExt.tsx
```diff
@@ -0,0 +1,60 @@
...
+  column: ColumnNodeExt
+  onChange: (uuid: string, colName: string, value: any) => void
+  disabled: boolean
+}
+
+type CellProps = {
+  value: string | Array<string>
+  onChange: (newValue: string | Option) => void
+  disabled: boolean
+  options?: Array<any>
```
minotogna: why options is a prop of cell? and why any ?

src/client/components/TableNodeExt/CellNodeExt/CellNodeExt.tsx
```diff
@@ -0,0 +1,60 @@
...
+type Props = {
+  datum: Record<string, any>
+  column: ColumnNodeExt
+  onChange: (uuid: string, colName: string, value: any) => void
+  disabled: boolean
+}
+
+type CellProps = {
+  value: string | Array<string>
+  onChange: (newValue: string | Option) => void
```
minotogna: onChange should not know if it's a select or text.

src/meta/nodeExt/nodeExt.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export enum ColumnNodeExtType {
+  text = 'text',
+  select = 'select',
+  multiselect = 'multiselect',
+}
+
+export enum NodeExtType {
+  contact = 'contact',
+  node_ext = 'node_ext',
```
minotogna, line 11: what is type node_ext ? I don't think it's what we want ?

src/client/components/TableNodeExt/CellNodeExt/CellNodeExt.tsx
sorja: This is only function because the underlaying component is a class component and we would need to refactor it
Should we create a new component instead or leave as is for now?

sorja: Create inner component that wraps that
src/client/components/TableNodeExt/CellNodeExt/TextInput
src/client/components/TableNodeExt/CellNodeExt/Select

src/meta/user/contact.ts
```diff
@@ -0,0 +1,16 @@
+import { SectionName } from 'meta/assessment'
+import { NodeExt } from 'meta/nodeExt'
+import { RoleName } from 'meta/user/userRole'
+
+export interface ContactProps {
+  index: number
```
sorja: rowindex

src/meta/nodeExt/nodeExt.ts
sorja: This will be changed in next PR, will not affect anything here

minotogna (approved):

PR #3218, opened 2023-11-10
3213 paneuropean linked nodes shows wrong data  1

minotogna:
- [x] generate meta cache for all assessments at once
- [x] add cross assessments dependant/dependency to meta cache


src/server/controller/assessment/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
```diff
@@ -16,56 +17,69 @@ const includesVariableCache = (variables: Array<VariableCache>, variable: Variab
...
+      const dependencies =
+        type === 'calculations'
+          ? AssessmentMetaCaches.getCalculationsDependencies(propsDependency)
+          : AssessmentMetaCaches.getValidationsDependencies(propsDependency)
+
+      if (!_includesVariableCache(dependencies, variable)) {
+        const path = [type, 'dependencies', tableName, variableName]
+        Objects.setInPath({ obj: metaCache, path, value: [...dependencies, variable] })
+      }
     }
```
minotogna, line 83: those two methods are similar. I will think later on to reduce redundant code

src/utils/objects/setInPath.ts
```diff
@@ -11,7 +11,7 @@ export const setInPath = (params: { obj: any; path: string[]; value: any; exclud
     if (i === path.length - 1) {
       objCurrent[pathPart] = value
     } else {
-      if (!Object.prototype.hasOwnProperty.call(objCurrent, pathPart)) {
+      if (!objCurrent[pathPart]) {
```
sorja, line 14: can this fail if it's nullish value eg

{
 [pathPart]: false
 }

minotogna: If that's the case , then the caller has a wrong format or wrong path

sorja (approved):

PR #3257, opened 2023-11-27
3251 - File access selector pt1. UI

sorja:
# UI

https://github.com/openforis/fra-platform/assets/5508251/63c7de4e-438f-4044-8599-b97d52c0bac5

Next PR:
- Backend
  - Don't allow changing to private if file is in use
  - Add middleware/controller checks for viewFile


src/client/components/PageLayout/Toolbar/Toolbar.scss
```diff
@@ -32,7 +32,7 @@
   width: 1px;
   height: 60%;
   background: radial-gradient($ui-border, lighten($ui-border, 12%));
-  margin: 0 $spacing-xxs;
+  margin: auto $spacing-xxs;
```
minotogna: Why this ?

sorja: resolves issue alignin the component in the Links.tsx without changing the old usage

minotogna: but that's the toolbar. what does it has to do with links ? @sorja I'm really wondering, if we did something wrong to fix

minotogna (approved):

PR #3258, opened 2023-11-27
3237 - deprecate DataGrid,VerticallyGrowingTextInput,TextInput + add TextArea and InputText

minotogna:
part 1


src/client/components/Inputs/InputText/InputText.scss
```diff
@@ -0,0 +1,26 @@
+@import 'src/client/style/partials';
+
+.input-text {
+  $inputOffset: 2px;
+  background-color: transparent;
+  border: 1px solid transparent;
```
sorja, line 6: transparent border?

src/client/components/Inputs/InputText/InputText.tsx
```diff
@@ -0,0 +1,25 @@
+import './InputText.scss'
```
sorja, line 1: what do you think naming them eg
```suggestion
import `'./InpuText.module.scss
```

sorja (approved):

PR #3271, opened 2023-11-30
Authorize file access change

sorja:


src/server/api/file/index.ts
```diff
@@ -52,7 +52,7 @@ export const FileApi = {
     express.delete(ApiEndPoint.File.Assessment.one(), AuthMiddleware.requireEditCountryFile, removeAssessmentFile)
     express.put(
       ApiEndPoint.File.Assessment.updateAccess(),
```
minotogna: can we also rename updateAccess to access (name and path). the put itself defines update. in endpoints it's better not to have words like get/update/delete/put/post/...

minotogna (approved):

PR #3276, opened 2023-12-01
Select component

sorja:
Example usage:

image


src/client/components/Select/Select.scss
```diff
@@ -0,0 +1,34 @@
+@import 'src/client/style/partials';
+
```
minotogna: can we put the Select inside components/Inputs folder ?

src/client/components/Select/Select.scss
```diff
@@ -0,0 +1,34 @@
...
+.fra-select {
```
minotogna: can we avoid fra prefix ?

src/client/components/Select/Select.tsx
```diff
@@ -0,0 +1,68 @@
...
+      return item.value === value
+    })
+  }, [options, value])
+
+  const _onChange = useCallback(
+    (option: TranslatedOption) => {
+      onChange(option?.value ?? null)
+    },
+    [onChange]
+  )
```
minotogna: can we have a hook called useValues that returns {defaultValue, options} ?

src/client/components/Select/Select.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Option } from './types'
+
+type Props = {
+  value: string
+  onChange: (newValue: string) => void
+  disabled: boolean
+  options: Array<Option>
+}
+
+type TranslatedOption = {
```
minotogna: does ReactSelect have its own type for this? I'm not a big fan of TranslatedOption type.
Also, TranslatedOptions type is not really needed.

sorja: I agree, I don't like the TranslatedOptions also, I tried to look in to the docs ( https://react-select.com/typescript ) and migration docs ( https://react-select.com/upgrade )


OptionTypeBase | no replacement | Options can be any type (if using getOptionValue and getOptionLabel) so there's no longer a base type for options
-- | -- | --


---

I liked the idea of passing `Label` type to the Select component, maybe we have to pass directly the translated instead?

PR #3287, opened 2023-12-05
3221 the writing box is still hard to read if writing a long text

### LINKED ISSUE #3221: The writing box is still hard to read if writing a long text
@minotogna 

The main text output reading part of the chatbox is nice. However, the issue is that it is still very difficult to read the entire input text in the writing box, as shown in the screenshot and video.
image

https://github.com/openforis/fra-platform/assets/99216009/6eb2bfc1-4efc-4162-b197-682de2271fc3

**Describe the solution you'd like**
The writing box should be freely stretchable within the frame of the chatbox. Alterative solution is to allow the writing box inside the chatbox to be proportionally stretched along with the main chatbox.

**Additional context**
If possible to implement, the chatbox should be allowed to move to any place on the screen, avoiding covering important data, as currently, the entire chatbox is stretchable _only in a small area._

yaguzmang:
As seen in #3221, the comment text area is hard to read for long comments. So:

- Added a TextArea component, with a initial maximum height. When the Topic is resized, the maximum height of the comment is adjusted accordingly (to 50 % of the available space).

TODO: Make the initial maximum height depend on the initial available space. Doing this by making the Resizable componet be a forward ref was not possible because the ReResizable component from the library is expecting a different type of ref "LegacyRef<Resizable>".

https://github.com/openforis/fra-platform/assets/41337901/57886122-31a2-46f8-af1a-a1872f64cb33


src/client/components/MessageCenter/Topic/Topic.tsx
```diff
@@ -157,10 +165,12 @@ const Topic: React.FC<TopicProps> = (props) => {
...
             (Users.isAdministrator(user) || Users.isReviewer(user, countryIso, cycle)))) && (
           <div className="topic-form">
-            <textarea
+            <TextArea
+              style={{ maxHeight: commentTextAreaMaxHeight }}
               value={message}
               placeholder={i18n.t('review.writeComment')}
               onChange={(e) => setMessage(e.target.value)}
+              rows={2}
             />
```
sorja: Would it make sense to instead use Resizable for `<div className="topic-form">`?
And only allow resizing from top

Example:

```suggestion
<Resizable
enable={{top: true}}
className="topic-form">
            <TextArea
              style={{ maxHeight: commentTextAreaMaxHeight }}
              value={message}
              placeholder={i18n.t('review.writeComment')}
              onChange={(e) => setMessage(e.target.value)}
              rows={2}
            />
```

I like you are using the new TextArea component though

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -1,10 +1,13 @@
 import './TextArea.scss'
 import React, { forwardRef, TextareaHTMLAttributes, useEffect, useImperativeHandle, useRef } from 'react'
 
-type Props = Pick<TextareaHTMLAttributes<HTMLTextAreaElement>, 'disabled' | 'onChange' | 'onPaste' | 'value'>
+type Props = Pick<
+  TextareaHTMLAttributes<HTMLTextAreaElement>,
+  'disabled' | 'onChange' | 'onPaste' | 'value' | 'placeholder' | 'style' | 'rows'
```
minotogna: I think we should leave out passing style from outside . Pass instead specific maxHeight prop (default null) and use Math.max when calculating dynamic height on value change (see inner use effect  ).

And please , keep the alphabetical order .

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -18,11 +21,13 @@ const TextArea = forwardRef<HTMLTextAreaElement, Props>((props, outerRef) => {
...
     <textarea
       className="textarea validation-error-sensitive-field"
+      style={style}
+      placeholder={placeholder}
       disabled={disabled}
       onChange={onChange}
       onPaste={onPaste}
       ref={textAreaRef}
-      rows={1}
+      rows={rows ?? 1}
```
minotogna: Better and more consistent to always use defaultProps

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -1,10 +1,13 @@
...
 
-type Props = Pick<TextareaHTMLAttributes<HTMLTextAreaElement>, 'disabled' | 'onChange' | 'onPaste' | 'value'>
+type Props = Pick<
+  TextareaHTMLAttributes<HTMLTextAreaElement>,
+  'disabled' | 'onChange' | 'onPaste' | 'value' | 'placeholder' | 'style' | 'rows'
+>
 
 const TextArea = forwardRef<HTMLTextAreaElement, Props>((props, outerRef) => {
-  const { disabled, onChange, onPaste, value } = props
+  const { disabled, onChange, onPaste, value, placeholder, style, rows } = props
```
minotogna: Always keep alphabetical order 😉

src/client/components/MessageCenter/Topic/Topic.tsx
```diff
@@ -123,6 +130,7 @@ const Topic: React.FC<TopicProps> = (props) => {
       maxWidth={800}
       maxHeight="100%"
       className="topic"
+      onResize={handleTopicResize}
```
minotogna, line 131: I'm thinking that it might be better to keep the textarea  height responsibility only within textarea component .

To do that , check if you can on resize here emit a window resize event .
The textarea on mount listen to this event and updates the height accordingly (on unmount unlisten to the event).

The textarea maxHeight must be a fixed number , proportional to the minimum height of this component 

Looking at the video , when resizing too small the topic opened seems to be resized too much . To fix this , we could check if the topic  can have minimum height I.e. it can't be resized vertically below a certain threshold).

What do you think ?

src/client/components/MessageCenter/Topic/Topic.tsx
minotogna: Do we really need resizable component here ? Can we avoid it ?

src/client/components/Inputs/TextArea/TextArea.tsx
yaguzmang: I added the maxHeight prop and fixed the alphabetical order, thanks.

src/client/components/Inputs/TextArea/TextArea.tsx
yaguzmang: Added it for rows and maxHeight, thanks.

src/client/components/Inputs/TextArea/TextArea.tsx
yaguzmang: Sure. Done.

src/client/components/MessageCenter/Topic/Topic.tsx
yaguzmang: It would have conflicts with the logic that is in the TextArea, as it grows when you are typing. We could scenarios where, for example, the user lowers the size and expects it to be set, but as soon as they type it would grow again.

src/client/components/MessageCenter/Topic/Topic.tsx
yaguzmang: You mean having something in Topic like:

```
const handleTopicResize = (_e, _direction, _elementRef: HTMLElement, _delta) => {
  window.dispatchEvent(new Event('resize')); // Trigger window resize event
};
```

And then listen to this event in the TextArea component. Right?
What I don't fully get is why we would need to listen to that event if the maxHeight is fixed and not dynamic? if the maxHeight will always stay the same, resizing the component won't affect the height, only typing would (at least in this case that the maxHeight would be low, around 200px because the minHeight of the topic is 300px).

yaguzmang: I implemented in the latest commit something similar to what you described. This is the behaviour:

https://github.com/openforis/fra-platform/assets/41337901/ef5d92f0-dfac-437c-a2e6-c8ee563a07bd

src/client/components/MessageCenter/Topic/Topic.tsx
```diff
@@ -157,10 +165,12 @@ const Topic: React.FC<TopicProps> = (props) => {
           (topic.status === MessageTopicStatus.resolved &&
             (Users.isAdministrator(user) || Users.isReviewer(user, countryIso, cycle)))) && (
           <div className="topic-form">
-            <textarea
-              value={message}
-              placeholder={i18n.t('review.writeComment')}
+            <TextArea
+              maxHeight={commentTextAreaMaxHeight}
```
minotogna: We could even inline this value . Used only here

src/client/components/MessageCenter/Topic/Topic.tsx
```diff
@@ -115,6 +116,12 @@ const Topic: React.FC<TopicProps> = (props) => {
     }
   }, [assessment, cycle, topic, dispatch])
 
+  const commentTextAreaMaxHeight = 200
+
+  const handleTopicResize = () => {
```
minotogna: In this case preferably with useCallback

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -1,31 +1,64 @@
...
+    window.addEventListener('resize', handleResize)
+
+    return () => {
+      window.removeEventListener('resize', handleResize)
+    }
+  }, [textAreaRef, maxHeight])
+
+  useEffect(() => {
+    adjustTextAreaHeight(textAreaRef, maxHeight)
+  }, [disabled, textAreaRef, value, maxHeight])
```
minotogna, line 18: Better to extract all the code above in an inner hook useResize that takes one param Props with {textareaRef, maxHeight}. We import inner hooks with relative path

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -1,31 +1,64 @@
...
-type Props = Pick<TextareaHTMLAttributes<HTMLTextAreaElement>, 'disabled' | 'onChange' | 'onPaste' | 'value'>
+type AdditionalProps = {
+  maxHeight?: number | null
+}
+
+type Props = Pick<
+  TextareaHTMLAttributes<HTMLTextAreaElement>,
+  'disabled' | 'onChange' | 'onPaste' | 'placeholder' | 'rows' | 'value'
+> &
+  AdditionalProps
```
minotogna: We could even inline this

src/client/components/MessageCenter/Topic/Topic.tsx
minotogna: > You mean having something in Topic like:
> 
> 
> 
> ```
> 
> const handleTopicResize = (_e, _direction, _elementRef: HTMLElement, _delta) => {
> 
>   window.dispatchEvent(new Event('resize')); // Trigger window resize event
> 
> };
> 
> ```
> 
> 
> 
> And then listen to this event in the TextArea component. Right?
> 
> What I don't fully get is why we would need to listen to that event if the maxHeight is fixed and not dynamic? if the maxHeight will always stay the same, resizing the component won't affect the height, only typing would (at least in this case that the maxHeight would be low, around 200px because the minHeight of the topic is 300px). 
> 
> 
> 
> 
> 
> 

Because the textarea height is dynamically calculated based on the content. This would fix a bug that on window resize event , making the content in overflow hidden . 
(Check table 1a , NDP click blue year column , and see references section in read only mode .

Also , I'm thinking : on window resize event fired , and the content is higher than the new calculated height , should be make the content overflow scroll 😅?

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -1,31 +1,64 @@
...
+
     textArea.style.height = 'auto'
-    textArea.style.height = `${textArea.scrollHeight}px`
-  }, [disabled, textAreaRef, value])
+    const newHeight = Math.min(textArea.scrollHeight, maxHeight)
+    textArea.style.height = `${newHeight}px`
+  }
+
+  useEffect(() => {
+    const handleResize = () => {
```
minotogna: We could even throttle this ? 250 ms? Whqt do you think ? In case check Functions.throttle

src/client/components/Inputs/TextArea/TextArea.tsx
yaguzmang: I will add it. We will also need `value` so that it resizes when typing if possible.

src/client/components/Inputs/TextArea/TextArea.tsx
yaguzmang: Yes, great suggestion. I added it with 200 ms because I also implemented enabling the scrolling and it was a little slow if the user tried to use it immediately.  

https://github.com/openforis/fra-platform/assets/41337901/103160ee-3a80-4fd6-943f-15faacc02b96

src/client/components/Inputs/TextArea/hooks/useResize.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect } from 'react'
+
+import { Functions } from 'utils/functions'
+
+const useResize = (
+  textAreaRef: React.MutableRefObject<HTMLTextAreaElement>,
```
minotogna: We usually prefer to pass one single object called props with type Props defined above the method . Check other inner hooks .
Sorry 😅for all those comments , but better to start now to get used to the code convention .

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -1,31 +1,38 @@
...
+      rows={rows}
       value={value}
     />
   )
 })
 
+TextArea.defaultProps = {
+  maxHeight: null,
+  rows: 1,
+} as Props
```
minotogna, line 36: Here 'as Props' might not be needed

src/client/components/Inputs/TextArea/TextArea.tsx
```diff
@@ -1,31 +1,38 @@
...
-import React, { forwardRef, TextareaHTMLAttributes, useEffect, useImperativeHandle, useRef } from 'react'
+import React, { forwardRef, TextareaHTMLAttributes, useImperativeHandle, useRef } from 'react'
 
-type Props = Pick<TextareaHTMLAttributes<HTMLTextAreaElement>, 'disabled' | 'onChange' | 'onPaste' | 'value'>
+import useResize from './hooks/useResize'
+
+type Props = Pick<
+  TextareaHTMLAttributes<HTMLTextAreaElement>,
+  'disabled' | 'onChange' | 'onPaste' | 'placeholder' | 'rows' | 'value'
+> & { maxHeight?: number | null }
```
minotogna: Or null might not be needed in this case . maxHeight is already optional as its default props can be undefined . 


| null has a different meaning . For example a value defined by the user if reset by the user can be null , otherwise if never defined , can be undefined .
In this context , it doesn't apply.

minotogna (commented):
This pr looks good 👍 just a few last details to follow our code convention

src/client/components/Inputs/TextArea/hooks/useResize.ts
yaguzmang: Yes, I am all for keeping the codebase consistent 🙌

src/client/components/Inputs/TextArea/TextArea.tsx
yaguzmang: If I don't add it I get `error    defaultProp "rows" has no corresponding propTypes declaration  react/default-props-match-prop-types` in that line. I think this is because eslint doesn't pick up that rows gets its type declaration from 
```
Pick<
  TextareaHTMLAttributes<HTMLTextAreaElement>,
  'disabled' | 'onChange' | 'onPaste' | 'placeholder' | 'rows' | 'value'
>
```
(If I add a default value for `placeholder`I get the same error)

src/client/components/Inputs/TextArea/TextArea.tsx
yaguzmang: Good point, thanks. I removed it.

src/client/components/Inputs/TextArea/TextArea.tsx
minotogna: okkkk strange. I tried as well. let's keep it as it is

minotogna:
@sorja this pr looks good to me. do you have any additional comments ?

sorja:
@minotogna 
Looks good!

minotogna (approved):

PR #3308, opened 2023-12-12
3174 - Contacts: Prefill with existing users

sorja:
Next pr: create new contact

- [x] API/Contacts/getMany: Return current users (`prefilled`)

Noteble changes:
- Moved `Contact` to `meta/cycleData/contact`
- Split `Contact` state to `prefilled` and `contacts` where `prefilled` are current platform users

Video note:
- Names and institution changed
- `contacts` (writable) shows existing users as mock data in second table (reality currently returns empty)

https://github.com/openforis/fra-platform/assets/5508251/e309310f-e8a0-45fd-897d-eeee36e8fd5d


src/server/controller/cycleData/contact/getMany.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  return client.tx(async (t) => {
+    const statuses = [UserStatus.active]
+    const withProps = true
+    const roles = [RoleName.COLLABORATOR, RoleName.NATIONAL_CORRESPONDENT, RoleName.ALTERNATE_NATIONAL_CORRESPONDENT]
+    const getManyProps = { assessment, cycle, countryIso, withProps, statuses, roles }
+    const users = await UserRepository.getMany(getManyProps, t)
+    const sections = await MetadataController.getSections({ assessment, cycle }, t)
+
+    const createContact = (user: User): Contact => {
```
sorja: Should I move this somewhere?

sorja: maybe `Contacts.userToContact` or similar?

src/client/components/TableNodeExt/TableNodeExt.tsx
```diff
@@ -14,24 +14,27 @@ type Props = {
   data: Array<NodeExt>
   disabled: boolean
   onChange: (uuid: string, colName: string, value: any) => void
+  gridTemplateColumns: string
+  header?: boolean
```
minotogna: Can we have only one data grid ? Data returned from controller is already merged ? What do you think ? (Adding read only to contacts from db)

sorja: Do we need to persist `contact.props.rowIndex` or can we just use the `map`s second argument for operations?

Maybe initial version will omit `rowIndex`?

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -61,7 +61,17 @@ const Contacts: React.FC<Props> = (props: Props) => {
     },
   ]
 
-  return <TableNodeExt disabled={disabled} onChange={onChange} columns={columns} data={contacts} />
+  const gridTemplateColumns = `12ch repeat(${columns.length - 1}, 1fr)`
+
+  return (
+    <TableNodeExt
+      gridTemplateColumns={gridTemplateColumns}
```
minotogna: can we sort properties alphabetically ?

src/client/components/TableNodeExt/TableNodeExt.tsx
```diff
@@ -42,7 +45,7 @@ const TableNodeExt = (props: Props) => {
                   uuid={uuid}
```
minotogna: can we sort properties alphabetically ?

src/client/store/data/actions/getContacts.ts
```diff
@@ -1,89 +1,15 @@
...
-  },
-]
+import { Contact } from 'meta/cycleData'
 
 export const getContacts = createAsyncThunk<Array<Contact>, CycleDataParams>('extNode/get/contacts', async (_props) => {
-  // TODO: implement backend endpoint
-  // const params = { countryIso, assessmentName, cycleName }
-  // const config = { params }
-  // const contacts = await axios.get(ApiEndPoint.CycleData.ExtNode.contacts(), config)
+  const { countryIso, assessmentName, cycleName } = _props
```
minotogna: I think better to call props rather than _props ?

src/meta/cycleData/contact/contact.ts
```diff
@@ -2,15 +2,22 @@ import { SectionName } from 'meta/assessment'
 import { NodeExt } from 'meta/nodeExt'
 import { RoleName } from 'meta/user/userRole'
 
-export interface ContactProps {
-  rowIndex: number
-  role: RoleName.NATIONAL_CORRESPONDENT | RoleName.ALTERNATE_NATIONAL_CORRESPONDENT | RoleName.COLLABORATOR
+export interface ContactValue {
   appellation: string
+  contributions: Array<SectionName>
```
minotogna: should these be section uuid ? so that we always use the translations from i18n ?

src/meta/cycleData/contact/contacts.ts
```diff
@@ -0,0 +1,29 @@
+import { Contact, ContactProps, ContactValue } from './contact'
+
+const initialProps: ContactProps = {
+  rowIndex: -1,
+  readOnly: false,
+}
+
+const initialValue: ContactValue = {
```
minotogna: better to call this placeholder that has props and value inline ? (those are used only here)

src/server/api/cycleData/index.ts
```diff
@@ -123,5 +124,14 @@ export const CycleDataApi = {
     // Activities
     express.get(ApiEndPoint.CycleData.activities(), AuthMiddleware.requireView, getActivities)
     express.get(ApiEndPoint.CycleData.activitiesCount(), AuthMiddleware.requireView, getActivitiesCount)
+
+    // ext node
```
minotogna, line 128: I;m wondering if we should start splitting the api initialization for cycle data as well ? what do you think? or later on ?

src/server/controller/cycleData/contact/getMany.ts
```diff
@@ -0,0 +1,54 @@
...
```
minotogna: I think this should be added to a utility Contacts.fromUser(user:User)=>Contact . export this utility from meta/cycleData (like Contact). what do you think ?

src/server/repository/public/user/getMany.ts
```diff
@@ -25,6 +25,8 @@ export const getMany = async (
 
     limit?: number
     offset?: number
+
+    withProps?: boolean
```
minotogna: is this needed ?

sorja: We don't fetch props when fetching user list, only when fetching single users
When fetching contacts, we include it

Should we always fetch it?

minotogna: as agreed, let's have a repository method only

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -61,7 +61,17 @@ const Contacts: React.FC<Props> = (props: Props) => {
     },
   ]
 
-  return <TableNodeExt disabled={disabled} onChange={onChange} columns={columns} data={contacts} />
+  const gridTemplateColumns = `12ch repeat(${columns.length - 1}, 1fr)`
```
minotogna: maybe this can be inline . or use Memo. useMemo for columns as well. I would add an inner hook useColumns

src/meta/cycleData/contact/contacts.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+  value: {
+    appellation: '',
+    contributions: [],
+    institution: '',
+    name: '',
+    role: undefined,
+    surname: '',
+  },
+} as Contact
```
minotogna: is 'as Contact' needed ?

src/meta/cycleData/contact/contacts.ts
```diff
@@ -0,0 +1,52 @@
...
+  return contacts.filter((contact) => contact.uuid)
+}
+
+const fromUser = (props: { countryIso: CountryIso; cycle: Cycle; user: User }): Contact => {
+  const { countryIso, cycle, user } = props
+  const userRole = Users.getRole(user, countryIso, cycle)
+
+  const { uuid } = user
+  const { name, surname, title: _appellation } = user.props
+  // TODO: Migration step to fix user.props.appellation
```
minotogna: what's this TODO ?

sorja:
```
select distinct props ->> 'title' as user_title from public.users where props ->> 'title' is not null
```

| user\_title |
| :--- |
| Other |
| Ms |
| Mr |

src/server/api/cycleData/index.ts
sorja: lets do later

src/client/pages/Section/Contacts/hooks/useColumns.ts
```diff
@@ -0,0 +1,49 @@
+import { useMemo } from 'react'
+
+import { ColumnNodeExtType } from 'meta/nodeExt'
+import { RoleName, Users, UserTitle } from 'meta/user'
+
+const allowedRoles = [RoleName.NATIONAL_CORRESPONDENT, RoleName.ALTERNATE_NATIONAL_CORRESPONDENT, RoleName.COLLABORATOR]
+const appellations = Object.values(UserTitle)
+
+export const useColumns = () => {
```
minotogna: it is always suggested to add return type

src/client/pages/Section/Contacts/hooks/useColumns.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const optionsAppellation = useMemo(() => {
+    return appellations.map((appellation) => {
+      const label = { key: `editUser.${appellation}` }
+      const value = appellation
+      return { label, value }
+    })
+  }, [])
+
+  return useMemo(
```
minotogna: and specify types for useMemo (3 in this file)

src/meta/user/users.ts
```diff
@@ -202,6 +218,7 @@ export const Users = {
 
   getRolesAllowedToEdit,
   getRolesAllowedToView,
+  getUserSections,
```
minotogna: since the util is about user, User prefix is not needed. I'd rather call this method getAllowedSections

src/meta/user/users.ts
```diff
@@ -188,6 +189,21 @@ const isPersonalInfoRequired = (user: User, role: UserRole<RoleName, any>) => {
...
+const getUserSections = (props: { user: User; countryIso: CountryIso; cycle: Cycle }): Array<string> => {
+  const { countryIso, cycle, user } = props
+  const userRole = getRole(user, countryIso, cycle)
+
+  if (isCollaborator(user, countryIso, cycle)) {
+    const { permissions }: { permissions: CollaboratorPermissions } = userRole
+    if (!Objects.isEmpty(permissions?.sections)) {
+      return Object.keys(permissions.sections)
+        .map((sectionUuid) => sectionUuid)
+        .sort((a, b) => a.localeCompare(b))
```
minotogna: are these uuids ? sorting should be by section anchor

src/server/controller/cycleData/contact/getMany.ts
```diff
@@ -0,0 +1,28 @@
...
+  cycle: Cycle
+  countryIso: CountryIso
+}
+
+type Returned = Array<Contact>
+
+export const getMany = async (props: Props, client: BaseProtocol = DB): Promise<Returned> => {
+  const { assessment, cycle, countryIso } = props
+
+  return client.tx(async (t) => {
```
minotogna: there's no real need to start a transaction here. there are just two read functions.

src/server/controller/cycleData/contact/getMany.ts
```diff
@@ -0,0 +1,28 @@
...
+
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+  countryIso: CountryIso
+}
+
+type Returned = Array<Contact>
+
+export const getMany = async (props: Props, client: BaseProtocol = DB): Promise<Returned> => {
```
minotogna: is client passed from outside now? if not, let's remove it

src/server/repository/adapter/contact.ts
```diff
@@ -0,0 +1,25 @@
+import { Objects } from 'utils/objects'
+
+import { Contact } from 'meta/cycleData'
+
+export const ContactAdapter = ({ permissions, ...contact }: any): Contact => {
```
minotogna: do not use any

src/server/repository/adapter/contact.ts
```diff
@@ -0,0 +1,25 @@
...
+    contact.value.contributions = ['none']
+  }
+
+  if (!Objects.isEmpty(permissions.sections) && typeof permissions.sections !== 'string') {
+    // eslint-disable-next-line no-param-reassign
+    contact.value.contributions = Object.keys(permissions.sections)
+  }
+
+  // eslint-disable-next-line no-param-reassign
+  contact.value.appellation = contact.value.appellation?.toLowerCase()
```
minotogna: why is this needed ?

src/server/repository/adapter/contact.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  return Objects.camelize(contact)
```
minotogna: I'm wondering, can the query build already the object you need ?

src/server/repository/public/user/getContacts.ts
```diff
@@ -0,0 +1,35 @@
...
+          jsonb_build_object(
+                  'role',ur.role,
+                  'appellation',u.props ->> 'title',
+                  'name',u.props ->> 'name',
+                  'surname',u.props ->> 'surname',
+                  'institution',ur.props ->> 'organization'
+          ) as value,
+          ur.permissions
+      from public.users u right join public.users_role ur on (u.id = ur.user_id)
+      where ur.role in ('COLLABORATOR','NATIONAL_CORRESPONDENT','ALTERNATE_NATIONAL_CORRESPONDENT') and ur.country_iso = $3 and ur.cycle_uuid = $2 and ur.assessment_id = $1
```
minotogna: put each and on a new line. more readable

src/server/repository/adapter/contact.ts
sorja: currently a workaround
, see previous comment
ref:

```
select distinct props ->> 'title' as user_title from public.users where props ->> 'title' is not null
```

| user\_title |
| :--- |
| Other |
| Ms |
| Mr |

src/server/repository/adapter/contact.ts
sorja: I tried but I couldn't find a clean solution for it, any suggestions?

The possible values for permissions:
`{}`
`{ sections: none }`
`{ sections: all }`
`{ sections: Array<SectionUUID> }`

sorja: One problem is that jsonb_object_keys filters out all the `{}` values which then contributes to a more messy code to work around that (using `coalesce`)

sorja: So in this solution we can see the workaround:
we dont populate the contact object `contributions` field but we pass an extra value in the row that we handle in the adapter called `permissions` which is 1:1 from user_role and we just handle it here

minotogna: well, now that I think better at this, our boss mentioned 'Contributions' which is different from 'Permissions'.
we should show the sections where the user has effectively edited some data, not the once he/she has access to.
So I think we should join the country activity log to get that information. (and then maybe handle this in a query directly). what do you think ?

src/server/repository/public/user/getContacts.ts
```diff
@@ -0,0 +1,39 @@
...
+                  'role',ur.role,
+                  'appellation',u.props ->> 'title',
+                  'name',u.props ->> 'name',
+                  'surname',u.props ->> 'surname',
+                  'institution',ur.props ->> 'organization'
+          ) as value,
+          ur.permissions
+      from public.users u right join public.users_role ur on (u.id = ur.user_id)
+      where
+              ur.role in ('COLLABORATOR','NATIONAL_CORRESPONDENT','ALTERNATE_NATIONAL_CORRESPONDENT')
```
minotogna: I would use our enum to avoid misspelled problems

src/server/repository/public/user/getContacts.ts
```diff
@@ -0,0 +1,39 @@
...
+          jsonb_build_object('readOnly', true) as props,
+          jsonb_build_object(
+                  'role',ur.role,
+                  'appellation',u.props ->> 'title',
+                  'name',u.props ->> 'name',
+                  'surname',u.props ->> 'surname',
+                  'institution',ur.props ->> 'organization'
+          ) as value,
+          ur.permissions
+      from public.users u right join public.users_role ur on (u.id = ur.user_id)
```
minotogna: safer to use left join

src/server/repository/adapter/contact.ts
sorja: Yes, this sounds correct

Should we handle this in separate PR?
In the PR we can:
- Introduce the multiselect UI component
- Populate data for existing users

minotogna (approved):

PR #3321, opened 2023-12-15
3174 - Contacts: NodeExt migration

sorja:


src/test/migrations/steps/20231215110302-step-update-node_ext-table.ts
```diff
@@ -0,0 +1,84 @@
...
+        createTable`${schemaCycle}`,
+        migrateOldData`${schemaCycle}`,
+        getCreateSchemaCycleOriginalDataPointViewDDL(schemaCycle),
+        dropOldTable`${schemaCycle}`,
+      ].join('\n')
+
+      const query = wrapWithIf(schemaCycle, queries)
+
+      await client.query(query)
+    })
```
minotogna, line 88: I would also regenerate cache for fra 2020/2025

sorja: Sure, added
Not sure if this changes anything as the data should be 1:1 unchanged

minotogna (approved):

PR #3342, opened 2023-12-21
3174 - Contacts: Delelete contact

sorja:


minotogna:
@sorja can we have a video ?

sorja:
@minotogna 
Sure, sent it to you

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -23,7 +24,7 @@ const Contacts: React.FC<Props> = (props: Props) => {
   const contacts = useContacts({ canEdit })
   const onChange = useOnChange()
   const { columns, fields } = useColumns()
-  const gridTemplateColumns = useMemo(() => `12ch repeat(${fields.length - 1}, 1fr)`, [fields.length])
+  const gridTemplateColumns = useMemo(() => `12ch repeat(${fields.length - 1}, 1fr) 4ch`, [fields.length])
```
minotogna: can we add the delete button only in edit mode ? plus I would use px , since it's a button, not text

src/client/pages/Section/Contacts/Delete/hooks/useOnClick.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { useAppDispatch } from 'client/store'
+import { DataActions } from 'client/store/data'
+import { useSectionRouteParams } from 'client/hooks/useRouteParams'
+
+export const useOnClick = (contact: Contact) => {
+  const { assessmentName, cycleName, countryIso, sectionName } = useSectionRouteParams<CountryIso>()
+  const { t } = useTranslation()
+  const { uuid } = contact
+  const dispatch = useAppDispatch()
```
minotogna: it's much nicer when the code is in order.
hooks first, new line before the return, return type, type to useCallback.

src/client/store/data/slice.ts
```diff
@@ -148,6 +149,17 @@ export const dataSlice = createSlice({
         contacts.push(contactUpdate)
       }
     })
+
+    builder.addCase(deleteContact.pending, (state, action) => {
```
minotogna, line 153: can we disable  the delete user button,
disable editing the user when action is in progress?

and remove the user when action is fulfilled ?

src/server/repository/assessmentCycle/countrySummary/_lastEditActivities.ts
```diff
@@ -2,6 +2,7 @@ import { ActivityLogMessage } from 'meta/assessment'
 
 export const activitiesLastEdit = [
   ActivityLogMessage.contactUpdate,
```
minotogna, line 5: do we need here contact create as well ?

minotogna: what about the country activity log ?

src/client/store/data/slice.ts
minotogna: Actually it's ok to remove it immediately. But if we want to follow this approach , on action  rejected user must be added again .

Also on the server , following row index users must be decremented

src/server/repository/assessmentCycle/countrySummary/_lastEditActivities.ts
sorja: Country ActivityLog checks this file:
`src/server/repository/assessmentCycle/countryActivityLog/_common/acceptedMessages.ts`

I have added a `todo` in the main PR to check all of the following working correctly:
- Contact create [persist whole contact]
- Contact update [persist updated nodeExt]
- Contact delete [persist uuid]

src/client/store/data/slice.ts
sorja: Decrementing is good note - missed this one
Still the rowIndex feels a bit funky in my opinion, lets see

minotogna: It looks funky because there's no reordering feature atm . But we might add it later on and we're prepared for it .
How would we handle that without rowIndex ?

sorja: It's true - now that I used it obviously makes sense, and it's handy.
Updating the rowIndex though doesn't look pretty, but luckily we don't have to stare at it:
`set props = jsonb_set(props, '{rowIndex}', ((props->>'rowIndex')::int - 1)::text::jsonb)`

src/client/pages/Section/Contacts/Delete/Delete.scss
```diff
@@ -0,0 +1,12 @@
...
+
+div.contacts__grid__cell-delete {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-right: 1px solid $ui-border;
+
+  svg {
+    color: $ui-destructive;
+  }
```
minotogna, line 11: (I'm just keeping a note for myslef) create generic grid column delete and common button delete ?

src/client/pages/Section/Contacts/Delete/Delete.tsx
```diff
@@ -0,0 +1,32 @@
...
+const Delete: React.FC<Props> = (props: Props) => {
+  const { disabled, contact } = props
+
+  const onClick = useOnClick(contact)
+
+  if (disabled) return <div />
+
+  return (
+    <DataCell className="contacts__grid__cell-delete" review>
+      <button className="btn-s btn-link-destructive" onClick={onClick} type="button">
```
minotogna: When if it's disabled , what if we disable the button instead of returning empty div ?

src/client/store/data/slice.ts
minotogna: It looks nice to me 😅😅, simple and pretty straightforward.

src/server/repository/assessmentCycle/nodeExt/removeContact.ts
```diff
@@ -0,0 +1,48 @@
...
+  )
+
+  const { countryIso, type, props: contactProps } = deletedContact
+  const { rowIndex } = contactProps
+
+  // After deleting a contact, update all the other contacts' row_index for the same country_iso and type
+  await client.query(
+    `
+         update ${schemaCycle}.node_ext
+         set props = jsonb_set(props, '{rowIndex}', ((props->>'rowIndex')::int - 1)::text::jsonb)
```
minotogna, line 36: Do we save rowIndex as number ?

src/client/pages/Section/Contacts/Delete/Delete.scss
sorja: My thoughts too but wanted to keep this PR explicit and avoid refactoring too much

src/server/repository/assessmentCycle/nodeExt/removeContact.ts
sorja: db `{"rowIndex": 2}`

minotogna: are u storing it as int when doing `((props->>'rowIndex')::int - 1)::text` ?

src/client/pages/Section/Contacts/Delete/Delete.tsx
minotogna: can we do it in this pr ?

sorja: Should we return empty div on `contact.props.readOnly: true` or just avoid rendering?
Sent you a screenshot if we only disable it

src/server/repository/assessmentCycle/nodeExt/removeContact.ts
sorja: Double checked:

Current solution working as expected:
image

If we cast it to numeric, the error is given:
```
cannot cast type numeric to jsonb
```

So it is working as expected

minotogna (approved):

PR #3350, opened 2023-12-24
3174 - create button

sorja:
- [x] Disable button after click  https://github.com/openforis/fra-platform/pull/3350/commits/04117efb1ce266b2108f8c82fda75b9ec4fb2633
- [x] Update backend (2 endpoints?) https://github.com/openforis/fra-platform/pull/3350/commits/dc09f74c701ebca8a42c923845dfcf471b557bb9

image


sorja:
@minotogna should we keep one endpoint or separate to create and update?

minotogna:
> @minotogna should we keep one endpoint or separate to create and update?

endpoint can be the same, but one for put, one for post. does it answer the question ?

src/client/store/data/actions/updateContact.ts
```diff
@@ -29,12 +21,6 @@ const _update = async (props: Props): Promise<void> => {
...
-export const upsertContact = createAsyncThunk<void, Props>('contact/upsert', async (props) => {
-  const { contact } = props
-
-  if (contact.placeholder) {
-    Functions.debounce(_create, 5_000, 'createContact')(props)
-  } else {
-    Functions.debounce(_update, 5_000, `updateContact${props.field}`)(props)
-  }
+export const updateContact = createAsyncThunk<void, Props>('contact/update', async (props) => {
+  Functions.debounce(_update, 1_000, `updateContact${props.field}`)(props)
```
minotogna: The debounced function key must include contact uuid as well

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -4,15 +4,16 @@ import { useTranslation } from 'react-i18next'
 
 import { Labels } from 'meta/assessment'
 
+import { useContacts } from 'client/store/data'
 import { DataCell, DataGrid } from 'client/components/DataGrid'
 import CellNodeExt from 'client/components/TableNodeExt/CellNodeExt'
-import Delete from 'client/pages/Section/Contacts/Delete'
-import { useGridTemplateColumns } from 'client/pages/Section/Contacts/hooks/useGridTemplateColumns'
 
-import { useContactsData } from './hooks/useContactsData'
```
minotogna, line 12: What happened to this ?

src/meta/assessment/activityLog.ts
```diff
@@ -9,6 +9,7 @@ export enum ActivityLogMessage {
   assessmentStatusUpdate = 'assessmentStatusUpdate',
   assessmentUpdate = 'assessmentUpdate',
   contactCreate = 'contactCreate',
+  contactFieldCreate = 'contactFieldCreate',
```
minotogna: One contact create is enough I suppose. It's not useful to persist history for each single filed , what do you think ?

src/client/pages/Section/Contacts/NewContactButton/NewContactButton.scss
```diff
@@ -0,0 +1,11 @@
+@import 'src/client/style/partials';
+
+.contacts__button-add-container {
+  display: flex;
+  justify-content: flex-end;
+
+  button.btn-add-contact {
```
minotogna: this can be moved out of here, plus use same convention then above container

src/client/pages/Section/Contacts/Contacts.tsx
```diff
@@ -6,13 +6,14 @@ import { Labels } from 'meta/assessment'
...
-import Delete from 'client/pages/Section/Contacts/Delete'
-import { useGridTemplateColumns } from 'client/pages/Section/Contacts/hooks/useGridTemplateColumns'
 
 import { useContactsData } from './hooks/useContactsData'
 import { useColumns, useFields } from './hooks/useDefinitions'
 import { useGetContacts } from './hooks/useGetContacts'
+import { useGridTemplateColumns } from './hooks/useGridTemplateColumns'
 import { useOnChange } from './hooks/useOnChange'
+import Delete from './Delete'
+import NewContactButton from './NewContactButton'
```
minotogna: I would use same name convention. either both have Button or none.
Plus, I would start using the convention to import components with absolute imports. only internal hook with relative import. does it work for you? or do we want to find a different convention ? as long as we are consistent and stop commenting always the same thing.

src/client/store/data/hooks/useContacts.ts
```diff
@@ -1,28 +1,16 @@
...
-  const contacts = useAppSelector((state) => state.data.contacts[assessmentName]?.[cycleName]?.[countryIso] ?? [])
-
-  return useMemo<Returned>(() => {
-    const lastPlaceholder = contacts.at(-1)?.placeholder
-
-    return canEdit && !lastPlaceholder
-      ? [...contacts, Contacts.newContact({ countryIso, rowIndex: contacts.filter((c) => !c.props.readOnly).length })]
-      : contacts
-  }, [canEdit, contacts, countryIso])
+  return [...contacts].sort((a, b) => a.props.rowIndex - b.props.rowIndex)
```
minotogna: they should be sorted server side, in the query

src/client/store/data/slice.ts
```diff
@@ -168,17 +168,26 @@ export const dataSlice = createSlice({
...
       }
     })
+
+    builder.addCase(createContact.fulfilled, (state, action) => {
+      const { assessmentName, cycleName, countryIso } = action.meta.arg
+      const contactAction = action.payload
+
+      const contacts = state.contacts[assessmentName][cycleName][countryIso]
+      contacts.push(contactAction)
+      contacts.sort((a, b) => a.props.rowIndex - b.props.rowIndex)
```
minotogna: as before, sorting doesn't apply here

src/client/pages/Section/Contacts/Contacts.tsx
sorja: My logic has been to use relative imports in sub components (eg here Contacts is parent, relative import for hooks, components that are in same folder or one level higher)

Should we instead use absolute imports?
We can use it for everything also, which is fine

minotogna: my thinking is that components can be reused and can be independent.
inner hooks are strictly business logic to their own component. and when they are relatively imported, everything seems more ordered.
what do you think?

src/client/pages/Section/Contacts/Contacts.scss
```diff
@@ -9,3 +9,13 @@
 .contacts__subTitle {
   margin-bottom: $spacing-xxs;
 }
+
+.contacts__button-add-container {
```
minotogna: can this be in new contact component ? also, better to use consistent names with components. 
each style must be place where it's used and use a consist name. what do you think ?
Also, I think better to call the component CreateContact ? and the style contacts__create-container ?

src/client/pages/Section/Contacts/Contacts.scss
```diff
@@ -9,3 +9,13 @@
...
+  display: flex;
+  justify-content: flex-end;
+
+  button.contacts__button-add {
```
minotogna: also, inner style, when using absolute css name, should be outside

minotogna (approved):

PR #3356, opened 2023-12-27
3174 - Refresh country activity log

sorja:
image

Links point to section 0


src/meta/assessment/section.ts
```diff
@@ -4,6 +4,11 @@ import { Label } from './label'
 
 export type SectionName = string
 
+export enum SectionNames {
```
sorja, line 7: Enum is called `SectionNames` because of the conflicting `SectionName`

Should we change this or leave?

src/meta/assessment/activityLogs.ts
```diff
@@ -94,9 +104,29 @@ const getLabelSection = (props: { cycle: Cycle; section?: SubSection; activity:
...
+}
+
+const getSectionLink = (props: GetSectionLinkProp): string => {
+  const { countryIso, assessmentName, cycleName, sectionName: sectionNameProp } = props
+
+  let sectionName = sectionNameProp
+
+  if (sectionName === SectionNames.contacts) {
+    sectionName = SectionNames.contactPersons
+  }
```
minotogna: isn't the target section introduction ?

minotogna: I would probably add a Record, to make it more flexible and to avoid several if

sorja: Not sure what you refer to with target section, but the `sectionName` for 0 is `contactPersons`

https://fra-data.review.fao.org/assessments/fra/2025/FIN/sections/contactPersons

sorja: > I would probably add a Record, to make it more flexible and to avoid several if

Sure, I can add this

minotogna: > Not sure what you refer to with target section, but the `sectionName` for 0 is `contactPersons`
> 
> https://fra-data.review.fao.org/assessments/fra/2025/FIN/sections/contactPersons

aaaa okkk forgot about that. let's just add the Record then.

minotogna (approved):

PR #3360, opened 2023-12-28
3296 - Datasources: Update select components

### LINKED ISSUE #3296: Text of  Type of data source - National Data disappeared
@minotogna 

**Describe the bug**
If it is not too time-consuming, please check why the text of the '**National Forest Inventory (NFI)**' has suddenly disappeared under the '**Type of Data Source - National Data**' for  Japan Table 1a, as no one has edited it.

**Screenshots**

sorja:
Note: Components are not refactored to avoid conflict with [3274-national-data-data-sources-grid-layout-error](https://github.com/openforis/fra-platform/tree/3274-national-data-data-sources-grid-layout-error)

**Only underlying `Select` component updated**

https://github.com/openforis/fra-platform/assets/5508251/8122f784-3193-47c7-8f1d-dfa2d6f69686

### DIFF
--- a/src/client/components/Inputs/Select/Select.tsx
+++ b/src/client/components/Inputs/Select/Select.tsx
@@ -16,6 +16,9 @@ const Select: React.FC<SelectProps> = (props) => {
   const value = useValue(props)
   const onChange = useOnChange(props)
 
+  // Keep menu open when selecting an option in multi-select mode
+  const closeMenuOnSelect = !isMulti
+
   return (
     <ReactSelect
       classNames={{
@@ -33,6 +36,7 @@ const Select: React.FC<SelectProps> = (props) => {
         singleValue: () => 'select__singleValue',
         valueContainer: () => 'select__valueContainer',
       }}
+      closeMenuOnSelect={closeMenuOnSelect}
       components={{ ClearIndicator, DropdownIndicator, IndicatorsContainer, IndicatorSeparator: null }}
       isClearable
       isDisabled={disabled}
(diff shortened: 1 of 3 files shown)


src/client/components/Inputs/Select/Select.tsx
```diff
@@ -16,6 +16,9 @@ const Select: React.FC<SelectProps> = (props) => {
   const value = useValue(props)
   const onChange = useOnChange(props)
 
+  // Keep menu open when selecting an option in multi-select mode
+  const closeMenuOnSelect = !isMulti
```
minotogna: should this be by default ? can u check if it looks good as well in contacts contributions ? in that case we can avoid to pass the prop closeMenuOnSelect and always keep it open. what do you think ?

minotogna: ps 3274 is merged I think

sorja: It looks weird for single select where you expect it to close on select
On multi select it makes sense - the user might want to select more than one option, especially when selecting many options it might be rather annoying always to reopen

https://github.com/openforis/fra-platform/assets/5508251/523172af-03a6-4bf6-8a15-b8a50ea33b26

src/client/components/Inputs/Select/Select.tsx
```diff
@@ -33,6 +36,7 @@ const Select: React.FC<SelectProps> = (props) => {
         singleValue: () => 'select__singleValue',
         valueContainer: () => 'select__valueContainer',
       }}
+      closeMenuOnSelect={closeMenuOnSelect}
```
minotogna: This can be inline

minotogna (approved):

PR #3364, opened 2023-12-28
3260 - fetch odp values as deps if odp in use

### LINKED ISSUE #3260: Table 3a Total Forest Area invalid validation error triggered
There seems to be some bug in table 3a in the Argentina report. The platform reports an error message in Table 3a mentioning that the total forest area for 2015 is higher than the forest area report in Table 1a, but it’s not.

sorja:
Issue caused by omitting odp data

### DIFF
--- a/src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
+++ b/src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
@@ -71,7 +71,6 @@ export const useDependencies = (props: Props): Returned => {
 
         internal.tableNames.add(tableName)
         staticDependencies[tableName]?.forEach((t) => internal.tableNames.add(t))
-
         if (withOdp) {
           internal.tableNames.add(TableNames.originalDataPointValue)
           internal.tableWithOdp = tableName
@@ -80,25 +79,8 @@ export const useDependencies = (props: Props): Returned => {
         if (table.calculationDependencies) {
           addDependencies(Object.values(table.calculationDependencies))
         }
-
         if (canEdit && table.validationDependencies) {
-          const dependenciesArray = Object.values(table.validationDependencies)
-          addDependencies(dependenciesArray)
-
-          // add originalDataPointValue to dependencies if ODPs are enabled (in 1b)
-          const found =
-            forestCharacteristicsUseOdp &&
-            dependenciesArray.find((deps) =>
-              deps.find(
-                (variableCache) =>
-                  variableCache.tableName === TableNames.extentOfForest ||
-                  variableCache.tableName === TableNames.forestCharacteristics
-              )
-            )
-
-          if (found) {
-            addDependencies([[{ tableName: TableNames.originalDataPointValue, variableName: undefined }]])
-          }
+          addDependencies(Object.values(table.validationDependencies))
         }
       })
     })
--- a/src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
+++ b/src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
@@ -71,10 +71,9 @@ export const useDependencies = (props: Props): Returned => {
 
         internal.tableNames.add(tableName)
         staticDependencies[tableName]?.forEach((t) => internal.tableNames.add(t))
-        if (withOdp) {
-          internal.tableNames.add(TableNames.originalDataPointValue)
-          internal.tableWithOdp = tableName
-        }
+
+        if (forestCharacteristicsUseOdp) internal.tableNames.add(TableNames.originalDataPointValue)
+        if (withOdp) internal.tableWithOdp = tableName
 
         if (table.calculationDependencies) {
           addDependencies(Object.values(table.calculationDependencies))
--- a/src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
+++ b/src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
@@ -72,14 +72,33 @@ export const useDependencies = (props: Props): Returned => {
         internal.tableNames.add(tableName)
         staticDependencies[tableName]?.forEach((t) => internal.tableNames.add(t))
 
-        if (forestCharacteristicsUseOdp) internal.tableNames.add(TableNames.originalDataPointValue)
-        if (withOdp) internal.tableWithOdp = tableName
+        if (withOdp) {
+          internal.tableNames.add(TableNames.originalDataPointValue)
+          internal.tableWithOdp = tableName
+        }
 
         if (table.calculationDependencies) {
           addDependencies(Object.values(table.calculationDependencies))
         }
+
         if (canEdit && table.validationDependencies) {
-          addDependencies(Object.values(table.validationDependencies))
+          const dependenciesArray = Object.values(table.validationDependencies)
+          addDependencies(dependenciesArray)
+
+          // add originalDataPointValue to dependencies if ODPs are enabled (in 1b)
+          const found =
+            forestCharacteristicsUseOdp &&
+            dependenciesArray.find((deps) =>
+              deps.find(
+                (variableCache) =>
+                  variableCache.tableName === TableNames.extentOfForest ||
+                  variableCache.tableName === TableNames.forestCharacteristics
+              )
+            )
+
+          if (found) {
+            addDependencies([[{ tableName: TableNames.originalDataPointValue, variableName: undefined }]])
+          }
         }
       })
     })
--- a/src/client/pages/Section/hooks/useGetTableData/useGetTableData.ts
+++ b/src/client/pages/Section/hooks/useGetTableData/useGetTableData.ts
@@ -3,7 +3,6 @@ import { useEffect } from 'react'
 import { CountryIso } from 'meta/area'
 
 import { useAppDispatch } from 'client/store'
-import { useCountry } from 'client/store/area'
 import { DataActions } from 'client/store/data/slice'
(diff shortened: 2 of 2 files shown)


src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
```diff
@@ -71,10 +71,9 @@ export const useDependencies = (props: Props): Returned => {
 
         internal.tableNames.add(tableName)
         staticDependencies[tableName]?.forEach((t) => internal.tableNames.add(t))
-        if (withOdp) {
-          internal.tableNames.add(TableNames.originalDataPointValue)
-          internal.tableWithOdp = tableName
-        }
+
+        if (forestCharacteristicsUseOdp) internal.tableNames.add(TableNames.originalDataPointValue)
+        if (withOdp) internal.tableWithOdp = tableName
```
sorja: if we are using ODPs from section 1b, fetch `originalDataPointValue`

src/client/pages/Section/hooks/useGetTableData/useDependencies.ts
```diff
@@ -71,10 +71,9 @@ export const useDependencies = (props: Props): Returned => {
 
         internal.tableNames.add(tableName)
         staticDependencies[tableName]?.forEach((t) => internal.tableNames.add(t))
-        if (withOdp) {
-          internal.tableNames.add(TableNames.originalDataPointValue)
-          internal.tableWithOdp = tableName
-        }
+
+        if (forestCharacteristicsUseOdp) internal.tableNames.add(TableNames.originalDataPointValue)
```
minotogna: Isn't this covered by withOdp variable ?

sorja: Maybe, I am trying to understand the logic:

```
        const withOdp =
          tableName === TableNames.extentOfForest ||
          (tableName === TableNames.forestCharacteristics && forestCharacteristicsUseOdp)

        if (withOdp) {
           internal.tableWithOdp = tableName
           internal.tableNames.add(TableNames.originalDataPointValue)
         }
```

This is obviously not working when looking for tables outside of group `[1a, 1b]`.

My reasoning for the above correction (correct me if I am wrong):

```
        if (forestCharacteristicsUseOdp) internal.tableNames.add(TableNames.originalDataPointValue)
        if (withOdp) internal.tableWithOdp = tableName
```

Here we have divided the `if` clause in 2 parts:

First part:

If we have enabled ODPs in table `1b`, we will include the `originalDataPointValue` table in the `internal` deps.
This affects the tables fetched: it will include the ODP values and the validations should work as expected.

Second part:
We leave `tableWithOdp` inside `withOdp` clause:
This is used here:

```
        if (tableWithOdp && canEdit) {
          dispatch(DataActions.getNodeValuesEstimations({ ...propsFetch, sectionName, tableName: tableWithOdp }))
          dispatch(DataActions.getODPLastUpdatedTimestamp({ ...propsFetch, sectionName }))
        }
```

It seems we only use this information for fetching estimations and last update of any ODP

It seems tableWithOdp can be one of group `[extentOFForest, forestCharacteristics]` if I didn't misinterpret something


Let me know

@minotogna

minotogna: @sorja isn't the problem related to table 1a? therefore in table 1a we should fetch odp ?

sorja: 3a

With old code, if you visited first 3a you saw incorrect validation, but if you go 3a -> 1a -> 3a you will see correct validation

minotogna: @sorja my mistake. I was thinking this is related to another issue.

the withOdp should be something like (see pseudo code for last or condition):

```
const withOdp =
          tableName === TableNames.extentOfForest ||
          (tableName === TableNames.forestCharacteristics && forestCharacteristicsUseOdp) ||
          (table has 1a or 1b validation dependency)
```          
would this fix the issue ?

sorja: I will investigate deeper, but are we sure we want to set this variable in that case?

```
        if (withOdp) internal.tableWithOdp = tableName
```

minotogna: I need to check the code

minotogna:
```
const withOdp =
          tableName === TableNames.extentOfForest ||
          (tableName === TableNames.forestCharacteristics && forestCharacteristicsUseOdp) ||
          (table has 1a or 1b validation dependency)
```
+          
```
if (withOdp) {
          internal.tableNames.add(TableNames.extentOfForest)
          internal.tableNames.add(TableNames.forestCharacteristics)
          internal.tableNames.add(TableNames.originalDataPointValue)
          internal.tableWithOdp = tableName
        }
```        
would this fix the issue ?

minotogna: it's not event that.
the 1a dependencies should be added here line 82
```
        if (canEdit && table.validationDependencies) {
          addDependencies(Object.values(table.validationDependencies))
        }
        ```
isn't it?

Comment (sorja) [reply src/client/pages/Section/hooks/useGetTableData/useDependencies.ts]: updated;

if odps are enabled and dependencies contain either `extentOfForest` or `forestCharacteristics` we add `originalDataPointValue`

File: src/client/pages/Section/hooks/useGetTableData/useDependencies.ts [hunk 3]
```diff
@@ -79,8 +80,25 @@ export const useDependencies = (props: Props): Returned => {
         if (table.calculationDependencies) {
           addDependencies(Object.values(table.calculationDependencies))
         }
+
         if (canEdit && table.validationDependencies) {
-          addDependencies(Object.values(table.validationDependencies))
+          const dependenciesArray = Object.values(table.validationDependencies)
```
minotogna: If 1a (which is a dependency table to fetch) , is fetched with mergeODp to true, fetching odp could be unnecessary.
Can you verify this ?

sorja: I think this is the correct solution!

Change here:
https://github.com/openforis/fra-platform/pull/3364/commits/3e6b9aed4eb7b0f86d832aba4fe1dabbcc771e15

src/client/pages/Section/hooks/useGetTableData/useGetTableData.ts
```diff
@@ -15,20 +16,21 @@ export const useGetTableData = (props: Props) => {
...
   const dispatch = useAppDispatch()
-  const { assessmentName, cycleName, countryIso: _countryIso } = useCountryRouteParams()
+  const { assessmentName, cycleName, countryIso } = useCountryRouteParams<CountryIso>()
   const canEdit = useCanEdit(sectionName)
   const { print } = useIsPrintRoute()
   const dependencies = useDependencies(props)
 
-  const countryIso = _countryIso as CountryIso
+  const country = useCountry(countryIso)
+  const mergeOdp = country.props.forestCharacteristics?.useOriginalDataPoint
```
minotogna: This can be removed with 
mergeODp:!tableWithOdp

Can u please verify ?

sorja: Then we go back to previous implementation and the current commit is redundant:

```
        const withOdp =
          tableName === TableNames.extentOfForest ||
          (tableName === TableNames.forestCharacteristics && forestCharacteristicsUseOdp)
```

We need to change this to 
```
        const withOdp =
          tableName === TableNames.extentOfForest ||
          (tableName === TableNames.forestCharacteristics && forestCharacteristicsUseOdp) ||
          <pseudo: table 1a or 1b as dep>
```

But then later in the same file the `originalDataPointValue` is added, making the current `mergeOdp` solution redundant?

```
        if (withOdp) {
          internal.tableNames.add(TableNames.originalDataPointValue)
          internal.tableWithOdp = tableName
        }
```

Let me know

minotogna: well why?
have you tried ?

minotogna: if tableWithOdp exists, then we are either in 1a or 1b, therefore mergeOdp must be false.
the opposite, if we are not in 1a or 1b, then when fetching data (and 1a and 1b) are within the tables to fetch then with mergeOdp to true, tables 1a and 1b will include odp then, validations might work/
have you tried ?

sorja: I tried - it doesn't work if we disable ODPs from section 1b (button: 'Don't use national data points')
The problem is that now we would always fetch the ODP data (when in section 3a for example) and validate against that

minotogna: that's strange. the withOdp should check that, maybe there's another bug?

sorja: _This is a note to keep track of investigation:_

So server must be incorrect in this case:
1. Disable ODPs from section 1b
2. Navigate to section 3a
3. You will still receive mergedODP data if mergeOdp flag is true (which it will be when we use `!tableWithOdp`

minotogna (approved):

PR #3368, opened 2023-12-29
3367 - Contacts: Check/remove pending invitations and disabled users + Sort users

### LINKED ISSUE #3367: Contacts: Check/remove pending invitations and disabled users + Sort users
Thank you for introducing this new contact feature; the structure is quite clear to me.

After some trials, two concerns have arisen that may be worth discussing:

1.The appearance of the person follows the 'first join, first show' rule (see attached 1). This issue could potentially be resolved by displaying individuals in the order of NC, ANC, C.
This arrangement should apply to contacts (role) added manually, as exemplified in (attached 2), where the NC has not signed up for the platform and needs to be added later.
The appearance by role NC/ANC/C will be done in the print view (non edit mode). 

`
In the edit/view mode, it will be done for the contacts prefetched from existing users only. 
The manually entered contacts will be positioned after existing users, otherwise, while editing new contacts, there might be unpleasant shifting of rows that will confuse the editor. 
`

2. Allow reviewers/ Admin to edit and remove existing contacts. 
The concern here is that someone might sign up for the platform but does not enter any personal information (see attached 1). Additionally, in some cases, users may forget their passwords and be unable to retrieve their accounts. Consequently, they might create a new account with a different role (see attached 2).

`
Users are required to enter their information. Plus, if users forget their password they can reset it through the platform. Users cannot be registered twice with same email address.
Admin has always the possibility to update missing information or disable users.
We will remove from contacts, disabled users and users that have not accepted the invitation.
`

minotogna:


src/client/pages/Section/Contacts/hooks/useContactsData.ts
```diff
@@ -4,12 +4,29 @@ import { useTranslation } from 'react-i18next'
...
 
 import { Contact, ContactField, Contacts } from 'meta/cycleData'
+import { RoleName } from 'meta/user'
 
 import { useContacts } from 'client/store/data'
 import { useIsPrintRoute } from 'client/hooks/useIsRoute'
 
 type Returned = Array<Contact>
 
+const compareContacts = (contactA: Contact, contactB: Contact): number => {
```
sorja, line 14: should this be in `meta/contacts`

sorja (approved):

src/client/pages/Section/Contacts/hooks/useContactsData.ts
minotogna: Not sure @sorja . I left it there because it's print view specific . If we'll need it somewhere else we'll move it then. What do you think ?

sorja: Sounds good

PR #3371, opened 2024-01-01
3318 - Enable global and filter countries in country selector

yaguzmang:
Enabling global in the Area selector and filtering by user countries in the country selector:

https://github.com/openforis/fra-platform/assets/41337901/a82a69cb-e4f2-4963-96e0-8c2c03717cb8

Global if the user has a reviewer role:

https://github.com/openforis/fra-platform/assets/41337901/64b46f66-6b94-4b9a-8c76-baa11bdedb3c

### DIFF
--- a/src/client/components/PageLayout/Toolbar/Toolbar.tsx
+++ b/src/client/components/PageLayout/Toolbar/Toolbar.tsx
@@ -1,5 +1,5 @@
 import './Toolbar.scss'
-import React from 'react'
+import React, { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import MediaQuery from 'react-responsive'
 
@@ -28,12 +28,13 @@ const Toolbar: React.FC = () => {
   const { print } = useIsPrintRoute()
   const user = useUser()
   const isInGeoPage = useIsGeoRoute()
+  const isAReviewer = useMemo<boolean>(() => Users.isAReviewer(user, cycle), [user, cycle])
 
   if (print) return null
 
   const isCountry = Areas.isISOCountry(countryIso)
   const isAdmin = Users.isAdministrator(user)
-  const includeGlobals = isAdmin || cycle.published
+  const includeGlobals = isAdmin || cycle.published || isAReviewer
   const includeRegions = isAdmin || cycle.published
   const withLock = user && isCountry && Users.hasEditorRole({ user, countryIso, cycle })
 
--- a/src/client/store/ui/dataExport/hooks/index.ts
+++ b/src/client/store/ui/dataExport/hooks/index.ts
@@ -7,6 +7,9 @@ import { useCountries } from 'client/store/area'
 import { DataExportActions, DataExportSelection } from 'client/store/ui/dataExport'
 import { useHomeCountriesFilter } from 'client/store/ui/home'
 import { useCountryIso } from 'client/hooks'
+import { useCycle } from 'client/store/assessment'
+import { useUser, useUserCountries } from 'client/store/user'
+import { Users } from 'meta/user'
 
 export const useDataExportCountries = (): Array<Country> => {
   const dispatch = useAppDispatch()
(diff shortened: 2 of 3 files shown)


src/client/components/PageLayout/Toolbar/Toolbar.tsx
```diff
@@ -45,7 +44,7 @@ const Toolbar: React.FC = () => {
         <AreaSelector
           enableDownload
           includeCountries
-          includeGlobals={includeGlobals}
+          includeGlobals
```
minotogna: includeGlobals must be true if cycle is published or user is admin or user is reviewer

minotogna (changes requested):

src/client/components/PageLayout/Toolbar/Toolbar.tsx
yaguzmang: I had this but removed it. So even if the user is a collaborator (or any other role) in any country, Global should be enabled if the user is a reviewer only?

minotogna: Global must be enabled only if user is a Reviewer or Admin or cycle is published. does that answer?

src/meta/user/users.ts
```diff
@@ -36,6 +36,9 @@ const isCollaborator = (user: User, countryIso: CountryIso, cycle: Cycle) =>
 const isReviewer = (user: User, countryIso: AreaCode, cycle: Cycle) =>
   isRole(user, RoleName.REVIEWER, countryIso, cycle)
 
+const isReviewerInAnyCountry = (user: User, countries: Country[], cycle: Cycle): boolean =>
+  countries.some((country) => isRole(user, RoleName.REVIEWER, country.countryIso, cycle))
```
minotogna: good call. just two comments:
1. Let's call the method isAReviewer(user, cycle)
2. you might not need to pass all countries. only cycle. it's enough to verify that for that cycle there's a reviewer role. would that work ?

src/client/pages/DataExport/CountrySelect/CountrySelect.tsx
```diff
@@ -21,7 +24,15 @@ const CountrySelect: React.FC = () => {
   const countries = useDataExportCountries()
```
minotogna: I see too many filter and to not alter the logic of available countries, (that is given by useDataExportCountries), I would return available countries directly from useDataExportCountries. what do you think ?

src/meta/user/users.ts
yaguzmang: 1. Done. That's better 👍 
2. Yes, that works. I added it now. I also added a video showing how it looks when there is no reviewer role vs. when there is.

src/client/pages/DataExport/CountrySelect/CountrySelect.tsx
yaguzmang: Since that hook is used elsewhere, I'll check those changes won't create unexpected behavior first. But I think yes, that's a better approach to encapsulate that logic.

minotogna: nice. it looks good. let me check the code :)

minotogna: why can't we filter countries directly from useDataExportCountries?

yaguzmang: I was checking whether it would affect something else or not, and seems it won't. I'll add the filter logic in the hook and test now

yaguzmang: Added it. Thanks for the suggestions 👌

src/client/store/ui/dataExport/hooks/index.ts
```diff
@@ -20,10 +23,16 @@ export const useDataExportCountries = (): Array<Country> => {
...
   const countriesAll = useCountries()
   const countriesFilter = useHomeCountriesFilter()
+  const cycle = useCycle()
+  const user = useUser()
+  const userCountries = useUserCountries()
 
   // initialize data export countries
   if (Objects.isEmpty(countries)) {
     let countriesDataExport = countriesAll
+    if (!cycle.published && !Users.isAdministrator(user)) {
```
minotogna, line 33: To be safe , this should be added in the above condition with ?: otherwise, if user filter countries in home dashboard this won't work.

minotogna:
@yaguzmang can we have only one branch please ?

PR #3411, opened 2024-01-15
3408 - Contacts: introduce edit icon for logged in admins

sorja:
Icon only shown for logged in **admins**

https://github.com/openforis/fra-platform/assets/5508251/3c4ab7c8-8301-4cfc-b8bf-cce753c9cee4

Fixed ODP:

https://github.com/openforis/fra-platform/assets/5508251/ea8dfd62-776d-498c-94f4-b2df61c13641

### DIFF
--- a/src/client/components/Buttons/ButtonDelete/ButtonDelete.tsx
+++ b/src/client/components/Buttons/ButtonDelete/ButtonDelete.tsx
@@ -4,14 +4,14 @@ import React from 'react'
 import Icon from 'client/components/Icon'
 
 type Props = {
-  onDelete: () => void
+  onClick: () => void
 }
 
 const ButtonDelete: React.FC<Props> = (props) => {
-  const { onDelete } = props
+  const { onClick } = props
 
   return (
-    <button className="btn-s btn-link-destructive btn-delete" onClick={onDelete} type="button">
+    <button className="btn-s btn-link-destructive btn-delete" onClick={onClick} type="button">
       <Icon className="icon-no-margin" name="trash-simple" />
     </button>
   )
--- a/src/client/components/Buttons/ButtonDelete/ButtonDelete.tsx
+++ b/src/client/components/Buttons/ButtonDelete/ButtonDelete.tsx
@@ -4,14 +4,14 @@ import React from 'react'
 import Icon from 'client/components/Icon'
 
 type Props = {
-  onClick: () => void
+  onDelete: () => void
 }
 
 const ButtonDelete: React.FC<Props> = (props) => {
-  const { onClick } = props
+  const { onDelete } = props
 
   return (
-    <button className="btn-s btn-link-destructive btn-delete" onClick={onClick} type="button">
+    <button className="btn-s btn-link-destructive btn-delete" onClick={onDelete} type="button">
       <Icon className="icon-no-margin" name="trash-simple" />
     </button>
   )
--- a/src/client/components/Buttons/ButtonEdit/ButtonEdit.tsx
+++ b/src/client/components/Buttons/ButtonEdit/ButtonEdit.tsx
@@ -1,3 +1,4 @@
+import './ButtonEdit.scss'
 import React from 'react'
 import { Link } from 'react-router-dom'
 
@@ -11,7 +12,7 @@ const ButtonEdit: React.FC<Props> = (props) => {
   const { url } = props
 
   return (
-    <Link target="_blank" to={url} type="button" className="btn-s btn-link">
+    <Link target="_blank" to={url} type="button" className="btn-s btn-link btn-edit">
       <Icon name="pencil" />
     </Link>
   )
--- a/src/client/components/Buttons/ButtonEdit/ButtonEdit.tsx
+++ b/src/client/components/Buttons/ButtonEdit/ButtonEdit.tsx
@@ -1,0 +1,20 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+
+import Icon from 'client/components/Icon'
+
+type Props = {
+  url: string
+}
+
+const ButtonEdit: React.FC<Props> = (props) => {
+  const { url } = props
+
+  return (
+    <Link target="_blank" to={url} type="button" className="btn-s btn-link">
+      <Icon name="pencil" />
+    </Link>
+  )
+}
+
+export default ButtonEdit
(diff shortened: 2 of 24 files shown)


src/client/components/DataGrid/DataRow/DataRow.tsx
```diff
@@ -1,10 +1,16 @@
...
+import { Routes } from 'meta/routes'
+
+import { useCountryRouteParams } from 'client/hooks/useRouteParams'
 import ButtonDelete from 'client/components/Buttons/ButtonDelete'
 import DataCell from 'client/components/DataGrid/DataCell'
+import Icon from 'client/components/Icon'
 import ReviewIndicator from 'client/components/ReviewIndicator'
 
 export type DataRowActions = {
+  userLink?: any
```
minotogna: Can this just be editLink? And pass generic link from outside ?
This is a generic component

src/client/components/DataGrid/DataRow/DataRow.tsx
```diff
@@ -35,17 +43,30 @@ const DataRow: React.FC<DataRowProps> = (props) => {
...
+                to={Routes.CountryUser.generatePath({
+                  countryIso,
+                  assessmentName,
+                  cycleName,
+                  id: userLink.userId,
+                })}
+                type="button"
+                className="btn-s btn-link"
+              >
+                <Icon name="pencil" />
```
minotogna: Let's add a ButtonEdit similar to ButtonDelete ?

minotogna (commented):
Please , also , restore the vertical line separating the actions cell

sorja:
> Please , also , restore the vertical line separating the actions cell

Sure, made created an Array of `Action` so we can have dynamic number of actions :)

This seemed to work pretty well

Added also type `DataRowActionType`

src/client/components/DataGrid/DataRow/types.ts
```diff
@@ -0,0 +1,23 @@
+export enum DataRowActionType {
+  EditLink = 'editLink',
```
sorja, line 2: should this be just `Edit`?
I was a bit 50% 50% with this

src/client/components/Buttons/ButtonDelete/ButtonDelete.tsx
```diff
@@ -4,14 +4,14 @@ import React from 'react'
...
+  onDelete: () => void
 }
 
 const ButtonDelete: React.FC<Props> = (props) => {
-  const { onClick } = props
+  const { onDelete } = props
 
   return (
-    <button className="btn-s btn-link-destructive btn-delete" onClick={onClick} type="button">
+    <button className="btn-s btn-link-destructive btn-delete" onClick={onDelete} type="button">
```
minotogna: I would still call this on click , because on clicking the button delete then the client will perform the action

src/client/components/DataGrid/DataCell/DataCell.scss
```diff
@@ -14,20 +15,28 @@ $border-editable: 1px dashed $ui-edit-border-color;
...
-  &.review {
+  &.actions {
     align-items: center;
     align-self: center;
     border: unset;
     display: flex;
-    justify-content: center;
+
+    // divider between actions
+    > * {
```
minotogna: So we have 1 column for all actions ?

sorja: Yes!
I think it's better this way, what do you think?

image

minotogna:
Does the ButtonEdit has same hover icon style ? (Like delete and review )?
Always better to post videos when the feature is more complex than just a screenshot

minotogna:
It seems the review icon is too close to delete button .

sorja:
Great, fixed the effect and added videos

Let me know

src/client/components/Buttons/ButtonEdit/ButtonEdit.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type Props = {
+  url: string
+}
+
+const ButtonEdit: React.FC<Props> = (props) => {
+  const { url } = props
+
+  return (
+    <Link target="_blank" to={url} type="button" className="btn-s btn-link btn-edit">
```
minotogna: Maybe no need for target blank ?

src/client/components/DataGrid/DataCell/DataCell.scss
minotogna: Not sure . The layout must not be delegated to buttons .
I'm also not in favor of using * for css classes

src/client/components/DataGrid/DataGrid/DataGrid.tsx
```diff
@@ -12,7 +12,7 @@ const DataGrid: React.FC<Props> = (props) => {
   const { children, className, gridTemplateColumns, withReview } = props
 
   const style = useMemo<CSSProperties>(() => {
-    if (withReview) return { gridTemplateColumns: `${gridTemplateColumns} 32px` }
+    if (withReview) return { gridTemplateColumns: `${gridTemplateColumns} auto` }
```
minotogna: Should this be called with actions ?

src/client/components/DataGrid/DataCell/DataCell.scss
minotogna: But it's a good intuition . Let's leave one column for actions . Let's add a container around each action with border right ?

sorja: Sure, This is what I was thinking also, eg div.actions > div.action

src/client/components/Buttons/ButtonEdit/ButtonEdit.tsx
sorja: What do you prefer?
I think the user wants to edit in another window in this case (sees incorrect data -> edit incorrect data in new window -> check and verify changes)

sorja:
I also agree not using * for css classes, I tried to find a css only solution, but I think as you suggested we must wrap them with a div.action and replace * with action in this case

minotogna:
> I also agree not using * for css classes, I tried to find a css only solution, but I think as you suggested we must wrap them with a div.action and replace * with action in this case

Yes, sounds good @sorja 👍

src/client/components/Buttons/ButtonEdit/ButtonEdit.tsx
minotogna: The problem is that when editing user in a different window, data will remain inconsistent in the original contacts page .
Instead, if he edits user in the same window, to go back to contacts he will press browser back button that will then refresh the contacts

minotogna (approved):

PR #3417, opened 2024-01-17
3408 Contacts: Allow editing User Role Properties only when Country Status EDITING

sorja:
https://github.com/openforis/fra-platform/assets/5508251/0731fa84-b46a-4252-9254-b3d9de409fdc

### DIFF
--- a/src/client/components/EditUserForm/UserRolePropsFields/hooks/useIsEnabled.ts
+++ b/src/client/components/EditUserForm/UserRolePropsFields/hooks/useIsEnabled.ts
@@ -1,0 +1,20 @@
+import { useMemo } from 'react'
+
+import { Areas, AssessmentStatus, CountryIso } from 'meta/area'
+import { Users } from 'meta/user'
+
+import { useCountry } from 'client/store/area'
+import { useUser } from 'client/store/user'
+import { useCountryRouteParams } from 'client/hooks/useRouteParams'
+
+export const useIsEnabled = (enabled: boolean): boolean => {
+  const { countryIso } = useCountryRouteParams<CountryIso>()
+
+  const country = useCountry(countryIso)
+  const user = useUser()
+  const isAdmin = Users.isAdministrator(user)
+
+  return useMemo<boolean>(() => {
+    return enabled && (isAdmin || Areas.getStatus(country) === AssessmentStatus.editing)
+  }, [enabled, isAdmin, country])
+}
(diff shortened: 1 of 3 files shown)


src/client/components/EditUserForm/UserRolePropsFields/hooks/useIsEnabled.ts
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from 'react'
+
+import { Areas, AssessmentStatus, CountryIso } from 'meta/area'
+import { Users } from 'meta/user'
+
+import { useCountry } from 'client/store/area'
+import { useUser } from 'client/store/user'
+import { useCountryRouteParams } from 'client/hooks/useRouteParams'
+
+export const useIsEnabled = (enabled: boolean): boolean => {
```
minotogna: What about canEditRoleProps?

sorja: Are you referring to
`const canEditRoles = !isReviewer && !isSelf && isAdministrator && Areas.isISOGlobal(countryIso)`

This is used for the Admin version of the role selector (where you can pick countries from the modal)

minotogna: I meant renaming this hook? I was thinking to create a common place where
To export auth hooks, e.g.

Import {canEdtData, canEditDescriptions,...} from 'client/hooks/useAuth'

...

Then we have good names now, it can be easier to identify them later on ..

What do you think ?

sorja: Sounds good, but I wonder if it should be under `Users`? and pass 
```
{
   cycle: Cycle
   country: Country
   user: User
}
```
?

minotogna: The hooks per not really .
But does this hook represents can edit role pros ? So if yes I would rename this hook to useCanEditRole ? useCanEditRoleProps? 
So it's easier to identify later on .
What do you think ?

sorja: Sure, I will rename the hook
Should I aforementioned path, `'client/hooks/useAuth'` or keep as is and refactor later?

minotogna: You can add that hook now , it's not a big deal and we will start the refactor later .. I'll keep a note

minotogna (approved):

PR #3425, opened 2024-01-19
3408 Contacts: Layout bug

sorja:


sorja:
Fixes a layout bug which occurs if there are no actions in edit mode (eg. no rights to edit)

image

minotogna (approved):

PR #3429, opened 2024-01-23
3222 - Zeros are partially removed from the previous bug 

### LINKED ISSUE #3222: Zeros are partially removed from the previous bug
@minotogna @sorja 

**Describe the bug**
Previous bug of tab 1b has been fixed(https://github.com/openforis/fra-platform/issues/3198#issuecomment-1786303909) and most zeros are removed, but not the sub-classes of **'Primary forest by climatic domain,'** in which the cells should also be empty. I can no longer manually remove the zeros as the report has been accepted.

**To Reproduce**
1. Go to 'Production- Bangladesh'
2. Click on 'Table 1b - Primary forest by climatic domain'

**Expected behavior**
Auto-filled zeros due to the previous bug(https://github.com/openforis/fra-platform/issues/3198#issuecomment-1786303909) should be removed as what have been done for the total.

**Screenshots**
Before the bug has been fixed(https://github.com/openforis/fra-platform/issues/3198#issuecomment-1786303909) : 
image

Current situation:
image

sorja:
https://github.com/openforis/fra-platform/assets/5508251/4f5969f4-abad-442a-9823-836943c3f155


minotogna:
@sorja default value will be removed if I understand correctly, is that what's been asked ?

sorja:
> @sorja default value will be removed if I understand correctly, is that what's been asked ?

This what I understood from last comment, thoughts?

My thoughts were that data process is soon done and we can revisit this feature for the future better
We can of course now implement whatever is required, just let me know what you prefer

src/test/migrations/steps/20240123101318-step-1b-subclasses-primary-forest-remove-default-value.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  await client.query(
+    `
+        update ${schemaName}.row
+        set props = props - 'calculateFn' - 'calculateIf'
+        where id in (select r.id
+                     from ${schemaName}.table t
+                              left join ${schemaName}.row r on t.id = r.table_id
+                     where t.props ->> 'name' = $1
+          and r.props ->> 'variableName' in ($2:list))
```
minotogna, line 26: Should Those cells be recalculated in order to update the cache or the node values should remain as they are ?

sorja: Node values should remain as they are as there is no change in the values, only in the logic

minotogna:
> > @sorja default value will be removed if I understand correctly, is that what's been asked ?
> 
> 
> 
> This what I understood from last comment, thoughts?
> 
> 
> 
> My thoughts were that data process is soon done and we can revisit this feature for the future better
> 
> We can of course now implement whatever is required, just let me know what you prefer

I understood from slack  @pengchenglai were about to ask to OJ . Can you please double check with him ?

minotogna (approved):

PR #3443, opened 2024-01-26
3214 - Files and links per Cycle: getMany files and links endpoint

sorja:


src/meta/api/endpoint/ApiEndPoint.ts
```diff
@@ -62,7 +62,10 @@ export const ApiEndPoint = {
     },
 
     Repository: {
-      one: () => apiPath('cycle-data', 'repository'),
+      one: () => apiPath('cycle-data', 'repository', 'repository'),
+      file: (uuid = ':uuid') => apiPath('cycle-data', 'repository', 'file', uuid),
+      files: () => apiPath('cycle-data', 'repository', 'files'),
+      links: () => apiPath('cycle-data', 'repository', 'links'),
     },
```
minotogna, line 68: Would it make sense only one endpoint with parameter targetCounyryIso optional ?

sorja: Not sure if we need targetCounyryIso? What would be the use case?

I think in this case but we could have repository.many with a parameter 'links', 'files' or undefined?

src/server/repository/assessmentCycle/repository/getMany.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  let condition = ''
+
+  if (files) {
+    condition = 'and file_uuid is not null'
+  }
+
+  if (links) {
+    condition = 'and link is not null'
+  }
```
minotogna: why this condition ? in UI there's a difference between global and country

src/meta/api/endpoint/ApiEndPoint.ts
minotogna: the code, if you look closely, is the same 
image

would it make sense to pass the global parameter as query param ?

src/meta/api/endpoint/ApiEndPoint.ts
```diff
@@ -62,7 +62,10 @@ export const ApiEndPoint = {
...
 
     Repository: {
-      one: () => apiPath('cycle-data', 'repository'),
+      one: () => apiPath('cycle-data', 'repository', 'repository'),
+      file: (uuid = ':uuid') => apiPath('cycle-data', 'repository', 'file', uuid),
```
minotogna, line 65: would it make sense to remain with our convention where params are passed as query param ?

src/server/api/cycleData/index.ts
```diff
@@ -150,5 +152,7 @@ export const CycleDataApi = {
       AuthMiddleware.requireEditAssessmentFile,
       createFile
     )
+    express.get(ApiEndPoint.CycleData.Repository.files(), AuthMiddleware.requireView, getManyRepositoryFiles)
+    express.get(ApiEndPoint.CycleData.Repository.links(), AuthMiddleware.requireView, getManyRepositoryLinks)
```
minotogna: didn't we have a custom middleware for authorised files or not?

src/meta/api/endpoint/ApiEndPoint.ts
sorja: Should we instead pass 'link' or 'files' as parameter?

src/meta/api/endpoint/ApiEndPoint.ts
sorja: We use it in filePath to avoid passing query params in links, should we instead add query params?
We used this convention in previous implementations, eg. 
`      one: (uuid = ':uuid') => apiPath('file', 'assessment', uuid),`

minotogna: ok that's fine

src/meta/api/endpoint/ApiEndPoint.ts
minotogna: I don't really like the ui convention terminology. those are just labels and I think and hope will change later

minotogna: we should express things for what they are, not how we label them in the UI. 
if you have a better idea for indicating that we are fetching global vs country..

sorja: Ok, let's clarify a bit the functionality here, so we have one point of reference:

in UI, we have links and repository which correspond to links and files, let's use for this discussion 'links' and 'files' to be clear.


## Links
Under the UI label Links for Country, we have:
- All global links
- All country links
to fetch those, you need to have access: view access.*

We should fetch them using a single endpoint, eg. 
`many() => apiPath('cycle-data', 'repository'),` 
- with query params ? (eg links, or filter in UI?)

## Files
Under the UI label Repository for Country, we have:
- All global files (not hidden files)
- All country files
to fetch those, you need to have access: view access.*

We should fetch them using a single endpoint, eg. 
`many() => apiPath('cycle-data', 'repository'),` 
- with query params ? (eg files, or filter in UI?)
- I am not sure I understood the global files/links, in global I can't see the Links & Repository, maybe you are referring to something else:
image




*I will recheck this, we might have separate authorizer for this, as you mentioned

minotogna: in UI Links and Repository do not mean Links and Files.

Under links we have global repository items
Under repository we have country repository items.

and by repository items we mean files and links stored in the repository table,

does it make sense now ?

sorja: yes! i will implement as this!

minotogna: @sorja the current implementation is this way. check that as well

src/server/api/cycleData/index.ts
sorja:
```
    express.get(ApiEndPoint.File.Assessment.many(), AuthMiddleware.requireView, getAssessmentFiles)
    express.get(ApiEndPoint.File.Assessment.one(), /* Auth handled in controller */ getAssessmentFile)

```

This is the old one

src/meta/api/endpoint/ApiEndPoint.ts
```diff
@@ -62,7 +62,9 @@ export const ApiEndPoint = {
     },
 
     Repository: {
-      one: () => apiPath('cycle-data', 'repository'),
+      file: (uuid = ':uuid') => apiPath('cycle-data', 'repository', 'file', uuid),
+      many: () => apiPath('cycle-data', 'repository'),
+      one: () => apiPath('cycle-data', 'repository', 'repository'),
```
minotogna: we should maybe use /cycle-data/repositories and /cycle-data/repositories/repository to be consistent . what do you think ?

src/server/api/cycleData/index.ts
```diff
@@ -150,5 +151,6 @@ export const CycleDataApi = {
       AuthMiddleware.requireEditAssessmentFile,
       createFile
     )
+    express.get(ApiEndPoint.CycleData.Repository.many(), AuthMiddleware.requireView, getManyRepository)
```
minotogna, line 154: isn't this requireEdit, can u check previous implementation ? in public view those resources are not accessible

src/server/api/cycleData/repository/getManyRepository.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const getManyRepository = async (req: Request, res: Response) => {
+  try {
+    const { assessmentName, cycleName, countryIso, global } = req.query
+
+    const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName })
+
+    const props = { assessment, cycle, countryIso, global }
+    const createdRepositoryEntity = await CycleDataController.Repository.getMany(props)
```
minotogna: no created. those are read , but I would call this just items or something like that

src/meta/api/endpoint/ApiEndPoint.ts
sorja: many: /cycle-data/repository/items
one: /cycle-data/repository/items/item

sorja:
Rename type Repository to RepositoryItem

src/server/api/cycleData/index.ts
sorja: as discussed, requireView

src/meta/cycleData/repository/repository.ts
```diff
@@ -5,7 +5,7 @@ export interface RepositoryProps {
   public?: boolean // File visible for non-logged in users
 }
 
-export type Repository = {
+export type RepositoryItem = {
```
minotogna, line 8: check the name of RepositoryProps , should it be RepositoryItemProps

sorja: Good eye... for some reason it rolled this change back, fixed.

minotogna (approved):

PR #3446, opened 2024-01-26
3431 - Init inivitation, redirects and search params hook

yaguzmang:
Redirects for accepted invitation:

https://github.com/openforis/fra-platform/assets/41337901/8c7d5062-9ac1-46d1-9a02-5ee14457ccbd

Collaborator button:
Before:

(Admin user management)
image

image
After:

image

image

### DIFF
--- a/src/client/components/UserList/UserListHeader/UserListHeader.tsx
+++ b/src/client/components/UserList/UserListHeader/UserListHeader.tsx
@@ -37,8 +37,10 @@ const UserListHeader: React.FC<Props> = (props) => {
           ))}
         {!readOnly && (
           <th className="user-list__header-cell user-list__edit-column">
-            <UserListButtonExport isAdmin={isAdminRoute} />
-            <InviteUserLink />
+            <div className="user-list__edit-column-buttons-container">
+              <UserListButtonExport isAdmin={isAdminRoute} />
+              {!isAdminRoute && <InviteUserLink />}
+            </div>
           </th>
         )}
       </tr>
--- a/src/client/hooks/useSearchParams.ts
+++ b/src/client/hooks/useSearchParams.ts
@@ -1,0 +1,13 @@
+import { useMemo } from 'react'
+import { useLocation } from 'react-router-dom'
+
+type BaseQueryParams = Record<string, string | undefined>
+
+export const useSearchParams = <SearchParams extends BaseQueryParams>(): Partial<SearchParams> => {
+  const { search } = useLocation()
+
+  return useMemo(() => {
+    const params = new URLSearchParams(search)
+    return Object.fromEntries(params) as Partial<SearchParams>
+  }, [search])
+}
--- a/src/client/pages/Login/Invitation/Invitation.tsx
+++ b/src/client/pages/Login/Invitation/Invitation.tsx
@@ -16,11 +16,15 @@ import Icon from 'client/components/Icon'
 import { isError, LoginValidator } from 'client/pages/Login/utils/LoginValidator'
 import { videoResources } from 'client/pages/Tutorials'
 
+import { useInitInvitation } from './hooks/useInitInvitation'
+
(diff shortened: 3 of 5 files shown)


src/client/components/UserList/UserListHeader/UserListHeader.tsx
```diff
@@ -37,8 +37,10 @@ const UserListHeader: React.FC<Props> = (props) => {
           ))}
         {!readOnly && (
           <th className="user-list__header-cell user-list__edit-column">
-            <UserListButtonExport isAdmin={isAdminRoute} />
-            <InviteUserLink />
+            <div className="user-list__edit-column-buttons-container">
```
minotogna, line 40: @yaguzmang the manage collaborators page , in country home, does still work properly ?

src/client/hooks/useSearchParams.ts
```diff
@@ -0,0 +1,13 @@
+import { useMemo } from 'react'
+import { useLocation } from 'react-router-dom'
+
+type BaseQueryParams = Record<string, string | undefined>
+
+export const useSearchParams = <SearchParams extends BaseQueryParams>(): Partial<SearchParams> => {
```
minotogna, line 6: 👍

src/client/pages/Login/Invitation/Invitation.tsx
```diff
@@ -73,13 +69,15 @@ const Invitation: React.FC = () => {
...
   }
 
-  if (userRole?.acceptedAt) {
-    return (
-      <div className="login__form">
-        <h3>{t('login.alreadyAcceptedInvitation')}</h3>
-      </div>
-    )
-  }
+  useEffect(() => {
```
minotogna: would it make sense to use a return <Redirect , without useEffect ? the useEffect has effect on the next render, return the current render . just wondering

src/client/pages/Login/Invitation/hooks/useInitInvitation.ts
```diff
@@ -0,0 +1,23 @@
...
+  const { invitationUuid } = useSearchParams<LoginInvitationQueryParams>()
+
+  const dispatch = useAppDispatch()
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    if (invitationUuid) {
+      dispatch(LoginActions.fetchUserByInvitation({ invitationUuid }))
+    } else {
+      navigate('/')
```
minotogna: can we Routes.Root.generatePath() instead of / ?

src/client/components/UserList/UserListHeader/UserListHeader.tsx
yaguzmang: Yes:

https://github.com/openforis/fra-platform/assets/41337901/7c74296a-f45b-4e53-9f00-6ecc1ec9e0a8

src/client/pages/Login/Invitation/Invitation.tsx
yaguzmang: Yes, that makes sense. I will use <Navigate because apparently <Redirect is not available in our react-router-dom version.

minotogna: right, Navigate then ! ;)

minotogna: use the replace=true

src/client/pages/Login/Invitation/hooks/useInitInvitation.ts
yaguzmang: Yes. Missed it before, thanks!

src/client/pages/Login/Invitation/Invitation.tsx
yaguzmang: Done 👍

minotogna (approved):

PR #3447, opened 2024-01-29
3431 - Refactor invitation component

yaguzmang:
Refactoring the login Invitation component by:

- Adding a new component InvitationLocal that handles the form input, following the same logic as before. This component is associated to the route `/login/invitation/local?` (LoginInvitationLocal)
- Adding AcceptInvitationButtons and AccessLimited components.
- Adding InvitationState type.
- Adding redux state for the Invitation Local form.

Scenarios (checking that it does the same as before):

- Accepting invitation while no user is logged-in (and form validation):

https://github.com/openforis/fra-platform/assets/41337901/09fb11e7-cb71-4f75-aabe-92b6d9a857b5

- Trying to accept an already accepted invitation while a user is logged-in:

https://github.com/openforis/fra-platform/assets/41337901/13696e68-5956-4dfd-aae9-e6d891d86e29

- Trying to accept an already accepted invitation while no user is logged-in:

https://github.com/openforis/fra-platform/assets/41337901/06abee33-dba6-46fa-8578-bacb568ea19c

- Accepting an invitation when the user already has an account:
 
https://github.com/openforis/fra-platform/assets/41337901/848bef3b-2a5a-4cd3-91be-522851f0a9d6

- Invitation has expired (while logged-in and not logged-in)

https://github.com/openforis/fra-platform/assets/41337901/59a9e52e-ad2b-4d07-b938-6168ad02732a

PS: This PR turned out a bit bigger that I would have liked it to be, but it's mostly moving existing logic to new components. Also, in order to be able to test, dividing it further wouldn't be possible.


src/client/pages/PageRoutes/hooks/useRoutes.tsx
```diff
@@ -67,7 +67,10 @@ export const useRoutes = () => {
             {/* Login */}
             <Route path={Routes.Login.path.relative} element={<Login />}>
               <Route index element={<LoginForm />} />
-              <Route path={Routes.LoginInvitation.path.relative} element={<LoginInvitation />} />
+              <Route path={Routes.LoginInvitation.path.relative}>
+                <Route index element={<LoginInvitation />} />
+                <Route path={Routes.LoginInvitationLocal.path.relative} element={<LoginInvitationLocal />} />
```
minotogna, line 71: LoginInvitationLocal should be inner route of LoginInvitation. that way you can use Outlet to render the local form and AcceptInvitationButtons can remain as it was , no need to call it twice

yaguzmang: I changed the routes to this and now I can use Outlet:
```
<Route path={Routes.Login.path.relative} element={<Login />}>
    <Route index element={<LoginForm />} />
    <Route path={Routes.LoginInvitation.path.relative} element={<LoginInvitation />}>
      <Route path={Routes.LoginInvitationLocal.path.relative} element={<LoginInvitationLocal />} />
    </Route>
    <Route path={Routes.LoginResetPassword.path.relative} element={<LoginResetPassword />} />
</Route>
```
However, to be able to keep the Buttons in LoginInvitation, I need to get this function from LoginInvitationLocal:
`const onAcceptInvitation = useOnAcceptInvitationLocal({ formData, invitationUuid, setErrors, showPassword2 })` in LoginInvitation, that way if I am not in Local, the button will take me there and otherwise it will trigger accepting the invitation.

To get that function I tried passing a ref through the Outlet so that the the LoginInvitationLocal can update it:
```
export type InvitationContext = {
  onAcceptInvitationLocalRef: React.MutableRefObject<() => void>
}
const Invitation: React.FC = () => {
...
  const onAcceptInvitationLocalRef = useRef<() => void>()
  const context: InvitationContext = { onAcceptInvitationLocalRef }
 ...
 return(
...
<Outlet context={context} />
...
)
```
This did not work as I couldn't trigger a re render to the parent Invitation from InvitationLocal after updating the function:
```
const InvitationLocal: React.FC = () => {
  ...
  const { onAcceptInvitationLocalRef } = useOutletContext<InvitationContext>()
  onAcceptInvitationLocalRef.current  = useOnAcceptInvitationLocal({ formData, invitationUuid, setErrors, showPassword2 })
...
```

I would appreciate any suggestions on how to handle this. 
The other option I see is creating a slice with the form state and dispatching actions, but I think that would need a new PR.

minotogna: > I changed the routes to this and now I can use Outlet:
> 
> ```
> <Route path={Routes.Login.path.relative} element={<Login />}>
>     <Route index element={<LoginForm />} />
>     <Route path={Routes.LoginInvitation.path.relative} element={<LoginInvitation />}>
>       <Route path={Routes.LoginInvitationLocal.path.relative} element={<LoginInvitationLocal />} />
>     </Route>
>     <Route path={Routes.LoginResetPassword.path.relative} element={<LoginResetPassword />} />
> </Route>
> ```
> 
> However, to be able to keep the Buttons in LoginInvitation, I need to get this function from LoginInvitationLocal: `const onAcceptInvitation = useOnAcceptInvitationLocal({ formData, invitationUuid, setErrors, showPassword2 })` in LoginInvitation, that way if I am not in Local, the button will take me there and otherwise it will trigger accepting the invitation.
> 
> To get that function I tried passing a ref through the Outlet so that the the LoginInvitationLocal can update it:
> 
> ```
> export type InvitationContext = {
>   onAcceptInvitationLocalRef: React.MutableRefObject<() => void>
> }
> const Invitation: React.FC = () => {
> ...
>   const onAcceptInvitationLocalRef = useRef<() => void>()
>   const context: InvitationContext = { onAcceptInvitationLocalRef }
>  ...
>  return(
> ...
> <Outlet context={context} />
> ...
> )
> ```
> 
> This did not work as I couldn't trigger a re render to the parent Invitation from InvitationLocal after updating the function:
> 
> ```
> const InvitationLocal: React.FC = () => {
>   ...
>   const { onAcceptInvitationLocalRef } = useOutletContext<InvitationContext>()
>   onAcceptInvitationLocalRef.current  = useOnAcceptInvitationLocal({ formData, invitationUuid, setErrors, showPassword2 })
> ...
> ```
> 
> I would appreciate any suggestions on how to handle this. The other option I see is creating a slice with the form state and dispatching actions, but I think that would need a new PR.

can the button trigger a conditional action based on location?
i.e. if in LoginInvitation -> navigate to LoginInvitationLocal
if in LoginInvitationLocal -> submit form.
would that be enough ?

yaguzmang: > if in LoginInvitation -> navigate to LoginInvitationLocal
> if in LoginInvitationLocal -> submit form.

To submit the form from the parent route LoginInvitation, I would need to access the `formData` that's in the child route LoginInvitationLocal. To do that while using Outlet, I would need to use context API and wrap both routes. Or, I could lift the form state up, but that would leave us with the same code as before 😅 

If context API is okay for you, I can implement it.

yaguzmang: Added Outlet to the most recent commit, but with conditional button render so we can take it from there 👍

minotogna: What if we put the form data model in redux ?

yaguzmang: Sure. Should I add it to the existing `store/login` or create a new slice in `store/ui/...`? since it is just form data

minotogna: same slice

yaguzmang: Added. Now the button is only present in Invitation and InvitationLocal only handles the form.

minotogna: it looks much cleaner now.

src/client/pages/Login/Invitation/Invitation.tsx
```diff
@@ -86,123 +67,36 @@ const Invitation: React.FC = () => {
...
-          <a
-            className="btn-help"
-            href={videoResources[1].url[i18n.resolvedLanguage] ?? videoResources[1].url.en}
-            rel="noreferrer"
-            target="_blank"
-          >
-            <Icon name="video" className="icon-sub" /> {t(videoResources[1].labelKeyShort)}
-          </a>
+          <AcceptInvitationButtons
+            onAcceptInvitationLocalClick={isInInvitationLocal ? onAcceptInvitationLocal : goToAcceptInvitationLocal}
```
minotogna: would it be possible to avoid passing this prop? if it's used only be AcceptInvitationButtons, then better to declare the logic inside the buttons themselves. what do you think ?

yaguzmang: Yes, good point. Now I moved the logic to handle the redirect or submit to AcceptInvitationButtons, so no prop needed 👍

minotogna (approved):

PR #3458, opened 2024-01-31
Prettier config update

sorja:


sorja:
There's something weird going on with my end of line characters, I randomly have CRLF and LF, hope this will fix it.
Also added spacing for brackets.

`{ x }` vs `{x}`


@minotogna 

Can you be so kind to verify that you are using also LF EOL?

.prettierrc
```diff
@@ -2,5 +2,7 @@
   "printWidth": 120,
   "singleQuote": true,
   "semi": false,
-  "trailingComma": "es5"
+  "trailingComma": "es5",
+  "bracketSpacing": true,
+  "endOfLine": "lf"
```
minotogna, line 7: Can u show the differences on how it will look please ?

sorja: you mean lf? it is invisible character denoting the end of the line, for windows systems its CLRF and for linux / other it's LF i think, it should say in the bottom bar of intellij what character is chosen (LF, CLRF)

for bracketspacing, i added above an example

.prettierrc
```diff
@@ -2,5 +2,7 @@
   "printWidth": 120,
   "singleQuote": true,
   "semi": false,
-  "trailingComma": "es5"
+  "trailingComma": "es5",
+  "bracketSpacing": true,
```
minotogna, line 6: shouldn't we add always new line in case of `{` ?
Or what do you think ?

sorja: What do you mean?
We use this way of writing objects a lot:
```
const params = { assessmentName, cycleName, countryIso, orderBy, page, path, limit }
```

vs

```
const params = {assessmentName, cycleName, countryIso, orderBy, page, path, limit}
```

minotogna (approved):

PR #3461, opened 2024-01-31
3333 - User fix and translate title

### LINKED ISSUE #3333: User: Fix and translate Title
Image

- [x] Persist in users table key, not value
- [x] Update current database values (with migration-steps)
- [x] Replace select with new Select component 
- [x] In select display value 
- [x] Update Users.getContacts query src/server/repository/public/user/getContacts.ts

yaguzmang:
DB after running migration:
image

UI:
- Translations working for all languages, and changing title while in any given language works:

https://github.com/openforis/fra-platform/assets/41337901/47462530-e01c-4ae8-8af5-253ad8c83f09

- User country role selector updated, since it also uses the `SelectField` from EditUserForm:
image

Closes #3333


src/client/components/EditUserForm/EditUserForm.tsx
```diff
@@ -12,6 +12,7 @@ import { useUser } from 'client/store/user'
 import { useCountryIso, useOnUpdate } from 'client/hooks'
 import DisableUser from 'client/components/EditUserForm/DisableUser'
 import UserCountryRoleSelector from 'client/components/EditUserForm/UserCountryRoleSelector'
+import { useOptionsAppellation } from 'client/pages/Section/Contacts/hooks/useOptionsAppellation'
```
yaguzmang: Should this hook be a more generic hook?

sorja: Is this used somewhere else in same format?

yaguzmang: Yes, it's used in `useColumns` for Contacts, in the same format { label, value }.

yaguzmang: If we make it more generic, where would be a good place to put it? `src/client/hooks` doesn't seem like is the right place, or is it?

minotogna: let's put it there for now.

yaguzmang: Moved it 👍

src/client/components/EditUserForm/SelectField/SelectField.tsx
```diff
@@ -7,19 +7,20 @@ import classNames from 'classnames'
 import { User } from 'meta/user'
 import { UserProps } from 'meta/user/user'
 
+import Select, { Option } from 'client/components/Inputs/Select'
+
 type Props = {
   name: string
   value: string
-  options: Record<string, string>
+  options: Array<Option>
```
minotogna, line 15: Have you checked if other edit user form fields use this ?

src/client/components/EditUserForm/UserCountryRoleSelector/hooks/useOptions.tsx
```diff
@@ -3,17 +3,21 @@ import { useTranslation } from 'react-i18next'
...
-        }
-        return acc
-      }, {}),
-    [t]
-  )
+
+  return useMemo<Returned>(() => {
+    const options = Object.keys(RoleName)
+      .filter((key) => key !== RoleName.ADMINISTRATOR)
+      .map((key) => ({
```
minotogna: Filter + map => better one reduce

src/test/migrations/steps/20240130202403-step-fix-user-tittle.ts
```diff
@@ -0,0 +1,16 @@
...
+export default async (client: BaseProtocol) => {
+  await client.query(`
+    UPDATE public.users
+    SET props = 
+      CASE 
+        WHEN props->>'title' = 'Ms' THEN jsonb_set(props, '{title}', '"ms"')
+        WHEN props->>'title' = 'Mr' THEN jsonb_set(props, '{title}', '"mr"')
+        WHEN props->>'title' = 'Mrs' THEN jsonb_set(props, '{title}', '"mrs"')
+        WHEN props->>'title' = 'Other' THEN jsonb_set(props, '{title}', '"other"')
+        ELSE props
```
minotogna, line 12: This could be simplified with just lower(props->>'title').
Just to mention , it is fine if you leave it as it is

minotogna (commented):
Does the contacts form use the new useOptionsAppellation hook ?

src/client/components/EditUserForm/SelectField/SelectField.tsx
yaguzmang: Yes, `UserCountryRoleSelector` uses it, so I also updated it. 

image

yaguzmang: And also yes, the hook was originally added for the Contacts form:
image

minotogna (approved):

PR #3462, opened 2024-02-01
Hide role selection in profile for admin users

sorja:


src/client/components/EditUserForm/EditUserForm.tsx
```diff
@@ -77,7 +77,8 @@ const EditUserForm: React.FC<Props> = ({ user, canEditPermissions, canEditRoles,
 
   const enabled = canEditUser
   const isAdmin = Users.isAdministrator(userInfo)
-  const showRoleSelector = !Areas.isGlobal(countryIso) && isAdmin
+  const isCurrentUserAdmin = Users.isAdministrator(user)
+  const showRoleSelector = !Areas.isGlobal(countryIso) && isAdmin && !isCurrentUserAdmin
```
minotogna: If current user is admin and target user is another role , does the select appear ?

sorja: Yes:

image

minotogna: But in the context above , who is current user and who is is admin ? To me they refer to logged user , that's why I'm confused

sorja: so you mean let's rename the variables to currentLoggedInUser and currentEditUser or similar?

yaguzmang: How about:
`const userInfo = useUser()` becomes `const editorUserInfo = useUser()`
and for the user to edit, there already is a state `userToEdit`.

Then:
`const isUserToEditAdmin = Users.isAdministrator(userToedit)`

So:
 ```
  const isEditorAdmin = Users.isAdministrator(editorUserInfo)
  const isUserToEditAdmin = Users.isAdministrator(userToEdit)
  const showRoleSelector = !Areas.isGlobal(countryIso) && isEditorAdmin && !isUserToEditAdmin
  ```

minotogna: > so you mean let's rename the variables to currentLoggedInUser and currentEditUser or similar?

That would be more clear . Or how about user and editUser / targetUser 
(Shorter versions)
What do you think?

sorja: Sure, I will change to targetUser
I think `user` has always referred to our currently logged in user, so we can keep it as is

`targetUser` we can agree that will refer to editable or secondary user

sorja:
@yaguzmang 
@minotogna 

Updated variables to

`user`
`targetUser`

let me know

minotogna (approved):

PR #3470, opened 2024-02-05
3214 - Files and links per Cycle: Download a file

sorja:


src/server/api/cycleData/index.ts
```diff
@@ -152,6 +153,7 @@ export const CycleDataApi = {
       AuthMiddleware.requireEditAssessmentFile,
       createFile
     )
+    express.get(ApiEndPoint.CycleData.Repository.file(), AuthMiddleware.requireViewFile, getRepositoryFile)
```
sorja, line 156: This is where we diverge a bit from previous implementation, we add the authorizer here because it's very fast to query the `cycleSchema.repository` table, so we can check the rights in the authorizer

The authorizer itself is 1:1, using the new RepositoryItem

The old countryFile authorizer has been deprecated and will be removed in later PRs

src/server/controller/cycleData/repository/getFile.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type Returned = {
+  file: Buffer
+  repositoryItem: RepositoryItem
+}
+
+export const getFile = async (props: Props): Promise<Returned> => {
+  const { assessment, cycle, uuid } = props
+
+  return DB.tx(async (t: BaseProtocol) => {
```
minotogna: Since only reads are involved , no need for transaction

src/server/controller/cycleData/repository/index.ts
```diff
@@ -1,10 +1,13 @@
 import { RepositoryRepository } from 'server/repository/assessmentCycle/repository'
 
 import { create } from './create'
+import { getFile } from './getFile'
 import { remove } from './remove'
 
 export const Repository = {
   create,
   getMany: RepositoryRepository.getMany,
+  getOne: RepositoryRepository.get,
```
minotogna: Better to call repo method getOne as well

src/server/repository/public/file/get.ts
```diff
@@ -0,0 +1,13 @@
+import type { File } from 'meta/file'
+
+import { BaseProtocol, DB } from 'server/db'
+
+type Props = {
+  fileUuid: string
+}
+
+export const get = async (props: Props, client: BaseProtocol = DB): Promise<File> => {
```
minotogna: I would call this getOne as well

src/server/repository/public/file/getOne.ts
```diff
@@ -0,0 +1,13 @@
...
+import { BaseProtocol, DB } from 'server/db'
+
+type Props = {
+  fileUuid: string
+}
+
+export const getOne = async (props: Props, client: BaseProtocol = DB): Promise<File> => {
+  const { fileUuid } = props
+
+  return client.one(`select * from public.file where uuid = $1`, [fileUuid])
```
minotogna: Without camelize , the return type doesn't match File properties I guess

minotogna (approved):

PR #3472, opened 2024-02-05
3214 - Files and links per Cycle: Refetch items on edit

sorja:


src/client/store/ui/repository/actions/upsertRepositoryItem.ts
```diff
@@ -26,5 +27,10 @@ export const upsertRepositoryItem = createAsyncThunk<void, Props, ThunkApiConfig
     const params = { countryIso, assessmentName, cycleName }
     const config = { params }
     await axios.post(ApiEndPoint.CycleData.Repository.one(), formData, config)
+    const path = ApiEndPoint.CycleData.Repository.many()
+    const limit: number = undefined
+    const page: number = undefined
+    const getDataProps = { assessmentName, cycleName, countryIso, path, limit, page }
+    dispatch(TablePaginatedActions.getData(getDataProps))
```
minotogna: I'm not a big fun of duplicated code. maybe create a common function called refetchRepositoryItems ?
Plus, I don't see how it fetches global vs country items..

sorja: Sure sounds good, I am not sure if global items change so often?
I can add it anyway

minotogna: well, if we allow the create/update/remove, also global repository items must be refetched otherwise the list won't reflect the changes

src/client/store/ui/repository/actions/fetchRepositoryItems.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export const fetchRepositoryItems = (props: Props) => {
+  const { dispatch } = props
+
+  const paths = ['', '?global=true'].map((path) => `${ApiEndPoint.CycleData.Repository.many() + path}`)
+  const limit: number = undefined
+  const page: number = undefined
+
+  paths.forEach((path) => {
```
minotogna: Why dispatching both ? Can't we just update the correspondent list ? I.e. global or country ?

src/client/store/types.ts
```diff
@@ -1,5 +1,7 @@
 import { RootState } from 'client/store/RootState'
+import { AppDispatch } from 'client/store/store'
 
 export type ThunkApiConfig = {
   state: RootState
+  dispatch: AppDispatch
```
minotogna, line 6: This is not really part of thunkconfig . Why do you need it here ?

sorja: It's the dispatch we use? Or what do you mean?

sorja: If we don't set it here,
then the dispatch in ThunkApiConfig is incorrectly typed:

image

Same as with state

Maybe I am mistaken?

image

src/client/store/ui/repository/listeners/useRepositoryItemChangeListener.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { ApiEndPoint } from 'meta/api/endpoint'
+
+import { useAppDispatch } from 'client/store'
+import { addAppListener } from 'client/store/middleware/listener'
+import { RepositoryActions } from 'client/store/ui/repository'
+import { TablePaginatedActions } from 'client/store/ui/tablePaginated'
+import { useCountryRouteParams } from 'client/hooks/useRouteParams'
+
+export const useRepositoryItemChangeListener = () => {
```
minotogna: this a a good idea 👍 
However, I would still move this under hooks src/client/store/ui/repository/hooks 
and import it from 'client/store/ui/repository'
what do you think ?

sorja: sure

src/client/store/types.ts
minotogna: @sorja if you want me to read the comments, please don't resolve them ;) I can resolve them once I read your answer :D

minotogna (approved):

PR #3478, opened 2024-02-07
added rough screenshot tool

sorja:
example output

screenshot_FIN


package.json
```diff
@@ -229,6 +229,7 @@
     "passport-local": "^1.0.0",
     "pg": "^8.7.3",
     "pg-promise": "^10.10.2",
+    "puppeteer-cluster": "^0.23.0",
```
minotogna: Should  this go under dev dependencies ?
Doesn't it get imported from puppeteer ?

src/tools/screenshotWebpage.js
```diff
@@ -0,0 +1,296 @@
+const puppeteer = require('puppeteer')
```
minotogna: Better to use typescript ?

src/tools/screenshotWebpage.js
```diff
@@ -0,0 +1,296 @@
...
+  'X16',
+  'X17',
+  'X18',
+  'X19',
+  'X20',
+  'YEM',
+  'ZAF',
+  'ZMB',
+  'ZWE',
+]
```
minotogna: Better to fetch country isos ?

src/tools/screenshotWebpage.js
sorja: sure, this was just a quick tool to quickly get the request done but i can change it to TS

sorja:
@minotogna let me know if this should be converted to TypeScript
from my side we can leave it as it is in JavaScript, as it is rarely used and is more a reference if we need to run this tool again

minotogna (approved):

minotogna (changes requested):
Why not ts ?

minotogna:
> @minotogna let me know if this should be converted to TypeScript
> 
> from my side we can leave it as it is in JavaScript, as it is rarely used and is more a reference if we need to run this tool again

@sorja it's always preferred typescript

sorja:
> Why not ts ?

to avoid compiling, but i can change to ts!
will do at some point, i guess this is not extremely high priority :)

minotogna (approved):

PR #3479, opened 2024-02-07
3214 - Files and links per Cycle: Add possibility to set a file as public or private

sorja:
By default, all files are private

Notes on UI: Only added simple checkbox, refactor for UI will be last step

### DIFF
--- a/src/client/pages/CountryHome/Repository/Panel/InputFieldCheckbox/InputFieldCheckbox.tsx
+++ b/src/client/pages/CountryHome/Repository/Panel/InputFieldCheckbox/InputFieldCheckbox.tsx
@@ -1,0 +1,29 @@
+import React, { useCallback } from 'react'
+import { useTranslation } from 'react-i18next'
+
+type Props = {
+  onChange: (name: string, value: boolean) => void
+  value: boolean
+  name: string
+}
+
+const InputFieldCheckbox: React.FC<Props> = (props: Props) => {
+  const { onChange, value = false, name } = props
+  const { t } = useTranslation()
+
+  const _onChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => onChange(name, event.target.checked),
+    [name, onChange]
+  )
+
+  const id = `repository_edit-input-access`
+
+  return (
+    <div>
+      <input checked={value} id={id} onChange={_onChange} name="access" type="checkbox" />
+      <label htmlFor={id}>{t('common.public')}</label>
+    </div>
+  )
+}
+
+export default InputFieldCheckbox
--- a/src/client/pages/CountryHome/Repository/hooks/useOpenPanel.ts
+++ b/src/client/pages/CountryHome/Repository/hooks/useOpenPanel.ts
@@ -8,7 +8,6 @@ import { RepositoryActions } from 'client/store/ui/repository'
 const initialRepositoryItem: RepositoryItem = {
   name: '',
   link: '',
-  file: '',
 } as unknown as RepositoryItem
 
 export const useOpenPanel = (repositoryItem?: RepositoryItem) => {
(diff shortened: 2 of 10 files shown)


minotogna:
> By default, all files are private
> 
> 
> 
> Notes on UI: Only added simple checkbox, refactor for UI will be last step

Can we see anyway a screenshot ? Or video ?

sorja:
> > By default, all files are private
> > Notes on UI: Only added simple checkbox, refactor for UI will be last step
> 
> Can we see anyway a screenshot ? Or video ?

Sure, here is a screenshot, let me know if you want a video

image

minotogna:
> > > By default, all files are private
> 
> > > Notes on UI: Only added simple checkbox, refactor for UI will be last step
> 
> > 
> 
> > Can we see anyway a screenshot ? Or video ?
> 
> 
> 
> Sure, here is a screenshot, let me know if you want a video
> 
> 
> 
> image
> 
> 

That's fine 👍🏻

src/client/pages/CountryHome/Repository/Panel/InputFieldCheckbox/InputFieldCheckbox.tsx
```diff
@@ -0,0 +1,29 @@
...
+const InputFieldCheckbox: React.FC<Props> = (props: Props) => {
+  const { onChange, value = false, name } = props
+  const { t } = useTranslation()
+
+  const _onChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => onChange(name, event.target.checked),
+    [name, onChange]
+  )
+
+  const id = `repository_edit-input-access`
```
minotogna, line 19: Maybe no need to crate a const . But why do you need it ?

src/client/pages/CountryHome/Repository/hooks/useOpenPanel.ts
```diff
@@ -8,7 +8,6 @@ import { RepositoryActions } from 'client/store/ui/repository'
 const initialRepositoryItem: RepositoryItem = {
   name: '',
   link: '',
-  file: '',
 } as unknown as RepositoryItem
```
minotogna: Is there a need for this cast ?

minotogna: Or it can just be Partial ?

src/client/pages/CountryHome/Repository/Panel/InputFieldCheckbox/InputFieldCheckbox.tsx
sorja: It is added so that the label is clickable
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

src/client/pages/CountryHome/Repository/hooks/useOpenPanel.ts
sorja: Partial works fine, thanks

src/client/pages/CountryHome/Repository/Panel/InputFieldCheckbox/InputFieldCheckbox.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
```
minotogna, line 21: Don't we have a ButtonCheckbox directly ? Can't remember that

sorja: Yes, but this is just to test the flow is working, let's refactor later PR?

> Notes on UI: Only added simple checkbox, refactor for UI will be last step

minotogna: Ok 👍🏻

minotogna (approved):

PR #3488, opened 2024-02-09
3394 - Allow invite with permissions on the backend

yaguzmang:
Adding the possibility to add permissions when sending an invite, in the backend. UI PR coming next.

### DIFF
--- a/src/client/store/ui/userManagement/actions/inviteUser.ts
+++ b/src/client/store/ui/userManagement/actions/inviteUser.ts
@@ -4,7 +4,7 @@ import axios from 'axios'
 import { ApiEndPoint } from 'meta/api/endpoint'
 import { CycleParams } from 'meta/api/request'
 import { Lang } from 'meta/lang'
-import { RoleName } from 'meta/user'
+import { CollaboratorPermissions, RoleName } from 'meta/user'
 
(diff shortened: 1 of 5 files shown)


src/client/store/ui/userManagement/actions/inviteUser.ts
```diff
@@ -16,10 +16,20 @@ type Params = CycleParams & {
...
+  body?: Body
+  params: Params
+}
+
+export const inviteUser = createAsyncThunk<void, Props>(
   'userManagement/post/invitation',
-  async (params, { dispatch }) => {
-    const { status } = await axios.post(ApiEndPoint.User.invite(), null, { params })
+  async (props, { dispatch }) => {
+    const { body, params } = props
```
minotogna: I don't think there's a real need to split the properties in the action. it's enough to add permissions to the current props. Then it's possible to extract the props and pass to the request body the permissions.
The callers of the actions do not need to know what is passed in the params and in the body of the request. we better abstract and hide request implementations to components

yaguzmang: Yes, that makes more sense. Fixed it now, thanks!

minotogna (approved):

PR #3490, opened 2024-02-09
3437 - Refactor Descriptions style

minotogna:
part 5


src/client/style/_partials/_print.scss
```diff
@@ -0,0 +1,5 @@
+@mixin print {
```
minotogna, line 1: added common mixin to better isolate components print style

sorja (approved):

PR #3494, opened 2024-02-12
3214 - Files and links per Cycle: File upload

sorja:
https://github.com/openforis/fra-platform/assets/5508251/16da741a-4b95-4b3f-b9ec-d74be7ccc643


src/client/pages/CountryHome/Repository/Panel/InputFieldFile/InputFieldFile.tsx
```diff
@@ -1,16 +1,25 @@
...
+import FileUpload from 'client/components/FileUpload'
+
+import { useOnDrop } from './hooks/useOnDrop'
+
+// TODO: move
+const humanFileSize = (size: number) => {
+  const i = size === 0 ? 0 : Math.floor(Math.log(size) / Math.log(1024))
+  // @ts-ignore
+  return `${(size / 1024 ** i).toFixed(2) * 1} ${['B', 'kB', 'MB', 'GB', 'TB'][i]}`
+}
```
sorja: Should we move this to numbers, strings, somewhere else or keep here?

sorja: Maybe `Files.humanReadableSize (number) => string` ?

minotogna: Yes @sorja , please add this utility function

src/client/components/FileUpload/FileUpload.scss
```diff
@@ -0,0 +1,3 @@
+.progress-bar-container {
```
minotogna: I would still add a prefix here e.g., file-upload__progress-bar-container

src/client/pages/CountryHome/Repository/Panel/InputFieldFile/InputFieldFile.tsx
```diff
@@ -20,7 +29,15 @@ const FileInputField: React.FC = () => {
       <label htmlFor={id} className="repository-form__label">
         {t(label)}
       </label>
-      <input id={id} onChange={_onChange} className="repository-form__input" name="file" type="file" />
+      <FileUpload id={id} onDrop={onDrop} />
+      {file && (
```
minotogna: should this belong to FileUpload directly ?

src/client/pages/CountryHome/Repository/Panel/InputFieldFile/hooks/useOnDrop.ts
```diff
@@ -0,0 +1,37 @@
...
+    const hasFiles = files?.length > 0
+    if (!hasFiles) return
+
+    const file = files[0]
+    if (file.uuid !== repositoryItem?.fileUuid) {
+      onChangeField('fileUuid', files[0].uuid)
+    }
+  }, [files, onChangeField, repositoryItem?.fileUuid])
+
+  return useCallback(
```
minotogna: should this belong to FileUpload directly ?

src/client/pages/CountryHome/Repository/hooks/useClosePanel.ts
```diff
@@ -1,16 +1,15 @@
...
-import { useFileUploadContext } from 'client/components/FileUpload'
 
 type Returned = () => void
 
 export const useClosePanel = (): Returned => {
-  const { setFiles } = useFileUploadContext()
   const dispatch = useAppDispatch()
   return useCallback<Returned>(() => {
-    setFiles(null)
+    dispatch(FileUploadActions.reset())
```
minotogna: should this belong to FileUpload unmount hook directly ?

src/client/pages/CountryHome/Repository/Panel/InputFieldFile/hooks/useOnDrop.ts
sorja: I am not sure, I think in the WYSIWYG editor we want to pass props for each file

But I can move now and re-evaluate later when refactoring that part

src/client/components/FileUpload/FileUpload.scss
```diff
@@ -0,0 +1,3 @@
+.file-upload__progress-bar-container {
+  height: 25px;
```
minotogna: can we use spacing-XXXXXX ?

src/client/components/FileUpload/hooks/useOnDrop.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const { assessmentName, cycleName, countryIso } = useCountryRouteParams<CountryIso>()
+
+  useEffect(() => {
+    const hasFiles = files?.length > 0
+    if (!hasFiles) return
+
+    const file = files[0]
+    if (file.uuid !== repositoryItem?.fileUuid) {
+      onChangeField('fileUuid', files[0].uuid)
```
minotogna: FileUpload is a generic component. It must not have Repository visibility, it's the inverse.
When file is uploaded, => then Repository should take action .
FileUpload is only responsible for uploading and visualising files.

src/client/components/FileUpload/useOnPanelClose.ts
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from 'react'
+
+import { useAppDispatch } from 'client/store'
+import { FileUploadActions } from 'client/store/ui/fileUpload'
+
+export const useOnPanelClose = (): void => {
```
minotogna: this should be called without OnPanelClose . maybe useResetFilesOnUnmount ?

src/client/components/FileUpload/FileUpload.scss
sorja: Isn't it weird to use spacing values for sizes eg. height, width?

minotogna: It depends . In this case it's a small progress bar , I would use spacing-s .
What do u think ?

minotogna: But why does it need height ? 
Isn't it inherited by the child progress bar ?

sorja: Sorry, I mean:
Using spacing (margin, padding) values (eg spacing-X) for setting element sizes seems a bit off.
In this case we set the element height to the progress bars height (25px) so avoid jumping the buttons and other elements position when the progress bar appears

What do you suggest?

src/client/pages/CountryHome/Repository/Panel/hooks/useOnChange.ts
```diff
@@ -32,15 +32,33 @@ export const useOnChange = (): Returned => {
...
+
+  useEffect(() => {
+    const hasFiles = files?.length > 0
+    if (!hasFiles) return
+
+    const file = files[0]
+    // When a file is selected, update the fileUuid
+    if (file.uuid !== repositoryItem?.fileUuid) {
+      onChangeField('fileUuid', files[0].uuid)
+    }
```
minotogna: this is the same then lines 36-42. can we avoid this duplication ?

sorja: Good eye, it was a typo left from refactoring

src/client/pages/CountryHome/Repository/Panel/hooks/useOnChange.ts
```diff
@@ -32,15 +32,23 @@ export const useOnChange = (): Returned => {
...
+
+    // When a file is selected and the name is empty,
+    // set the name to the file name
+    if (repositoryItem?.name === '') {
```
minotogna: maybe better Objects.isEmpty? or, now that I think, if a new file is uploaded, should we always update the name? just wondering, what do you think?

sorja: Sure, but we might lose some user inputted data

src/client/pages/CountryHome/Repository/Panel/hooks/useOnChange.ts
```diff
@@ -32,15 +33,16 @@ export const useOnChange = (): Returned => {
...
+
+    const file = files[0]
+    const fileUuid = file.uuid
+    if (fileUuid !== repositoryItem?.fileUuid) {
       const name = file.name.split('.')[0]
-      onChangeField('name', name)
+      dispatch(RepositoryActions.setRepositoryItem({ ...repositoryItem, name, fileUuid }))
     }
-  }, [files, onChangeField, repositoryItem?.name])
+  }, [dispatch, files, onChangeField, repositoryItem, repositoryItem?.fileUuid, repositoryItem?.name])
```
minotogna: do we need repositoryItem?.name in the dependency hook ?

src/client/pages/CountryHome/Repository/Panel/hooks/useOnChange.ts
```diff
@@ -32,15 +33,16 @@ export const useOnChange = (): Returned => {
...
+
+    const file = files[0]
+    const fileUuid = file.uuid
+    if (fileUuid !== repositoryItem?.fileUuid) {
       const name = file.name.split('.')[0]
-      onChangeField('name', name)
+      dispatch(RepositoryActions.setRepositoryItem({ ...repositoryItem, name, fileUuid }))
     }
-  }, [files, onChangeField, repositoryItem?.name])
+  }, [dispatch, files, repositoryItem])
```
minotogna: this hook will be triggered every time a repositoryItem updates. would it be better to add only repositoryItem?.uuid ?

minotogna: I think it might be more efficient to have an action setRepositoryItemProp that takes only the name and value. what do you think ?

minotogna: or props as Partial

sorja: sure, ill add it

src/client/pages/CountryHome/Repository/Panel/hooks/useOnChange.ts
```diff
@@ -15,32 +19,35 @@ export const useOnChange = (): Returned => {
...
-      const file = files[0]
+    if (!files || files.length <= 0) return
+
+    const file = files[0]
+    const fileUuid = file.uuid
+    if (!Objects.isEmpty(repositoryItem) && fileUuid !== repositoryItem?.fileUuid) {
       const name = file.name.split('.')[0]
-      onChangeField('name', name)
+      dispatch(RepositoryActions.setRepositoryItemProp({ key: 'fileUuid', value: fileUuid }))
+      dispatch(RepositoryActions.setRepositoryItemProp({ key: 'name', value: name }))
```
minotogna: Miro, can we have setRepositoryItemProps that takes in input Partial<.... ?

src/client/pages/CountryHome/Repository/Panel/hooks/useOnChange.ts
```diff
@@ -15,32 +19,35 @@ export const useOnChange = (): Returned => {
...
-  // When a file is selected and the name is empty,
-  // set the name to the file name
   useEffect(() => {
-    if (files?.length > 0 && repositoryItem?.name === '') {
-      const file = files[0]
+    if (!files || files.length <= 0) return
+
+    const file = files[0]
+    const fileUuid = file.uuid
+    if (!Objects.isEmpty(repositoryItem) && fileUuid !== repositoryItem?.fileUuid) {
```
minotogna: why should this be empty ?

sorja: When we close the panel, we set the repository item to `undefined`
If we don't set this, the panel keeps updating and re-opening

src/client/pages/CountryHome/Repository/Panel/hooks/useOnChange.ts
sorja: sure, can you elaborate a bit what u mean with partial?
what should the call look like?



> Miro, can we have setRepositoryItemProps that takes in input Partial<.... ?

you mean that we use it as `setRepositoryItemProps({ repositoryItem: { name: 'a' } })` and
`state.repositoryItem = { ...state.repositoryItem, ...payload }`

From here I understood so:

> I think it might be more efficient to have an action setRepositoryItemProp that takes only the name and value. what do you think ?

minotogna (approved):

PR #3499, opened 2024-02-12
3394 - UI: invite collaborator with permissions

yaguzmang:
Adding the collaborator:

https://github.com/openforis/fra-platform/assets/41337901/b72a1eb1-7d21-4a80-8e05-656e97af10dc

Checking the permissions:
Table Data: 1a, 1b, 1c
Descriptions: 1a, 1d

https://github.com/openforis/fra-platform/assets/41337901/7cc3e381-bfb7-4779-97b1-733f1e908ce2


src/client/components/InviteUserForm/hooks/initialState.ts
```diff
@@ -5,5 +5,8 @@ import { UserToInvite } from 'client/components/InviteUserForm/userToInvite'
 
 export const useInitialState = (): UserToInvite => {
   const language = useLanguage()
-  return useMemo(() => ({ name: '', surname: '', email: '', role: '', lang: language }), [language])
+  return useMemo(
+    () => ({ name: '', surname: '', email: '', role: '', lang: language, permissions: { sections: 'all' } }),
```
minotogna: permissions should be here only when user is collaborator, otherwise we save this permissions in all user role props. isn't it ?

src/client/components/EditUserForm/CollaboratorPermissions/CollaboratorPermissions.tsx
```diff
@@ -33,52 +37,40 @@ const CollaboratorPermissions = (props: Props) => {
...
-        </div>
+        <CollaboratorAccessList permissions={permissions} />
 
         <button className="btn-xs btn-primary" onClick={_onEditPermissionsClick} type="button">
           {t('userManagement.editPermissions')}
         </button>
       </div>
 
-      <CollaboratorAccessModal open={modalOptions.open} userRole={userRole} onClose={_onEditPermissionsClose} />
+      <CollaboratorAccessModal
```
minotogna: I think this a good moment to organise this.
From outside, we don't need to render every time the modal, only the PermissionsEdit component .
the PermissionsEdit component will render the modal by itself if needed.

I propose to extract the components within client/components/user:
CollaboratorPermissionsEditor and inside itself render the ModalEditor that becomes an inner component of CollaboratorPermissionsEditor.
(Please take note of the components name).

What do you think @yaguzmang @sorja ?

sorja: Sounds good!

src/client/components/InviteUserForm/hooks/initialState.ts
yaguzmang: There is a [check](https://github.com/openforis/fra-platform/pull/3499/commits/6971d12ab2d915fd6d6537e6dc6ebd78f0afbe1d#diff-a549cfe6efbb3f5b71ac8e36d235d6d51e57294edd89ee635e4bbde654923192R51) for this in the `useOnUserInvite` hook. But I'll remove this initial state so it's not confusing.

src/client/components/EditUserForm/CollaboratorPermissions/CollaboratorPermissions.tsx
yaguzmang: Added `CollaboratorPermissionsEditor` to  `client/components/user` and `ModalEditor` within it. It's more organized now, thanks!

minotogna: what I meant is that CollaboratorAccessList becomes CollaboratorPermissionsEditor, so that only that component must be rendered. I see now you always render them in couple, which is redundant .
The CollaboratorPermissionsEditor displays the permissions + handle the logic to open the modal.

yaguzmang: What do you think of moving`<CollaboratorAccessList>` to be an inner component of the current `<CollaboratorPermissionsEditor>`, and then the entire editor would be:
```
CollaboratorPermissionsEditor:
    <>
      <CollaboratorAccessList/>

      <button/>

      <ModalEditor />
    </>
``` 
So that the logic of `<CollaboratorAccessList>` (generating the table and description permissions) is kept separate. 
Then I also only render `<CollaboratorPermissionsEditor>` as you suggested above.

minotogna: Sure , just rename the ColaboratorAccessList to Permissions .
Then try to isolate the style , you can even return a div instead of the fragment

yaguzmang: Added 👍 I reorganized the styles too (looks the same).

minotogna (approved):

PR #3527, opened 2024-02-16
3394 - Bug fix: prevent accepting an already accepted invitation

yaguzmang:
While looking at the activity log, found a bug where a user can accept an invitation more than once.
Cause: The UI didn't handle loading state when the submit button was clicked, and the backend didn't check if the invitation was already accepted or not. 

Before:

https://github.com/openforis/fra-platform/assets/41337901/55941da5-7481-469f-89b9-00b93ac45c14

Now (the form looks like this because the branch doesn't have the latest dev changes):

https://github.com/openforis/fra-platform/assets/41337901/39deaf57-9da0-4700-b7dd-7b3d03068926

### DIFF
--- a/src/server/controller/user/acceptInvitation.ts
+++ b/src/server/controller/user/acceptInvitation.ts
@@ -23,7 +23,7 @@ export const acceptInvitation = async (
 
   return client.tx(async (t) => {
     if (UserRoles.isInvitationExpired(userRole)) throw new Error('login.invitationExpired')
-    if (!UserRoles.isInvitationPending(userRole)) throw new Error('login.alreadyAcceptedInvitation')
+    if (userRole.acceptedAt !== null) throw new Error('login.alreadyAcceptedInvitation')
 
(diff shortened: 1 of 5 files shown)


src/server/controller/user/acceptInvitation.ts
```diff
@@ -23,7 +23,7 @@ export const acceptInvitation = async (
 
   return client.tx(async (t) => {
     if (UserRoles.isInvitationExpired(userRole)) throw new Error('login.invitationExpired')
-    if (!UserRoles.isInvitationPending(userRole)) throw new Error('login.alreadyAcceptedInvitation')
+    if (userRole.acceptedAt !== null) throw new Error('login.alreadyAcceptedInvitation')
```
yaguzmang, line 26: The dates seem to be considered empty by Objects.isEmpty
Screenshot 2024-02-16 at 2 33 49 PM

minotogna (approved):

PR #3536, opened 2024-02-19
3214 - Files and links per Cycle: Static links

sorja:
image

image

example fra 2020
image


src/meta/cycleData/repository/repository.ts
```diff
@@ -1,8 +1,17 @@
 import { AreaCode } from 'meta/area'
+import { Lang } from 'meta/lang'
+
+export type Translations = {
```
minotogna: I was thinking this could actually be a generic type isolated . Under meta/translation . What do you think ?

Also , there's no real need to make English mandatory . All Lange optional I would say . What do you think ?

src/meta/cycleData/repository/repositoryItems.ts
```diff
@@ -17,6 +18,12 @@ const getURL = (props: GetFileURLProps) => {
   return `${ApiEndPoint.CycleData.Repository.file(datum.uuid)}?${queryParams.toString()}`
 }
 
+const getName = (props: { repositoryItem: RepositoryItem; language: Lang }): string => {
+  const { repositoryItem, language = Lang.en } = props
+  return repositoryItem.props.translations[language] ?? repositoryItem.props.translations.en
+}
+
 export const RepositoryItems = {
+  getName,
```
minotogna: Also , I would create a generic Translations.getLabel

src/meta/cycleData/repository/repository.ts
sorja: Sure, should English be optional in both Translation generic type and RepositoryItem or be mandatory in only RepositoryItem?

minotogna: why do you want to make en mandatory in translations ? wouldn't it be easier to just delegate validations to the users ? in this case repository item ?

src/meta/cycleData/repository/repositoryItems.ts
minotogna: Should we have a Translations utility instead ?

sorja: It is using it:

```
return Translations.getLabel({ translation: repositoryItem.props.translation, language })
```

minotogna: That's where I get confused. I thought that this RepositoryItems.getName is superfluous, you could use Translations.getLabel({translation:repositoryItems.props.translation, language") directly from outside . less code.
what do you think ?

sorja: sure, updated

minotogna (approved):

PR #3540, opened 2024-02-20
3214 - Files and links per Cycle: Admin can add global Repository Items

sorja:
- Also hide edit buttons when UI lock is locked
- Default sort by ID repository items so they don't jump in the list

NC View:

image

Admin view (Unlocked):
image

Admin view (Locked):
image


sorja:
@minotogna Let me know if you prefer a video
The UI's only change is that we hide the action when UI lock is locked.

src/client/pages/CountryHome/Repository/Actions/Actions.tsx
```diff
@@ -1,27 +1,34 @@
...
   repositoryItem: RepositoryItem
 }
 
 const Actions: React.FC<Props> = (props) => {
   const { repositoryItem } = props
-  const dispatch = useAppDispatch()
+  const user = useUser()
+  const isAdmin = Users.isAdministrator(user)
+  const openPanel = useOpenPanel({ repositoryItem })
+  const locked = useIsDataLocked()
```
minotogna: we better don't use in this case useIsDataLocked, but we create specific auth hooks. 
add 2 hooks (for now) in src/client/store/user/hooks/index.ts (bottom) called useIsCountryRepositoryEditable and useIsGlobalRepositoryEditable

src/client/pages/CountryHome/Repository/ButtonOpenPanel/ButtonOpenPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const ButtonOpenPanel: React.FC<Props> = (props: Props) => {
+  const { isAdmin } = props
+  const { countryIso } = useCountryRouteParams<CountryIso>()
+  const { t } = useTranslation()
+
+  const openPanel = useOpenPanel({ countryIso: isAdmin ? undefined : countryIso })
+
+  return (
+    <button className="btn-s btn-primary" onClick={openPanel} type="button">
```
minotogna: let's call this ButtonAdd plus let's use new Button component (size m)

src/client/pages/CountryHome/Repository/ButtonOpenPanel/ButtonOpenPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+type Props = {
+  isAdmin?: boolean
+}
+
+const ButtonOpenPanel: React.FC<Props> = (props: Props) => {
+  const { isAdmin } = props
+  const { countryIso } = useCountryRouteParams<CountryIso>()
+  const { t } = useTranslation()
+
+  const openPanel = useOpenPanel({ countryIso: isAdmin ? undefined : countryIso })
```
minotogna: admin can also be able to add country items + repository items. does it work this way?

minotogna:
> @minotogna Let me know if you prefer a video The UI's only change is that we hide the action when UI lock is locked.

@sorja a screenshot might be helpful, for example do we have two add buttons ?

src/client/pages/CountryHome/Repository/Actions/Actions.tsx
sorja: Sure, do we still want to show the Edit Action even though the UI data lock is closed?

src/client/pages/CountryHome/Repository/ButtonOpenPanel/ButtonOpenPanel.tsx
sorja: yes

src/client/pages/CountryHome/Repository/Actions/Actions.tsx
minotogna: no, edit actions are visible only is editable

src/client/pages/CountryHome/Repository/Repository.tsx
```diff
@@ -27,15 +29,11 @@ const Repository: React.FC = () => {
         header={false}
         path={`${ApiEndPoint.CycleData.Repository.many()}?global=true`}
       />
+      {isAdmin && <ButtonAdd isAdmin={isAdmin} />}
```
minotogna: I think isAdmin here is superfluous since the check is also performed within the ButtonAdd itself. plus, I would call the prop isGlobal to indicate that we are adding a global file.

minotogna: also, are the two ButtonAdd visible only if repository is editable ?

src/client/store/user/hooks/index.ts
```diff
@@ -70,3 +70,15 @@ export const useIsDescriptionEditable = (props: {
 
   return useMemo<boolean>(() => canEdit && editEnabled, [canEdit, editEnabled])
 }
+
+export const useIsCountryRepositoryEditable = (): boolean => {
+  const locked = useIsDataLocked()
+  return !locked
```
minotogna, line 76: this should also check user rights (same below)

src/client/pages/CountryHome/Repository/Repository.tsx
sorja: We cannot check for isAdmin in the ButtonAdd, as we need to render two buttons
Should we pass a prop `hidden={!isAdmin}` for first button or keep as is?

sorja: Resolved with:

```
<ButtonAdd isGlobal />

...
  if (isGlobal && !isGlobalRepositoryEditable) {
    return null
  }
```

I think this is what you meant :)

src/client/store/user/hooks/index.ts
sorja: Sounds good, I need to update the Auth methods anyway (remove references to AssessmentFile)
If it's ok with you, I will open in next PR :
- Update FE (these 2 methods), update the authorizer itself, update backend, remove old code

Let me know

minotogna:
Are the add buttons visible only if editable is true ?

src/client/store/user/hooks/index.ts
minotogna: Sure

minotogna (approved):

PR #3551, opened 2024-02-22
3214 - Files and links per Cycle: Error on changing a file to private that is in use and public

sorja:
🟢 Adding files works for global and country
🟢 Editing repostiory item name doesn't change file name in edit screen

https://github.com/openforis/fra-platform/assets/5508251/bf0628d7-54af-4ed1-b6ae-771e6442d69a

🟢 Selecting a file in repository dialog makes it public
🟢 Unable to change file from public to private if it's in use
🟢 Show error

https://github.com/openforis/fra-platform/assets/5508251/949858e1-eaa7-45fe-8fb3-c8888721be6e


src/client/pages/CountryHome/Repository/EditForm/InputFieldFile/SelectedFile/SelectedFile.tsx
```diff
@@ -3,22 +3,23 @@ import React from 'react'
...
+  const repositoryFile = useRepositoryFile()
 
   if (!repositoryItem) {
     return null
   }
 
   return (
     <div className="repository-form__selected-file">
-      <RepositoryLink datum={repositoryItem as RepositoryItem} />
+      <RepositoryLink datum={repositoryItem as RepositoryItem} name={repositoryFile.name} />
```
sorja, line 22: In select file we want to show the file name that the user has selected, and not the repository item name that is _public_ / visible in the ui

minotogna: that is the link. and don't we want show the label only ?

minotogna: Then maybe we only need a generic FileLink which accepts only one prop. the label to display.

src/meta/api/endpoint/ApiEndPoint.ts
```diff
@@ -75,6 +75,7 @@ export const ApiEndPoint = {
       file: (uuid = ':uuid') => apiPath('cycle-data', 'repository', 'file', uuid),
       many: () => apiPath('cycle-data', 'repository', 'items'),
       one: () => apiPath('cycle-data', 'repository', 'items', 'item'),
+      updateAccess: () => apiPath('cycle-data', 'repository', 'update-access'),
```
minotogna, line 78: this should be /cycle-data/repository/items/item/access

src/server/repository/assessmentCycle/repository/index.ts
```diff
@@ -10,4 +11,5 @@ export const RepositoryRepository = {
   update,
   getMany,
   getOne,
+  getUsage,
```
minotogna, line 14: getUsages

src/client/pages/CountryHome/Repository/EditForm/hooks/useOnChange.ts
```diff
@@ -29,13 +34,36 @@ export const useOnChange = (): Returned => {
...
+      const params = { repositoryItem: _repositoryItem, countryIso, assessmentName, cycleName }
+      dispatch(RepositoryActions.updateRepositoryItemAccess(params))
+    },
+    [assessmentName, countryIso, cycleName, dispatch, repositoryItem]
+  )
+
+  const onChangeAccess = useCallback<OnChange>(
+    (name: string, value: boolean | string) => {
+      if (repositoryItem?.uuid) {
+        _updateExistingAccess(value as boolean)
```
minotogna, line 59: I wouldn't make a separate request to update the access only, because the user sees 'Save' that means the changes will be persisted only on Save click. instead this doesn't , but it changes it on the fly

src/meta/api/endpoint/ApiEndPoint.ts
minotogna: check my last comment, we don't need this endpoint

PR #3570, opened 2024-02-28
3496 - Update Mosaic layer redux & Add fetch hook

yaguzmang:
Updating some mosaic layer redux state types and logic & adding a use fetch hook. The logic to fetch remains mostly the same as it is now in the component `GeoMapMenuMosaic`, but abstracted to a hook and called from `useMapLayersHandler`.

The wireframe UI is not included in this PR, but just to show how it works:

https://github.com/openforis/fra-platform/assets/41337901/1a811ae1-0e10-4e97-bb69-3794c012964d


src/client/pages/Geo/GeoMap/hooks/useFetchMosaicLayer.ts
```diff
@@ -0,0 +1,43 @@
...
+  useAppliedMosaicOptions,
+  useMosaicSelected,
+  useMosaicStatus,
+  useMosaicUrl,
+} from 'client/store/ui/geo'
+import { LayerFetchStatus } from 'client/store/ui/geo/stateType'
+import { useCountryIso } from 'client/hooks'
+import { mapController } from 'client/utils'
+
+export const useFetchMosaicLayer = () => {
```
minotogna, line 17: should this be responsibility of the Mosaic Section ?

yaguzmang: In this case, if this hook was in the Mosaic Section, the re fetch would depend on whether or not that component is currently rendered. If the user were to change the country while the `Satellite Mosaic` nav section is not collapsed, then the mosaic layer wouldn't be fetched for the new country because the hook is not present. 
It worked in the previous UI because the logic was in the "Background" button, which was always on the screen.

yaguzmang: To handle the country iso change for the other layers we have the hook `useCountryIsoChangeHandler`. I can add some more logic there to also fetch the mosaic layer, and then I can move the rest of this hook to the Mosaic Section component. What do you think?

minotogna: That's fine as it is 👍🏾let's see later on

minotogna (approved):

PR #3572, opened 2024-02-28
3214 - Files and links per Cycle: Allow file download in side panel

sorja:
https://github.com/openforis/fra-platform/assets/5508251/2b05e24d-edae-498c-a821-3ab2084209af

### DIFF
--- a/src/client/pages/CountryHome/Repository/EditForm/InputFieldFile/InputFieldFile.tsx
+++ b/src/client/pages/CountryHome/Repository/EditForm/InputFieldFile/InputFieldFile.tsx
@@ -29,12 +29,7 @@ const FileInputField: React.FC = () => {
         </label>
       </DataCell>
       <DataCell editable error={Boolean(error)} lastCol lastRow>
-        <FileUpload
-          canDownload
-          id={id}
-          onChange={onChange}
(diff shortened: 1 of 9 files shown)


src/client/pages/CountryHome/Repository/EditForm/InputFieldFile/InputFieldFile.tsx
```diff
@@ -30,7 +30,7 @@ const FileInputField: React.FC = () => {
       </DataCell>
       <DataCell editable error={Boolean(error)} lastCol lastRow>
         <FileUpload
-          canDownload={false}
+          canDownload
```
minotogna: do we still need this prop ? anyway files can be download only and only if we pass value. what do u think ?

minotogna (commented):
good pr 👍 just a doubt about a prop

src/client/pages/CountryHome/Repository/EditForm/InputFieldFile/InputFieldFile.tsx
sorja: you are correct

minotogna (approved):

PR #3575, opened 2024-02-28
3214 - Files and links per Cycle: Confirm delete

sorja:
Add generic label 'are you sure'

### DIFF
--- a/src/i18n/resources/en/common.js
+++ b/src/i18n/resources/en/common.js
@@ -3,6 +3,7 @@ module.exports = {
   add: 'Add',
   apply: 'Apply',
   areYouSureYouWantToDeleteAllTableData: 'Are you sure you want to delete all data from the table?',
+  areYouSure: 'Are you sure?',
   cancel: 'Cancel',
   clearTable: 'Clear table',
   column: 'Column',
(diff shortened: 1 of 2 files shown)


src/i18n/resources/en/common.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   add: 'Add',
   apply: 'Apply',
   areYouSureYouWantToDeleteAllTableData: 'Are you sure you want to delete all data from the table?',
+  areYouSure: 'Are you sure?',
```
minotogna: Not very nice message . Don't we have already a better message ?

minotogna: Maybe : 
Are you sure you want to delete this item? This operation cannot be undone.

sorja: I know, I wanted to create a more generic message that could be used elsewhere, any thoughts?

`Are you sure you want to continue?`
`Are you sure you want to ...`?

Let me know if you can find a generic confirmation, otherwise let's use what you wrote here!

minotogna: It's a good idea , but maybe my message above is already for generic delete . Then if we need other will see . What do you think?

sorja: Sure, let's do so

minotogna (approved):

PR #3585, opened 2024-03-01
3214 - Files and links per Cycle: Add translations

sorja:


src/i18n/resources/en/validation.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   repositoryItem: {
     eitherFileOrLink: 'Only one between file and link must be specified',
-    fileOrLink: 'One between link and file is required',
+    fileOrLink: 'Either a file or a link is required',
```
minotogna, line 4: This might not be consistent with translations asked

sorja: I will check, thanks

minotogna (approved):

PR #3588, opened 2024-03-01
3214 - Files and links per Cycle: remove unused and deprecated code

sorja:
Added new todos in main PR:

```
 Get hidden files ( getHiddenFile.ts )
Update import private files tools/importPrivateFiles/index.ts
```


src/server/api/file/getHiddenFile.ts
```diff
@@ -2,21 +2,21 @@ import { Request, Response } from 'express'
...
   try {
-    const { assessmentName, fileName } = req.query
+    // const { assessmentName, fileName } = req.query
 
-    const assessment = await AssessmentController.getOne({ assessmentName })
-    const assessmentFile = await FileController.getHiddenAssessmentFile({ assessment, fileName })
-
-    Responses.sendAssessmentFile(res, assessmentFile)
+    // const assessment = await AssessmentController.getOne({ assessmentName })
+    // TODO
```
minotogna, line 16: has the url of hidden file changed? with this I think we shouldn't change it, or at least ask where they are user

src/tools/importPrivateFiles/index.ts
```diff
@@ -7,9 +7,11 @@ import * as path from 'path'
 import { AssessmentNames } from 'meta/assessment'
 
 import { AssessmentController } from 'server/controller/assessment'
-import { AssessmentFileRepository } from 'server/repository/assessment/file'
 import { Logger } from 'server/utils/logger'
 
+// TODO
```
minotogna: what is this ? 😅

sorja: Should we remove this tool?
I added it for now, let's implement when it's needed again?

src/server/api/file/getHiddenFile.ts
sorja: Should we introduce redirect with `HTTP 301`?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301

src/tools/importPrivateFiles/index.ts
minotogna: oh right, totally forgot about this ;P . sure let's get rid of it for now. we might even build an admin page where they can do this from there directly. what do u think ?

src/server/api/file/getHiddenFile.ts
minotogna: yes, I think so. those links where shared with people. another pr ?

sorja: Yes

minotogna (approved):

PR #3589, opened 2024-03-01
3496 - Geo UI Refactor - Add recipe selector

yaguzmang:
Adding the Recipe Selector component.
The Forest Layers component is not included in this PR. It is included in the video just to show how it would look. I will be adding later a component that renders the sections dynamically with the current `sections` array we have in `meta/geo`:
`export const sections: Array<LayerSection> = [forestLayers, protectedAreaLayers, burnedAreaLayers]`

https://github.com/openforis/fra-platform/assets/41337901/ae5212d6-21a6-48c1-9553-b6516b7b8908

And, fixing the Beta message which was overlapping with the new components in the toolbar:

Before:

https://github.com/openforis/fra-platform/assets/41337901/2c010242-eafb-4932-9425-892f7aba131d

Now:

https://github.com/openforis/fra-platform/assets/41337901/4f04fc62-d0d0-439b-949e-d66e52497017

### DIFF
--- a/src/client/components/Inputs/Select/Select.tsx
+++ b/src/client/components/Inputs/Select/Select.tsx
@@ -11,7 +11,7 @@ import { useValue } from './hooks/useValue'
 import { SelectProps } from './types'
 
 const Select: React.FC<SelectProps> = (props) => {
-  const { disabled, isMulti, options } = props
+  const { disabled, isClearable, isMulti, options } = props
 
   const value = useValue(props)
   const onChange = useOnChange(props)
@@ -35,7 +35,7 @@ const Select: React.FC<SelectProps> = (props) => {
       }}
       closeMenuOnSelect={!isMulti}
       components={{ ClearIndicator, DropdownIndicator, IndicatorsContainer, IndicatorSeparator: null }}
-      isClearable
+      isClearable={isClearable}
       isDisabled={disabled}
       isMulti={isMulti}
       isSearchable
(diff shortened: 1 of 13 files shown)


src/client/components/Inputs/Select/Select.tsx
```diff
@@ -49,4 +49,9 @@ const Select: React.FC<SelectProps> = (props) => {
   )
 }
 
+Select.defaultProps = {
+  // eslint-disable-next-line react/default-props-match-prop-types
+  isClearable: true,
```
yaguzmang, line 54: Eslint has a problem with the optional 'isClearable' property from `Pick<ReactSelectProps, 'isClearable' | 'isMulti'> & ...`:

image

After trying some fixes, the problem seems to be that the type declaration is in another file and eslint doesn't record it. The same error pops up if we add a default value for `disabled`, even though it is defined as `disabled?: boolean`.

One option to avoid this `eslint-disable-next-line` would be to make the prop non-optional and add `isClearable={true}` to all instances where the Select component is currently used (not ideal). I appreciate any suggestions.

src/client/components/Navigation/NavGeo/RecipeSelector/RecipeSelector.scss
```diff
@@ -0,0 +1,14 @@
+@import 'src/client/style/partials';
+
+.geo-recipe-selector {
+  background-color: $ui-bg-hover;
```
yaguzmang, line 4: This is how it would look like with white instead:

https://github.com/openforis/fra-platform/assets/41337901/f1e90b4d-a41f-4302-9de7-982b645e0b65

minotogna (approved):

PR #3601, opened 2024-03-07
3496 - Geo UI Refactor - Agreement Layer

yaguzmang:
Adding a refactored Agreement Layer Control UI component with its Reducer Scale selector:

https://github.com/openforis/fra-platform/assets/41337901/c47c1657-cef4-4543-b827-a81e2a00b86f

Now the checkbox color of the agreement layer matches the selected level color instead of having a generic one.


src/client/components/Navigation/NavGeo/AgreementLevelSelector/AgreementLevelSelector.scss
```diff
@@ -0,0 +1,42 @@
...
+    appearance: none;
+    border-radius: 3px;
+    border: 1px solid #949495;
+    cursor: pointer;
+    height: $spacing-s;
+    width: $spacing-s;
+
+    &:checked {
+      background-color: #262627;
+      border-color: #5f5f5f;
```
yaguzmang: I tried using existing colors for these and other instances in this file but it looked off. Should I add new color variables for them?

src/client/components/Navigation/NavGeo/AgreementLevelSelector/AgreementLevelSelector.scss
```diff
@@ -0,0 +1,42 @@
+@import 'src/client/style/partials';
+
+.geo-agreement-level-selector__container {
+  display: flex;
+  justify-content: space-between;
+
+  .geo-agreement-level-selector__checkbox-container {
```
minotogna: No need to nest css if there's the prefix

src/client/components/Navigation/NavGeo/AgreementLevelSelector/AgreementLevelSelector.scss
minotogna: Can u add here a screenshot of these specific usages ?

yaguzmang: Sure

For this example with level 3 selected:
image


`border: 1px solid #949495;`: Border for level 4 (selectable and higher than current selected)

`border-color: #5f5f5f;`: Border for checked levels (1 up to 3 in this case).

`background-color: #262627`: Background for levels 1 & 2 (lower than selected level)

 `border-color: #dedfdf;`& `box-shadow: 0 0 10px 2px #efeeee inset;`: Styles for levels 5, 6 and the other disabled levels.

minotogna: not sure now. let's leave it as it is. we'll check this later on ;)

minotogna (approved):

PR #3613, opened 2024-03-11
3496 - Geo UI Refactor - Add empty Statistics Side Panel

yaguzmang:
Moving and resizing the map buttons, and adding the (empty) statistics side panel:

https://github.com/openforis/fra-platform/assets/41337901/33c5307f-ed34-4fa7-94f5-b7603eb5c26b

Zooming in on the panel and button: 

https://github.com/openforis/fra-platform/assets/41337901/ccc8a929-6f39-48ea-86db-8a3a51345fd1

### DIFF
--- a/src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
+++ b/src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
@@ -4,7 +4,6 @@ import { CountryIso } from 'meta/area'
 
 import { useAppDispatch } from 'client/store'
 import { GeoActions } from 'client/store/ui/geo'
-import { usePrevious } from 'client/hooks'
 import { useCountryRouteParams } from 'client/hooks/useRouteParams'
 import { getCountryBounds } from 'client/pages/Geo/utils/countryBounds'
 import { mapController } from 'client/utils'
@@ -28,7 +27,6 @@ export const useGeoMap = (props: Props): Returned => {
   const [map, setMap] = useState<google.maps.Map>()
 
   const { countryIso } = useCountryRouteParams<CountryIso>()
-  const previousCountryIso = usePrevious(countryIso, countryIso)
 
   useEffect(() => {
     if (!ref.current || map) return
@@ -62,19 +60,11 @@ export const useGeoMap = (props: Props): Returned => {
     mapController.setMap(mapSetup)
     setMap(mapSetup)
     dispatch(GeoActions.setMapAvailability(true))
-
-    getCountryBounds(countryIso).then((response) => {
-      if (response?.data) {
-        mapSetup.panTo(response.data.centroid)
-        mapSetup.fitBounds(response.data.bounds)
-      }
-    })
   }, [countryIso, dispatch, map, ref, viewport, zoom])
 
   // Move and center the map to the new country location.
   useEffect(() => {
-    if (countryIso === previousCountryIso) return
-    if (!map) return
+    if (!map || !countryIso) return
 
     getCountryBounds(countryIso).then((response) => {
       if (response?.data) {
@@ -82,7 +72,7 @@ export const useGeoMap = (props: Props): Returned => {
         map.fitBounds(response.data.bounds)
       }
     })
-  }, [countryIso, previousCountryIso, map])
+  }, [countryIso, map])
 
   return { map, ref }
 }
--- a/src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
+++ b/src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
@@ -1,0 +1,87 @@
+import { MutableRefObject, useEffect, useRef, useState } from 'react'
+
+import { useAppDispatch } from 'client/store'
+import { GeoActions } from 'client/store/ui/geo'
+import { useCountryIso, usePrevious } from 'client/hooks'
+import { getCountryBounds } from 'client/pages/Geo/utils/countryBounds'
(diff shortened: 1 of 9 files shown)


src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const useGeoMap = (props: Props): Returned => {
+  const { viewport, zoom } = props
+
+  const dispatch = useAppDispatch()
+
+  const ref = useRef<HTMLDivElement>(null)
+  const [map, setMap] = useState<google.maps.Map>()
+
+  const countryIso = useCountryIso()
```
minotogna: This can be taken from useCountryRouteParams

src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
```diff
@@ -0,0 +1,87 @@
...
+      mapTypeControlOptions: {
+        mapTypeIds: [google.maps.MapTypeId.ROADMAP, google.maps.MapTypeId.SATELLITE, google.maps.MapTypeId.HYBRID],
+        position: google.maps.ControlPosition.TOP_CENTER,
+        style: google.maps.MapTypeControlStyle.DROPDOWN_MENU,
+      },
+      mapTypeId: google.maps.MapTypeId.HYBRID,
+      rotateControl: true,
+      zoom,
+      zoomControl: true,
+      zoomControlOptions: {
```
minotogna: Can we horizontally align the zoom controls ?

src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (response?.data) {
+        mapSetup.panTo(response.data.centroid)
+        mapSetup.fitBounds(response.data.bounds)
+      }
+    })
+  }, [countryIso, dispatch, map, ref, viewport, zoom])
+
+  // Move and center the map to the new country location.
+  useEffect(() => {
+    if (countryIso === previousCountryIso) return
```
minotogna: Is the check with previous county iso really needed ?

src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
yaguzmang: Natively it is only possible to position them, it doesn't seem possible to align them. Another option could be to have the zoom on the left center and the full screen button in the top left, like in this [example](https://developers.google.com/maps/documentation/javascript/examples/control-positioning) (swapping street view with full screen):

image

src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
yaguzmang: It prevents an unnecessary extra call to the BE on the first render

minotogna: strange, even if you check if map and countryIso ?

minotogna: actually because you are calling it twice, see line 66/71.
I think you only need this hook, and remove lines 66/71. so you will be able to safely  remove the check with previous country iso .
can u verify ?

src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
minotogna: I would suggest to move all controls top - right of the map and small size. make space above the panel.
what do u think ? 
plus, I see in the video only satellite option, now we have map and satellite.

src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
yaguzmang: Good point. It is indeed not needed to do it in the setup useEffect. I removed it there, it works as previously. Thanks!

src/client/pages/Geo/GeoMap/hooks/useGeoMap.ts
yaguzmang: Sure. Will add them to the top right corner. And the Map/Satellite is now a dropdown. I changed it so it takes less space. Should I revert it back to the style in the example above? (Clickable Map | Satellite text instead of a dropdown).

yaguzmang: This is how it would look like with smaller buttons on the top right, a bigger margin for the side panel, and Map/Satellite buttons as they were before:
 
image

image

It covers the zoom out button, but I can make the side panel margin bigger and fix that if we want. What do you think?

minotogna: Leave the dropdown , didn't realize that .
And yes , make space for the buttons about the panel , that's what I meant

minotogna: We could even remove the zoom buttons. Mouse wheels do the job nowadays . What do you think ?

yaguzmang: I agree the zoom buttons aren't necessary, so I removed them. I moved the other buttons, and added an updated video to the description. 
I ended up not using the dropdown for the map type because the size change doesn't apply to it and it would look off:
image

minotogna (approved):

PR #3619, opened 2024-03-14
3496 - Geo UI Refactor - Statistics UI

yaguzmang:
Adding the statistics sections to the side panel, reducing the padding of the cells in the tables, and deleting the old statistics UI components.

https://github.com/openforis/fra-platform/assets/41337901/a55dbd07-2816-4fab-bb67-72aeee01cc9a

To do: 
- Add scrolling for NavGeo.
- Add styles and translation to the error message if the map fails to load.


src/client/pages/Geo/GeoMapWrapper/geoMapWrapper.scss
```diff
@@ -1,33 +1 @@
 @import 'src/client/style/partials';
```
yaguzmang, line 1: This file wasn't deleted because I will use it in an upcoming PR to add styles to the error message when the map fails to load. I tried deleting it and renaming it with the capital letter -> GeoMapWrapper.scss, but this would cause the same issue we had before with "Fileusages.scss".

src/client/components/ButtonTableExport/ButtonTableExport.tsx
```diff
@@ -25,6 +29,21 @@ const ButtonTableExport: React.FC<Props> = (props) => {
 
   if (print) return null
 
+  if (!Objects.isEmpty(csvData)) {
```
minotogna: Is this needed for geo part ? Can u use a grid instead and use the button grid export ?
If not , let's add a custom csv button export only for geo and under geo .
We'll need to remove those tables in the future .

yaguzmang: Yes, we need it for geo. There is one table where we need to export a csv file that's slightly different from what's on the screen:
image

The exported csv has an extra column "Land area", and the unit of Forest area is only in the header:
image

yaguzmang: Added a button under geo to export the custom CSV, and updated the table component to use DataGrid instead. The video is now updated 👍

minotogna: Nice , thank you !

minotogna:
Can the csv download button be on top-left corner of the table ? Now i see it in the bottom right corner

src/client/pages/Geo/ButtonCSVExport/ButtonCSVExport.tsx
```diff
@@ -0,0 +1,34 @@
...
+  csvData: CSVData
+  filename?: string
+}
+
+const ButtonCSVExport: React.FC<Props> = (props) => {
+  const { csvData, filename } = props
+
+  return (
+    <CSVLink
+      className="data-grid__btn-export btn-xs btn-primary no-print"
```
minotogna: Use the useButtonClassName hook . See other csv download button (for example in data table )

yaguzmang: Done. I added it and noticed that the style didn't match the ButtonGridExport style, so I checked ButtonGridExport and realized it did not match the style of ButtonTableExport.

For example here, ButtonGridExport on the top, ButtonTableExport on the bottom.
Screenshot 2024-03-15 at 11 17 04 AM

I updated ButtonGridExport to use the same logic and styles introduced to ButtonTableExport in the `Files and links per Cycle` feature:

image

And now the export button for Geo matches those previous two (on the top left now):
image

minotogna (approved):

PR #3646, opened 2024-03-25
3631 - Migrate Pan-European files / enable repository: migrate files

sorja:


minotogna (commented):
Should be update the links in data sources as well (if applicable)?

sorja:
> Should be update the links in data sources as well (if applicable)?

Not applicable (also double checked)!
- PanEuropean has no ODPs
- Only way would be applicable if uploaded in PanEuropean from EditorWYSIWYG and content pasted to FRA - luckily no cases like this!

minotogna:
> > Should be update the links in data sources as well (if applicable)?
> 
> Not applicable (also double checked)!
> 
> * PanEuropean has no ODPs
> * Only way would be applicable if uploaded in PanEuropean from EditorWYSIWYG and content pasted to FRA - luckily no cases like this!

what about repository links in data sources reference field ?

sorja:
> > > Should be update the links in data sources as well (if applicable)?
> > 
> > 
> > Not applicable (also double checked)!
> > 
> > * PanEuropean has no ODPs
> > * Only way would be applicable if uploaded in PanEuropean from EditorWYSIWYG and content pasted to FRA - luckily no cases like this!
> 
> what about repository links in data sources reference field ?

Do you mean this one?

https://github.com/openforis/fra-platform/pull/3646/files#diff-4a617ebac28965a554557bb905361c1ea3d6c7dacfa0d835a3e2d034516d95d7R38-R39


There is single case for pan european but we handle it generally:
Country A had uploaded files through data sources upload file function (repository modal, editotorwysiglinks)
They were using old api

This is fixed

minotogna:
> > > > Should be update the links in data sources as well (if applicable)?
> > > 
> > > 
> > > Not applicable (also double checked)!
> > > 
> > > * PanEuropean has no ODPs
> > > * Only way would be applicable if uploaded in PanEuropean from EditorWYSIWYG and content pasted to FRA - luckily no cases like this!
> > 
> > 
> > what about repository links in data sources reference field ?
> 
> Do you mean this one?
> 
> https://github.com/openforis/fra-platform/pull/3646/files#diff-4a617ebac28965a554557bb905361c1ea3d6c7dacfa0d835a3e2d034516d95d7R38-R39
> 
> There is single case for pan european but we handle it generally: Country A had uploaded files through data sources upload file function (repository modal, editotorwysiglinks) They were using old api
> 
> This is fixed

yes @sorja thanks !

minotogna (approved):

PR #3649, opened 2024-03-25
3270 - Print - Cache pdf

yaguzmang:
Persisting the pdfs in the database so we don't have to generate them each time. 
If the country has had any changes made since the last pdf was generated, then a new one will be built and saved. 
Also, added a flag `forceRefresh` to generate and save the file, regardless if it already exists and is valid.

Generating for the first time and getting an already generated file:

https://github.com/openforis/fra-platform/assets/41337901/642b506a-9ee2-47e9-bb7b-f4318c562e2f

Simulating that changes were made (using `country_summary` view, `last_update` field):
Note: I had to refresh the country summary view manually for the field to be updated. How does this work in prod? 

https://github.com/openforis/fra-platform/assets/41337901/a055097f-7a9a-4822-92b0-0c3c85025730

Passing the for `forceRefresh` flag:

https://github.com/openforis/fra-platform/assets/41337901/3ad16186-8ce1-4986-94ae-b73f2526e293

(UPDATE) Language and only tables support:

https://github.com/openforis/fra-platform/assets/41337901/e7b82ff4-2672-40d3-96d7-873c62da6a64


src/server/api/cycleData/print/report.ts
```diff
@@ -39,13 +43,46 @@ const getPdf = async (req: Request): Promise<Buffer> => {
...
+  const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName })
+
+  const [cachedPdfInfo, countryCycleLastUpdate] = await Promise.all([
+    PdfRepository.getOne({ assessment, countryIso, cycle, fileName }),
+    CycleDataController.getLastUpdate({ assessment, countryIso, cycle }),
+  ])
+
+  if (Objects.isEmpty(cachedPdfInfo)) {
+    const pdfBuffer = await buildPdf(req)
+    await PdfRepository.create({ assessment, buffer: pdfBuffer, countryIso, cycle, fileName })
```
minotogna: I think in this case we need a controller, but it's just a primarily renaming issue.
And It must be accessed from `CycleDataController` as property `Report`.
e.g.
```
CycleDataController.Report.create
```

would that work ?

yaguzmang: Yes. Fixed it, thanks!

src/server/api/cycleData/print/report.ts
```diff
@@ -39,13 +42,46 @@ const getPdf = async (req: Request): Promise<Buffer> => {
...
+}
+
 export const report = async (req: Request, res: Response) => {
   try {
     const { assessmentName, cycleName, countryIso } = req.query
 
-    const pdf = await getPdf(req)
     const fileName = `${assessmentName}_${cycleName}_${countryIso}.pdf`
 
+    const pdf = await getPdf(req, fileName)
```
minotogna, line 87: forgot to mention, file name must include also onlyTables. basically each country has two variations of the pdf. see ui -> 
image
the button to the right opens the onlyTables option , that users can download as well. 

ps update your branch, pushed the dev merge

src/server/controller/cycleData/report/updateFile.ts
```diff
@@ -0,0 +1,50 @@
...
+    const newFile = await FileRepository.create({ file: pdfMulterFile }, t)
+
+    const updateRepositoryItemProps: RepositoryItem = {
+      ...repositoryItem,
+      fileUuid: newFile.uuid,
+    }
+    const updateRepositoryProps = { assessment, cycle, repositoryItem: updateRepositoryItemProps }
+    const target = await RepositoryRepository.update(updateRepositoryProps, t)
+
+    await FileRepository.removeOne({ uuid: fileToRemoveUuid }, t)
```
minotogna: we might actually not need to remove the file. there's a job scheduled periodically that cleans up the file table. @sorja cc

src/server/repository/assessmentCycle/repository/getOne.ts
```diff
@@ -8,14 +9,20 @@ import { BaseProtocol, DB, Schemas } from 'server/db'
...
 export const getOne = async (props: Props, client: BaseProtocol = DB): Promise<RepositoryItem> => {
-  const { assessment, cycle } = props
+  const { assessment, countryIso, cycle } = props
   const schemaCycle = Schemas.getNameCycle(assessment, cycle)
 
   const values = 'uuid' in props ? [props.uuid] : [props.fileName]
-  const where = 'uuid' in props ? 'uuid = $1' : "props -> 'translation' ->> 'en' = $1"
+  let where = 'uuid' in props ? 'uuid = $1' : "props -> 'translation' ->> 'en' = $1"
+
+  if (!Objects.isEmpty(countryIso)) {
```
minotogna: can u 100% be sure we really need to pass the countryIso here? It might not be needed if the fileName is unique, and it is I assume

 .

src/server/repository/assessmentCycle/countrySummary/index.ts
```diff
@@ -10,6 +11,7 @@ export const CountrySummaryRepository = {
   dropMaterializedView,
   getCount,
   getLastEditOdpData,
+  getLastUpdate,
```
minotogna: I think getLastEditOdpData and getLastUpdate could be replaced by introducing getOne that returns CountrySummary object. then the callers will need to be updated by using the needed props. what do u think? would it be a big change ?

src/server/api/cycleData/print/report.ts
minotogna: there's a third variant to consider: The language. report can be generated in 6 languages.
the number of unique files we can have is 2 * 6

yaguzmang: Will add these. 
Since not every report request will have a user from which I could get the language, I will add it as a parameter.
The filename will be `${assessmentName}_${cycleName}_${countryIso}_${lang}${<_tables or empty>}.pdf`

src/server/repository/assessmentCycle/repository/getOne.ts
yaguzmang: It should be unique, yes, I will remove the param. But good to note, there isn't any database level unique check so two files can't be in the same country with the same name. 

One potential issue we could have is if a user adds a pdf file to the country repository with the same name as the one we generate for the reports (e.g `fra_2020_CAF.pdf`), then the `client.one` function in the repository will throw an error (multiple rows match), and generating a report would fail. 
Should we add any logic to handle this? Since the file name is all we have to query an existing report, a solution I see is adding a new param to the repository props like 'report: true', and then we could also use that to make the query more specific.

src/server/api/cycleData/print/report.ts
minotogna: yes, get it from current i18n.language

src/server/repository/assessmentCycle/repository/getOne.ts
minotogna: we shouldn't simply add two file with same name for now. ok?

src/server/repository/assessmentCycle/countrySummary/index.ts
yaguzmang: I added `getOneOrNone` since client.oneOrNone is what getLastEditOdpData and getLastUpdate were already using.

src/server/controller/cycleData/report/updateFile.ts
minotogna: I meant this invocation as well as the method can be removed . What do u think ?

minotogna (approved):

PR #3668, opened 2024-03-28
3654 - History

### LINKED ISSUE #3654: Data History: Section -> Data Sources
Access rules:
- Only user Editors can see (eg reviewers, etc, edit access required)
- NC doesn't see it in review

Tasks:

- [x] Add primary button next to data source edit `History`
  - [x] Enabled only when UI Lock is open and Data Sources Edit is closed
  - [x] And vice versa
- [x] On click History: Dispatch action : `HistoryActions.toggle` with (sectionKey, sectionLabelKey)
- [x] When adding history item fetch table data (`history/?sectionName,type=dataSources`)
- [x] When history is active (if items is not empty)
  - [x] Replace navigation content with history panel
  - [x] Lock UI, and when history is not active, restore data lock state (check show ndp button 1a)
- [x] History Panel
  - [x] For each section show sectionLabel + date + entries (see recent activity ui)
  - [x] On click entry update data accordingly
- [x] When history becomes inactive, restore original data state in redux store
- [x] EndPoint returns history.push(currentSitatution) (history+current) and on inactivate history pop current to redux state
- [x] Use `public.activity_log`
- [x] Rename Button type to black / Use key in store without sectionName
- [x] Use Table Paginated
- [x] Show text diffs in UI + missing/added rows + highlight button on the left of current diff 
- [ ] Enable History for all Comments fields 

- Investigate redux history plugin

```typescript
// under store: cycleData.history

type HistoryItemSectionKey = string

// meta?
type HistoryItemState<HistoryItemStateDatum> = {
  sectionLabelKey: string // example: title.headers.dataSource
  sectionKey: HistoryItemSectionKey // example: nationalData-dataSources
}

// redux store state
type HistoryState {
  items: Record<HistoryItemSectionKey, HistoryItemState>
}

```

sorja:
- [x] on unmount, reset history state (eg. activate history mode, select home icon, select country, history is deactive)
- [x] Fix malformed history items ( eg. reference: text sometimes => reference.text: text )

-- 

- [ ] a record of sectionName-> record target array history items  ?


src/server/repository/public/activityLog/getMany.ts
```diff
@@ -0,0 +1,37 @@
...
+            jsonb_build_object('id', u.id, 'props', u.props) as user
+     from public.activity_log al
+              join public.users u on al.user_id = u.id
+     where assessment_uuid = $1
+       and cycle_uuid = $2
+       and country_iso = $3
+       and section = $4
+       and message = $5
+       and target ->> 'name' = $6
+     limit $7 offset $8`,
```
sorja, line 34: sort by time @minotogna

minotogna (approved):

Issue #3496, opened 2024-02-12
GEO: UI Restyling


minotogna:
https://github.com/openforis/fra-platform/assets/1138156/eb3729fe-ad71-41a4-9387-4687dabdc7ff

https://github.com/openforis/fra-platform/assets/1138156/6189cedf-803d-4bef-8f03-12e04da3031b

----------------------------------------------

TODO:
- [x] FIx: Disable selection of GEE Asset it  layer (@minotogna )
- [x] Toolbar Map/Satellite buttons (@minotogna )
- [x] Toolbar Center/zooms buttons (@minotogna )
- [ ] Statistics: Improve chart legend readability (@yaguzmang )
- [x] Fix layout of geo beta message (@minotogna )
- [ ] in area selector: show only countries (@yaguzmang ) 
- [ ] Translations 

----------------------------------------------
- [x] Render empty navigation menu (hide current geo menu)
- [x] Backgrounds
    - [x] Add GeoSections component for the navigation
    - [x] Update redux state type for mosaic
    - [x] Add wire frame ui
    - [x] Remove old Mosaic component
    - [x] Add UI styles
- [x] Forest layers
    - [x] Add LayersSection component that renders layers based on a config object
        - [x] Add Global Opacity
        - [x] Add Recipe Selector
        - [x] Add LayerOpacityControl
        - [x] Add CustomAssetControl
        - [x] Add AgreementLevelControl
        - [x] Add TreeCoverPercentageControl
        - [x] Add YearControl 
- [x] Add translation keys to all sections and layers
- [x] Protected area layers
- [x] Burned area layers
- [x] Statistics
    - [x] Add translation keys to all panels and graphs.
    - [x] Add collapsable section component.
    - [x] Make side panel scrollable.
    - [x] Render statistics.
- [x] Google maps buttons
- [x] Add scrolling for NavGeo (see NavAssessment > hooks > useMaxHeight.ts)
- [x] Add styles and translation to the error message if the map fails to load.
- [x] Fix translations mentioned in review doc.

Left menu:

- [x] Same exact style as Navigation 
- [x] Menu exandable sections: Satellite mosaic, Forest Layers, Protected Area Layers, Burned Area Layers
- [x] Forest Layers->Recipes: Use common Select component 

Statistics Panel:
- [x] Move statistics Panel to the right ( 40/45% window width ) , height 32 px below toolbar and 32 px above page bottom margin.

Map --- IF possible:
[...]

PR #3675, opened 2024-03-29
3496 GEO UI: map satellite buttons

minotogna:
https://github.com/openforis/fra-platform/assets/1138156/05243aaf-ef8b-49b0-8fa1-a51f24f6b755


yaguzmang (commented):
I noticed that in a slightly smaller screen size, the Beta message overlaps the new github issues button and the desk study text. Should we place the message in between the buttons on the left and right?
 Also, when the user clicks on a Area that isn't a country, we should navigate them out of geo. What do you think?
 
https://github.com/openforis/fra-platform/assets/41337901/40909423-0aa3-4b77-8918-ef6aa903a59e

minotogna:
> I noticed that in a slightly smaller screen size, the Beta message overlaps the new github issues button and the desk study text. Should we place the message in between the buttons on the left and right? Also, when the user clicks on a Area that isn't a country, we should navigate them out of geo. What do you think?
> 
>  2024-03-29.08-43-20.mp4

I think we need to : 

- [ ] Fix layout of geo beta message (I will take care of this)
- [ ] in area selector: show only countries (you can take care of this please?)

I have added those two points in the main issue, below video see TODO: https://github.com/openforis/fra-platform/issues/3496#issue-2130175846
We'll work on the main issue items in separate PRs.

what do u think?

yaguzmang (approved):

PR #3679, opened 2024-03-29
3648 - Fix cursor changing position when pasting HTML

### LINKED ISSUE #3648: Copy and paste issue
@minotogna 

**Describe the bug**
Colleagues have been troubled by this issue for a very long time: copying and pasting tables or text into the text box has failed multiple times to place it in the right position.

I know there was an investigation done in the past. Could you please keep an eye on it and see if there is a new solution? At the very least, we need to revise it before(for) FRA 2030.

yaguzmang:
Preventing the `onChange` callback to run when `onBlur` is triggered by the Jodit Editor HTML processor popup. That fixes the cursor changing position, and keeps the logic that sanitizes the content before sending it to the BE.

https://github.com/openforis/fra-platform/assets/41337901/eb8f8f6a-69ee-4f20-81ab-35f95c541084


minotogna (approved):
👍

PR #3680, opened 2024-03-29
3496 Geo - Add center zoom buttons / fix beta message layout

minotogna:
https://github.com/openforis/fra-platform/assets/1138156/ac3940bd-b439-4393-937f-e257a7574ed2


src/client/components/PageLayout/Toolbar/Options/GeoOptions/Zoom/Zoom.tsx
```diff
@@ -1,15 +1,30 @@
...
 import Button, { ButtonSize, ButtonType } from 'client/components/Buttons/Button'
+import { mapController } from 'client/utils'
 
 const Zoom: React.FC = () => {
+  const dispatch = useAppDispatch()
+  const { minZoom, maxZoom } = useGeoMapOptions()
+
+  const setZoom = useCallback(
+    (_zoom: number) => {
+      if (_zoom >= minZoom && _zoom <= maxZoom) {
```
yaguzmang, line 14: Is it possible that we can have a situation like:
minZoom: 3
current zoom in map: 3.5
Then when lowering the zoom, it would be `setZoom(mapController.getMap().getZoom() - 1)` -> `setZoom(3.5 - 1)`
Since the value is lower than the min zoom, we wouldn't update it. 

I think in that case we should set the zoom to the min. Same goes for the max zoom.
What do you think?

minotogna: I think zoom is integer

yaguzmang (approved):

PR #3685, opened 2024-04-02
3496 - Geo UI Refactor - Statistics Table source name color

yaguzmang:
NOTE: This branch is based on the current `3496-geo-ui-refactor-graph-updates` branch since the small refactor of the table data type is necessary to color the cells.

image

### DIFF
--- a/src/client/components/Navigation/NavGeo/Layer/ToggleControl/ToggleControl.scss
+++ b/src/client/components/Navigation/NavGeo/Layer/ToggleControl/ToggleControl.scss
@@ -30,6 +30,16 @@
       color: white;
     }
   }
+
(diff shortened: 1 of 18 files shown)


src/client/components/Navigation/NavGeo/Layer/ToggleControl/ToggleControl.scss
```diff
@@ -30,6 +30,16 @@
       color: white;
     }
   }
+
+  &.legend {
```
yaguzmang, line 34: `div.legend` didn't work here. An alternative would be `&.data-cell.legend`, or simply `&.legend` as it is now.

PR #3687, opened 2024-04-02
3507 - Fix mojibake in status notification

### LINKED ISSUE #3507: Mojibake in notification title
@minotogna 

Mojibake in notification title
image

yaguzmang:
Fixing a mojibake in the status notification email.

The root reason is that during i18next interpolation, the single quotes are escaped by default prevent mitigate XSS attacks. This can be toggled off by adding a dash : `{{- country}}`.  
https://www.i18next.com/translation-function/interpolation#unescape

An alternative to fix all instances would be to toggle off escape globally during the i18next initialization.

Fixed text:
image

### DIFF
--- a/src/i18n/resources/en.js
+++ b/src/i18n/resources/en.js
@@ -871,10 +871,10 @@ The FRA team fra@fao.org
     panEuropean: 'pan-European',
     deskStudy: 'Desk study',
     statusChangeNotification: {
-      subject: '{{country}} status changed to {{status}} on FRA Platform',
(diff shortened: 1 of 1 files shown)


sorja:
Could you be so kind as to check the `emails.js` that it has all country names escaped with a dash `-`?

minotogna (approved):

minotogna:
> Could you be so kind as to check the `emails.js` that it has all country names escaped with a dash `-`?

@sorja sorry I approved , didn't realize about your comment 😓

PR #3695, opened 2024-04-04
3654 - History: On closing history browser - reset to original state

sorja:
Todo:
- Investigate a better way to implement this.
- Investigate if we can always expect the data head to be the latest state? Probably not in case we have manipulated data with eg. migrations? @minotogna 

https://github.com/openforis/fra-platform/assets/5508251/ff695b33-b21e-4075-be7a-1ae645c2a6a9

### DIFF
--- a/src/client/components/Navigation/NavAssessment/History/Items/hooks/useResetState.ts
+++ b/src/client/components/Navigation/NavAssessment/History/Items/hooks/useResetState.ts
@@ -1,0 +1,29 @@
+import { useEffect } from 'react'
+
+import { Histories, HistoryItemSectionKey } from 'meta/cycleData'
+
+import { useAppDispatch } from 'client/store'
+import { DataActions } from 'client/store/data'
+import { useSectionRouteParams } from 'client/hooks/useRouteParams'
+import { useData } from 'client/components/Navigation/NavAssessment/History/hooks/useData'
+import { getStatePath } from 'client/components/Navigation/NavAssessment/History/Items/utils/getStatePath'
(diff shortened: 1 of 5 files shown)


src/client/components/Navigation/NavAssessment/History/Items/hooks/useResetState.ts
```diff
@@ -0,0 +1,29 @@
...
+export const useResetState = (sectionKey: HistoryItemSectionKey): void => {
+  const { assessmentName, cycleName, countryIso, sectionName } = useSectionRouteParams()
+  const dispatch = useAppDispatch()
+  const data = useData(sectionKey)
+
+  useEffect(() => {
+    return () => {
+      if (data) {
+        const { name } = Histories.getHistoryItemKeyParts(sectionKey)
+        const value = getTargetValue[name](data?.at(0))
```
sorja, line 23: Current implementation expects the `head` of `data`-arr to be the latest and equal to state.

I will investigate better way of implementing this, eg. using `useRef` or similar?

minotogna:
> Todo:
> 
> - Investigate a better way to implement this.
> 
> - Investigate if we can always expect the data head to be the latest state? Probably not in case we have manipulated data with eg. migrations? @minotogna 
> 
> 
> 
> https://github.com/openforis/fra-platform/assets/5508251/ff695b33-b21e-4075-be7a-1ae645c2a6a9
> 
> 
> 
> 
> 
> 
> 
> 

@sorja I think it's ok for now . 

Also , I think we should enable the history if: user is reviewer or admin and , obviously, he/she can edit the data .
What do you think?

minotogna (approved):

PR #3702, opened 2024-04-06
3666 - Editor WYSIWYG -  target blank new page and link validation

yaguzmang:
Adding target "_blank" and rel "nofollow" by default to links added using the editor WYSIWYG (repository links have target = "_blank" by default already).

I spent a bit of time looking for the editor options to do this. For future reference, the official Jodit editor documentation site does not have all the available options documented. However, I have found that the [Jodit editor repo](https://github.com/jodit/jodit-react](https://github.com/xdan/jodit/)) is a good source to find examples. Those can be usually found in the test.js files. For instance, the options for this PR I found [here](https://github.com/xdan/jodit/blob/main/src/plugins/link/link.test.js#L451-L463).

https://github.com/openforis/fra-platform/assets/41337901/d83fd2d2-6963-4576-ada0-da37b659d022

Link validation:

https://github.com/openforis/fra-platform/assets/41337901/ba1821a8-73c9-4be9-8914-7c8fcecd11e3

TODO: Add translations for 'Invalid link' when received.


src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/hook/useValidationError.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const useValidationError = (props: Props): Returned => {
+  const { dataSource } = props
+  const { t } = useTranslation()
+  const editor = useCanEditCycleData()
+
+  const validationError = useMemo<Returned>(() => {
+    if (!dataSource.placeholder && Objects.isEmpty(dataSource.reference)) return t('generalValidation.notEmpty')
+
+    if (editor && !validateLinks(dataSource.reference)) return t('generalValidation.invalidLink')
```
yaguzmang: Currently, the editor WYSIWYG component is agnostic of validations and error states, so I decided to add this logic here instead (to be added to other places if we decide to continue this way). Would we want to keep all validations as a responsibility of the caller components or should we introduce validation logic to the editor itself?

minotogna: @yaguzmang please , keep in mind this invalid link validation should be applied in data sources as well as everywhere we use the Editor , maybe we should add a common EditorValidators ? @sorja be in the loop

yaguzmang: Sure, I keep that in mind. We will keep this validation as responsibility of the EditorWYSIWYG component callers then. 
Another option could be to do something like:
```
<EditorWYSIWYG
     disabled={disabled}
     onChange={onChange}
     ...
     validators={[EditorValidators.links, EditorValidators.otherValidator]}
     onValidationErrors={setErrors}
 >
 ```
 There are pros and cons to both approaches. What do you think?  
 cc @sorja

Comment (minotogna) [reply src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/hook/useValidationError.ts]: I think if all the Editor will use the same validator, there's no reason to pass the same validator over and over again from outside. the Validation will be triggered conditionally. i.e. if user has edit access to that resource. 
what do u guys think @sorja @yaguzmang ?

Comment (yaguzmang) [reply src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/hook/useValidationError.ts]: Since all we need to do now is show a tooltip, I agree. If we needed to do something else in case of error, I would say better keep it to the caller, but that's not the case. 👍

**Comment by minotogna** [comment]
Looks good to me 👍🏾. @sorja your review

File: src/i18n/resources/en/generalValidation.js [hunk 2]
```diff
@@ -8,6 +8,7 @@ module.exports = {
   forestAreaExceedsExtentOfForest: `Exceeds Forest area (1a) ({{value}})`,
   forestAreaReportedIsDifferentFromPreviousCycle: `Reported area ({{forestArea2025}}) differs from previously reported ({{forestArea2020}}). Please provide a comment explaining why it is different.`,
   forestSumAreaExceedsExtentOfForest: `The sum of the reported FRA categories exceeds the forest area reported in table 1a`,
+  invalidLink: 'Invalid link',
```
minotogna, line 11: We need translations @yaguzmang

sorja (approved):

PR #3723, opened 2024-04-11
3519 - Fix app crashing when switching between countries for FRA2020

### LINKED ISSUE #3519: Bug appears when switching between countries for FRA2020
@minotogna 

If you have time to investigate:

**Describe the bug**
There is an error showed up when switching between countries for FRA2020

**To Reproduce** (I cannot reproduce with my laptop)
Reported by Erica, may be linked with the local environment. The browser version has the issue:
Chrome is up to date
Version 121.0.6167.185 (Official Build) (64-bit)

image

**Priority**
Very low priority

yaguzmang:
The app crashes for some users when opening the country select. To reproduce the issue, log in to FRA 2020 with a user that has access to an atlantis country ISO for that cycle:

https://github.com/openforis/fra-platform/assets/41337901/6ee71e4f-7d87-425f-9f65-9ba3fc4ddbb5

After the fix:

With FRA 2020 atlantis enabled: 

https://github.com/openforis/fra-platform/assets/41337901/af1481cb-cd48-459e-a9b8-7d7fd8732977

With FRA 2020 atlantis disabled: 

https://github.com/openforis/fra-platform/assets/41337901/e8542614-6194-482c-864c-50fa6eda937d

### DIFF
--- a/src/client/components/AreaSelector/CountryList/CountryListRow/CountryListRow.tsx
+++ b/src/client/components/AreaSelector/CountryList/CountryListRow/CountryListRow.tsx
@@ -1,6 +1,7 @@
 import React, { useCallback, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 
+import { Objects } from '@openforis/arena-core'
 import classNames from 'classnames'
 import { Dates } from 'utils/dates'
 
@@ -49,16 +50,16 @@ const CountryListRow: React.FC<Props> = (props: Props) => {
 
   return (
     <div
+      aria-hidden="true"
       className={classNames('country-selection-list__row', role, { expanded, selected })}
       onClick={(e) => {
         e.preventDefault()
         onElementSelect(countryIso)
       }}
-      aria-hidden="true"
     >
       <div ref={countryNameRef}>{i18n.t<string>(Areas.getTranslationKey(countryIso))}</div>
 
-      {hasRole && (
+      {hasRole && !Objects.isEmpty(country) && (
         <>
           <div>
             <CountryStatusIndicator status={status} />
--- a/src/client/components/AreaSelector/CountryList/CountryListRow/CountryListRow.tsx
+++ b/src/client/components/AreaSelector/CountryList/CountryListRow/CountryListRow.tsx
@@ -1,9 +1,9 @@
 import React, { useCallback, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { Objects } from '@openforis/arena-core'
 import classNames from 'classnames'
 import { Dates } from 'utils/dates'
+import { Objects } from 'utils/objects'
 
 import { Areas, CountryIso, Global, RegionCode } from 'meta/area'
 import { UserRoles } from 'meta/user/userRoles'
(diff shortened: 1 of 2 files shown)


src/client/components/AreaSelector/CountryList/CountryListRow/CountryListRow.tsx
```diff
@@ -49,16 +50,16 @@ const CountryListRow: React.FC<Props> = (props: Props) => {
...
       onClick={(e) => {
         e.preventDefault()
         onElementSelect(countryIso)
       }}
-      aria-hidden="true"
     >
       <div ref={countryNameRef}>{i18n.t<string>(Areas.getTranslationKey(countryIso))}</div>
 
-      {hasRole && (
+      {hasRole && !Objects.isEmpty(country) && (
```
yaguzmang, line 62: After looking into how we fetch the areas, we check [if atlantis is allowed](https://github.com/openforis/fra-platform/blob/development/src/server/repository/assessmentCycle/country/getMany.ts#L17-L20) for the assessment/cycle. Since atlantis is not allowed for FRA 2020 and the user has a role in it,  the `useCountry` hooks returns undefined and later the app crashes when the properties are accessed. 
Is this actually a data issue? that we have user roles in atlantis FRA 2020 when atlantis is not allowed for it. If that is the case then the fix should instead be cleaning the user roles.

minotogna: @yaguzmang Good catch 👍🏾
   The best fix would actually be when fetching user (with roles), the roles array is populated only with allowed countries per cycle .
Can you have a look at this ?

yaguzmang: Added it :+1: 
And I'm keeping this `!Objects.isEmpty(country)` here because if you switch cycles fast enough or the network is slow, the same issue will be present while the new countries info is fetched.

src/server/repository/public/user/getOne.ts
```diff
@@ -51,12 +53,39 @@ export const getOne = async (props: Props, client: BaseProtocol = DB): Promise<U
...
+
   return client
     .oneOrNone<User>(
       `
         select ${selectFields}, jsonb_agg(to_jsonb(ur.*)) as roles
         from public.users u
-        left join users_role ur on u.id = ur.user_id and (ur.accepted_at is not null or ur.invited_at is null or ur.role = 'ADMINISTRATOR') ${join}
+        left join ${
+          usersRoleSubquery.length > 0 ? usersRoleSubquery : 'users_role'
+        } ur on u.id = ur.user_id and (ur.accepted_at is not null or ur.invited_at is null or ur.role = 'ADMINISTRATOR') ${join}
```
minotogna, line 88: Looks good to me 🥶😳😀
@sorja want to take a look ?

src/client/components/AreaSelector/CountryList/CountryListRow/CountryListRow.tsx
minotogna: Sure 👍🏾 sounds safer

src/server/repository/public/user/getOne.ts
sorja: looks good, can be approved

minotogna (approved):

PR #3732, opened 2024-04-15
3654 - History: Add state history compare item + button ui

minotogna:


src/client/store/data/reducers/resetHistory.ts
```diff
@@ -1,10 +1,9 @@
 import { Draft } from '@reduxjs/toolkit'
-import { Objects } from 'utils/objects'
 
-import { DataState } from 'client/store/data/stateType'
+import { DataState } from 'client/store/data/state'
 
 export const resetHistory = (state: Draft<DataState>) => {
-  Objects.unset(state.history, ['items'])
+  state.history = {}
```
sorja, line 6: is this preferred over unset?

src/client/store/data/reducers/toggleCompareHistoryItem.ts
```diff
@@ -0,0 +1,23 @@
...
+
+type Action = PayloadAction<{
+  datum: ActivityLog<never>
+  target: HistoryTarget
+}>
+
+export const toggleCompareHistoryItem = (state: Draft<DataState>, action: Action) => {
+  const { datum, target } = action.payload
+
+  if (state.history.compareItem?.[target]?.id === datum.id) {
```
sorja, line 17: better uuid ?

src/client/store/data/reducers/resetHistory.ts
minotogna: It depends . Here I just need to create an empty object and this way is certainly more performant than unsetting 2 objects

src/client/store/data/reducers/toggleCompareHistoryItem.ts
minotogna: Ok . Next pr ?

minotogna: Do we have uuid in activity log ?

sorja: If we have, check please
If not, should we migrate ?

either way, next pr is fine

sorja (approved):

src/client/store/data/reducers/toggleCompareHistoryItem.ts
minotogna: I don't know if we need to migrate that . These activities are inserted very frequently and having a uuid for those might make the insert slower. Let's talk tomorrow

PR #3733, opened 2024-04-16
3645 - Dashboard: Primary Designated Management Objective

sorja:
image

### DIFF
--- a/src/client/components/Dashboard/Dashboard.scss
+++ b/src/client/components/Dashboard/Dashboard.scss
@@ -1,0 +1,15 @@
+@import 'src/client/style/partials';
+
+.dashboard {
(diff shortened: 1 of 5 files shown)


src/client/components/Dashboard/Dashboard.scss
```diff
@@ -0,0 +1,15 @@
+@import 'src/client/style/partials';
+
+.dashboard {
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  grid-gap: $spacing-m;
```
minotogna, line 6: Better to also check previous style that included responsive breakpoints styles

minotogna (approved):

PR #3737, opened 2024-04-16
3370 - Select Component - Add toggleAll prop

### LINKED ISSUE #3370: National data sources: select all
ÖJ:

For Table 2b when adding reference to data sources (see below) users should be able to “select all” variables

Image

yaguzmang:
Adding the toggleAll prop to the Select component.

Testing with national data variables, groups, and testing that the single select remains unaffected:

https://github.com/openforis/fra-platform/assets/41337901/333db055-491e-4fe2-a79f-ae0c1af037c9

(UPDATE) With separator:
image

Multi Select without toggle all:

https://github.com/openforis/fra-platform/assets/41337901/1e9eb4ea-1ae9-476d-81ed-428b32a20d3b


src/client/components/Inputs/Select/Indicators/Indicators.scss
```diff
@@ -46,3 +46,18 @@ div.select__clearIndicator {
     display: none;
   }
 }
+
+.select__toggleAllOption-checkbox {
```
yaguzmang: This is very similar to the deprecated `ButtonCheckBox`. 
The new `ButtonCheckbox`  component has a extra border for the label so it doesn't fit that well in this case:
image

Should we add another version of the checkbox similar to the previous implementation? Maybe by passing a 'variant' prop that allows to choose.

sorja:
Looks good, should we add horizontal line under 'Select all'?

minotogna:
> Looks good, should we add horizontal line under 'Select all'?

Maybe yes 👍🏾

src/client/components/Inputs/Select/Indicators/Indicators.scss
minotogna: Which style did u apply , the one in the video or this ? I like the one in the video .
Regarding generic Chekhov , let's see later on

src/client/components/Inputs/Select/Indicators/Indicators.tsx
```diff
@@ -41,3 +50,31 @@ export const ClearIndicator: React.FC<ClearIndicatorProps> = (props) => {
...
+    if (inputRef.current) {
+      inputRef.current.indeterminate = isInputIndeterminate
+    }
+  }, [isInputIndeterminate])
+
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    <components.Option {...props}>
+      <input
+        key={`${data.value}-${isInputIndeterminate}`}
```
minotogna: Is the key needed ?

src/client/components/Inputs/Select/Indicators/Indicators.tsx
```diff
@@ -41,3 +50,31 @@ export const ClearIndicator: React.FC<ClearIndicatorProps> = (props) => {
     </components.ClearIndicator>
   )
 }
+
+export const MultiSelectOption: React.FC<OptionProps<Option>> = (props) => {
```
minotogna: Maybe better to move this component in its own file . What do u think ?

src/client/components/Inputs/Select/Indicators/Indicators.scss
yaguzmang: The one in the video is the one that's applied. And okay, sounds good 👍

src/client/components/Inputs/Select/Indicators/Indicators.tsx
yaguzmang: Yes, we need it here because the `indeterminate` attribute can't be passed directly to the input. So, if that attribute changes, the input won't be re-rendered. The `key` takes care of that by adding `isInputIndeterminate` to it.

src/client/components/Inputs/Select/Select.scss
```diff
@@ -135,15 +135,19 @@ div.select__option {
...
   }
 
   &.isSelected {
     background-color: transparent;
     color: $text-body;
-    cursor: unset;
     font-weight: 600;
+
+    &:not(.toggleAll) {
+      cursor: unset;
```
minotogna, line 150: what does this do ?

src/client/components/Inputs/Select/hooks/useToggleAllConfig.ts
```diff
@@ -0,0 +1,51 @@
...
+  const { isMulti, options, toggleAll, value } = props
+  const { t } = useTranslation()
+
+  return useMemo<Returned>(() => {
+    const components: Returned['components'] = {
+      ClearIndicator,
+      DropdownIndicator,
+      IndicatorsContainer,
+      IndicatorSeparator: null,
+    }
```
minotogna: components doesn't belong here. but in the parent. 
in case, you append the option Component if returned from this hook

src/client/components/Inputs/Select/index.ts
```diff
@@ -1,2 +1,3 @@
 export { default } from './Select'
 export type { Option, OptionsGroup, OptionsOrGroups, SelectProps } from './types'
+export { selectAllOptionValue } from './types'
```
minotogna: do we need to export it outside ?

src/client/components/Inputs/Select/Select.tsx
```diff
@@ -29,21 +34,22 @@ const Select: React.FC<SelectProps> = (props) => {
...
-        option: ({ isFocused, isSelected }) => classNames('select__option', { isFocused, isSelected }),
+        option: ({ isFocused, isSelected }) => classNames('select__option', { isFocused, isSelected, toggleAll }),
         placeholder: () => `select__placeholder`,
         singleValue: () => 'select__singleValue',
         valueContainer: () => 'select__valueContainer',
       }}
       closeMenuOnSelect={!isMulti}
-      components={{ ClearIndicator, DropdownIndicator, IndicatorsContainer, IndicatorSeparator: null }}
+      components={components}
+      hideSelectedOptions={hideSelectedOptions}
```
minotogna: this should be false always

src/client/components/Inputs/Select/Select.scss
yaguzmang: For the single select we unset the cursor if the option is selected. If the select includes toggleAll, then we don't want to unset the cursor even if the option is already selected: 

https://github.com/openforis/fra-platform/assets/41337901/9726cf2d-01bf-4b15-bbc0-29ae569723a1

This targets selected elements that don't have the toggleAll class:
```
&.isSelected {
    ...
    &:not(.toggleAll) {
```

src/client/components/Inputs/Select/Select.tsx
yaguzmang: If it is always set to false, when we pass the 'isMulti' prop but not the 'toggleAll' prop, it would behave like this:

https://github.com/openforis/fra-platform/assets/41337901/b7baedda-e584-41bf-a826-88eeee2fbd7d

Instead of how it currently is:

https://github.com/openforis/fra-platform/assets/41337901/a0577ca9-3731-4dac-a416-cb36ec954e3c

Is that what we want?

minotogna: the new Multi Select component should always be used when it's multi, independently if there's toggleAll or not .

minotogna: and yes, the behavior should be as your first video

minotogna: also noticed, when partially checked , the checkbox Unselect all should be green
image

src/client/components/Inputs/Select/index.ts
yaguzmang: It is used in different places but only within the Select component. I'll export it better from a new types.ts file in the MultiSelectOption folder, what do you think?

minotogna: leave it where it is and don't export it from the root index

src/client/components/Inputs/Select/Select.tsx
yaguzmang: I've been looking into setting the color of the checkbox in the indeterminate state but it is not possible to do that by simply adding an scss attribute. I would need to unset default styles and create custom styles from scratch. Should we do that later on when we add the simple checkbox option?

minotogna: let's do this later. but what is the simple checkbox option ?

minotogna (approved):

src/client/components/Inputs/Select/Select.tsx
yaguzmang: It would be what I mentioned in the other comment about maybe adding a 'variant' prop to the checkbox. And when you pass for instance variant='simple', you would get this simple checkbox without any extra borders or other elements.

minotogna: Okkkk sure! let's later on 😎

PR #3740, opened 2024-04-17
3645 - Dashboard: PieChart

sorja:
https://github.com/openforis/fra-platform/assets/5508251/30d406d7-e75f-47bd-9f23-45f718e973a6

Note:
Next PR: 
- Refine the pie chart styling
- Translate labels


src/meta/chart/pieChart.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { ChartColor } from 'meta/chart/chart'
+
+export type PieChartData = {
+  name: string
+  value: number
+  color: string
+}
+
+type PieCell = { key: string; color: ChartColor }
```
sorja: Maybe more meaningful to change to

```suggestion

type PieCell = { name: string, color: ChartColor }
export type PieChartData = PieCell & { value: string }

```

src/client/pages/CountryHome/Overview/meta/forestAreaPercentOfLandArea/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    ],
+  },
+]
+
+export const forestAreaPercentOfLandArea = (cycle: Cycle): DashboardPieChart => ({
+  type: DashboardItemType.pieChart,
+  title: { key: 'statisticalFactsheets.forestArea.title' },
+  table: getTable({ cycle, cols: cols[cycle.name], tableId, rowMetadata, tableName }),
+  chart: {
+    dataKey: 'value',
```
sorja: I was thinking should we hardcode this to ChartPie.tsx and always expect that value to be under key `value` ?

src/client/components/Chart/ChartPie/ChartPie.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { Cell, Legend, Pie, PieChart, Tooltip } from 'recharts'
+
+import { PieChart as PieChartType, PieChartData } from 'meta/chart'
+
+type Props = {
+  data: Array<PieChartData>
+  chart: PieChartType
+}
+const ChartPie = (props: Props) => {
```
minotogna: maybe even remove the prefix Chart from this component.

src/client/pages/CountryHome/Overview/meta/forestAreaPercentOfLandArea/index.ts
minotogna: we should pass our Data format as we for tables, the cells are calculated formula. same as table.
what do u think?

src/meta/dashboard/dashboard.ts
```diff
@@ -12,5 +13,6 @@ export type DashboardItem<Type = DashboardItemType> = {
 }
 
 export type DashboardTable = DashboardItem<DashboardItemType.table> & { table: Table }
-// export type DashboardPieChart = DashboardItem<DashboardItemType.pieChart> & { chart: PieChart }
+
+export type DashboardPieChart = DashboardItem<DashboardItemType.pieChart> & { table: Table; chart: PieChart }
```
minotogna, line 17: do we need to pass the whole table or just a properties subset ?

src/client/pages/CountryHome/Overview/meta/forestAreaPercentOfLandArea/index.ts
sorja: Yes, this makes sense as it's a wrapper component

minotogna (approved):

PR #3749, opened 2024-04-19
3654 - History: Update translations + refactor/cleanup Select

minotogna:


src/client/components/Inputs/Select/toggleAllOptions.ts
```diff
@@ -0,0 +1,17 @@
+import { TFunction } from 'i18next'
+
+import { Option, ValueInput } from 'client/components/Inputs/Select/types'
+
+const VALUE = '*'
```
sorja, line 5: enum ?

sorja: or symbol?

sorja (approved):

src/client/components/Inputs/Select/toggleAllOptions.ts
minotogna: It should be of type string , it's an option value

src/client/components/Inputs/Select/hooks/useComponents.ts
```diff
@@ -0,0 +1,26 @@
...
+export const useComponents = (props: SelectProps): Returned => {
+  const { isMulti } = props
+
+  return useMemo<Returned>(() => {
+    return {
+      ClearIndicator,
+      DropdownIndicator,
+      IndicatorsContainer,
+      IndicatorSeparator: null,
+      Option: isMulti ? MultiSelectOption : null,
```
yaguzmang, line 23: This is breaking the single select. If isMulti is false, then Option becomes null. React-select then uses null when rendering Options, instead of using the default component. This would also happen if you do Option: undefined. 
Instead you need to set it with something like:
```
    if (shouldUseCustomOption) {
      components.Option = optionComponent
    }

```

yaguzmang: Fix: https://github.com/openforis/fra-platform/pull/3752

PR #3754, opened 2024-04-19
3265 - Datagrid highlight row on review opened

### LINKED ISSUE #3265: DataGrid: highlight row on review opened
The data grid , e.g. in original data point view should highlight current review open row, like we do for Table

Image

yaguzmang:
(UPDATED) Highlighting data grid rows and titles on review opened:

https://github.com/openforis/fra-platform/assets/41337901/66f5a3ef-cb24-46bd-a677-c3a19f3961d3

ODP Comment review indicator alignment:

https://github.com/openforis/fra-platform/assets/41337901/aca7d74a-c916-4a7e-b39b-9c4bd28c9ffe

Contacts:

image

TO DO:
- [x] Close topics when user is not editing
- [x] Fix outline when the cell contains a Links editor
- [x] Fix ODP Comment review indicator alignment


src/client/components/DataGrid/DataCell/DataCell.scss
```diff
@@ -38,6 +38,28 @@ $border: 1px solid $ui-border;
   &.lastRow {
     border-bottom: $border;
   }
+
+  &.highlighted {
+    &:not(.noBorder) {
```
yaguzmang: This prevents titles like this one to be highlighted: 
image


Alternatively we can pass another flag `highlightEnabled` , like `<DataRow actions={actions} highlightEnabled={false} >` and handle that logic in the component itself.

Or do we want to highlight those titles as well?

src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/YearForDataSource.tsx
```diff
@@ -37,10 +38,16 @@ const YearForDataSource: React.FC<Props> = (props: Props) => {
       data-tooltip-content={validationError ? t('generalValidation.notEmpty') : ''}
       data-tooltip-id={TooltipId.error}
       editable={!disabled}
+      highlighted={highlighted}
```
minotogna: In data sources grid components , is it necessary to pass the highlighted to the cell all the time ? Can't it be taken from grid row directly ?

yaguzmang: Those components are wrapping the DataCell, so the `highlighted` prop is passed to those wrappers but not the DataCell component itself. 
Couple of ideas to avoid defining the prop every time:
1. Pass the prop like this so it's not necessary to define it:
```
<DataCell
      className={classNames({ 'validation-error': validationError })}
      data-tooltip-id={TooltipId.error}
      ...
      {...props}
    >
```
2. Use context api in DataRow and DataCell (maybe not the best for performance)
3. In DataRow, check recursively if any child component (or child of a child) is a DataCell, and pass the props directly to it (also maybe not the best for performance).
If you can think of any other please let me know.

I think defining and passing the `highlighted` prop makes the logic easier to follow. It's like we already do with other props like `lastCol`, `lastRow` and `disabled`.

minotogna: We already do that in DataRow. Can't we wrap the cells with DataRow ? That was the idea

src/client/components/DataGrid/DataCell/DataCell.scss
minotogna: In that case the tile should be highlighted when opening the review comments

minotogna: I mean the entire row

src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/YearForDataSource.tsx
yaguzmang: We are wrapping the DataCells with DataRow, yes. The problem is the data sources grid components also wrap the DataCell and capture the added props from DataRow:
normally it would be:
```
DataRow -> adds highlighted prop
   DataCell -> gets the highlighted prop
```
but with the Data Source component also wrapping the DataCell:
```
DataRow -> adds highlighted prop
   YearForDataSource -> gets the highlighted prop
      DataCell -> renders with props passed in YearForDataSource component
```
To avoid that in-between component from capturing the props from DataRow we would need to implement one of those 3 ideas I mentioned before.

src/client/pages/Section/Descriptions/CommentableDescription/Description.scss
```diff
@@ -1,7 +1,7 @@
 @import 'src/client/style/partials';
 
 .description {
-  grid-row-gap: $spacing-xxxs;
+  grid-row-gap: $spacing-xxxxs;
```
minotogna, line 4: Why this increased gap ?

src/client/pages/Section/Descriptions/CommentableDescription/Title/Title.scss
```diff
@@ -4,6 +4,7 @@ div.description-title {
   align-items: center;
   display: flex;
   justify-content: flex-start;
+  padding: $spacing-xxxxs 0px;
```
minotogna, line 7: Why this padding ?

src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/Comments.tsx
```diff
@@ -10,20 +10,26 @@ import { useOnChange } from './hook/useOnChange'
...
 const Comments: React.FC<Props> = (props: Props) => {
-  const { dataSource, disabled, lastRow, sectionName } = props
+  const { dataSource, disabled, highlighted, lastRow, sectionName } = props
 
   const onChange = useOnChange({ sectionName, dataSource })
   const _onChange: React.ChangeEventHandler<HTMLTextAreaElement> = (event) => onChange('comments', event.target.value)
 
   return (
-    <DataCell editable={!disabled} lastCol lastRow={lastRow}>
+    <DataCell editable={!disabled} highlighted={highlighted} lastCol lastRow={lastRow}>
```
minotogna: Can't this be handled by its own DataRow directly ?

src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/YearForDataSource.tsx
minotogna: What if we return data cells here ?

Something like 😀 
<DataCell....>
  <Year...
  ...
</DataCell>


The highlighted and firstCol should be handled only from DataRow

src/client/pages/Section/Descriptions/CommentableDescription/Title/Title.scss
yaguzmang: So the Title border is like this:
image

Instead of:
image

src/client/pages/Section/Descriptions/CommentableDescription/Description.scss
yaguzmang: With the added padding to the title, I reduced this gap so the total spacing stays the same

src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/Columns/YearForDataSource.tsx
yaguzmang: Added it in the latest commit👍 
To avoid the prop drilling, the DataCells need to be direct children of DataRow, so I updated the `DataSourceRow` component accordingly.

yaguzmang: The only place left to change I see is the `ContactRow` component. That is a bit of an special case because there we use the generic component `CellNodeExt`, which returns a `DataCell` :
```
ContactRow
  DataRow 
    CellNodeExt -> gets the extra props from DataRow and does not pass them unless prop drilling
       DataCell 
         Component -> a component based  on column type (multiselect, text, or link, etc)
```
Since `CellNodeExt` is a generic component, do we want to add there the optional props `highlighted` and `firstCol`?

minotogna: Would it make sense to use a similar approach as per data sources ?

yaguzmang: Since `CellNodeExt` is only used in two places, yes. I used the same approach to keep it consistent 👍 
image

minotogna (approved):

PR #3759, opened 2024-04-23
3645 - Dashboard: Titles show correct year

sorja:
image


src/i18n/resources/en/statisticalFactsheets.js
```diff
@@ -14,31 +14,31 @@ module.exports = {
     otherOrUnknown: 'Other or unknown',
   },
   forestArea: {
-    title: 'Forest Area (1990 — 2020)',
+    title: 'Forest Area ({{startYear}} - {{endYear}})',
```
minotogna, line 17: Do we have only English translations?

minotogna (approved):

PR #3768, opened 2024-04-25
3645 - Dashboard: Only for accepted countries or published cycles

sorja:


src/client/pages/CountryHome/FraHome/hooks/useSections.tsx
```diff
@@ -19,15 +21,18 @@ type Section = {
...
+
   const cycle = useCycle()
 
   return useMemo(() => {
     const sections: Array<Section> = []
 
     if (!cycle) return null
 
-    sections.push({ name: SectionNames.Country.Home.overview, component: Overview })
+    if (cycle.published || country.props.status === AssessmentStatus.accepted)
```
minotogna, line 34: Nope. Users must have access to view data. Why this?

minotogna (changes requested):

src/client/pages/CountryHome/FraHome/hooks/useSections.tsx
sorja: Thought we agreed to show the dashboard only for above rules in UI, otherwise omit the tab?

If not, this can be closed

PR #3769, opened 2024-04-25
3645 - Dashboard: Hide tiny values in Pie chart

sorja:
https://github.com/openforis/fra-platform/assets/5508251/a8f36234-daec-49e4-93d4-95a24a42fc26


sorja:
@minotogna 

Should we instead filter out the data that is 0 from the array that is passed to the chart?

Currently we include _all_ data, but just hide the tiny values label (as they would overlap, and add no value)

minotogna (approved):

PR #3770, opened 2024-04-25
3718 - Geo Translations - Add line break for long checkbox texts

yaguzmang:
(UPDATED)

https://github.com/openforis/fra-platform/assets/41337901/1cd26d26-e6e9-48dc-8600-e1f0459fc607

Before/after
ar:
image
image

ru:

image
image

### DIFF
--- a/src/client/components/Buttons/Button/Button.scss
+++ b/src/client/components/Buttons/Button/Button.scss
@@ -8,14 +8,15 @@
   display: flex;
   font-family: $font-family-1;
   font-weight: 600;
-  line-height: 0;
   justify-content: start;
   letter-spacing: 0.25px;
   text-decoration: none;
   text-transform: uppercase;
   transition: all 0.2s;
+  white-space: normal;
 
   svg {
+    flex-shrink: 0;
     transition: all 0.2s ease;
   }
 
@@ -43,9 +44,9 @@
 // === size styles
 .button__size-xs {
   font-size: $font-xxxs;
-  height: 20px;
-  padding: 0 6px;
   gap: 3px;
+  min-height: 20px;
+  padding: 0 6px;
 
   svg {
     height: 9px;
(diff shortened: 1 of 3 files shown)


src/client/components/Navigation/NavGeo/Layer/ToggleControl/ToggleControl.scss
```diff
@@ -6,6 +6,14 @@
   button.button-checkbox {
     border-color: $color;
     color: darken($color, 15%);
+    height: unset;
+    line-height: 1.4;
+    min-height: 24px;
```
minotogna: Good fix. 

Can we spacing ? And text aligned to the left ? (Or right Arabic)?

minotogna: I'm wondering if this can be applied to Button style directly

yaguzmang: This overrides styles from the most generic `.button` class. We could add it there but we would need to check that every button doesn't get an unexpected look.

src/client/components/Buttons/Button/Button.scss
```diff
@@ -8,14 +8,15 @@
   display: flex;
   font-family: $font-family-1;
   font-weight: 600;
-  line-height: 0;
```
yaguzmang, line 11: Now that this line is deleted, the default value for line-height is `normal`.
For some reason, on my laptop screen it doesn't affect the alignment:

https://github.com/openforis/fra-platform/assets/41337901/4e2bd791-bfc3-4a3e-9560-9f21ce74ef9a

But on a big screen it behaves differently and this happens 🤔 : 

https://github.com/openforis/fra-platform/assets/41337901/ffe0f1f2-de1c-4237-bd9a-2c37aa5e17c4

It's slightly off and changing the line-height to a value here would affect the other buttons:

https://github.com/openforis/fra-platform/assets/41337901/e6ef50ed-dd48-40dd-bceb-99e104824f4b

minotogna: Line height for general button should probably stay 0. What do you think ?

minotogna (approved):

PR #3771, opened 2024-04-25
3265 - Datagrid highlight - Fix DataCell borders for rtl direction

yaguzmang:
Before:
image

After:
image


src/client/components/DataGrid/DataCell/DataCell.scss
```diff
@@ -134,9 +133,27 @@ $border: 1px solid $ui-border;
...
+
+      &.firstCol.lastCol::before {
+        box-shadow: $highlight-top-bottom, $highlight-left, $highlight-right, $highlight-border-shadow;
+        clip-path: inset(-4px -4px -4px -4px); // Adds shadow to top, bottom, left, right
+      }
+    }
   }
 
-  .data-cell.editable {
+  .data-cell.editable:not(.noBorder) {
```
yaguzmang, line 155: This is to fix this border by the review indicator, which is added unnecessarily:
image

src/client/components/DataGrid/DataCell/DataCell.scss
```diff
@@ -134,9 +133,27 @@ $border: 1px solid $ui-border;
...
+    border-left: unset;
```
yaguzmang, line 156: This fixes the duplicated borders in between cells that looked like this:
 
image

And now look like this (how we have it for non-rtl):

image

src/client/components/DataGrid/DataCell/DataCell.scss
minotogna: Was this bug introduced in the recent changes ?
Please , have you tested all cases.

src/client/components/DataGrid/DataCell/DataCell.scss
minotogna: How come we introduced these bugs ?

src/client/components/DataGrid/DataCell/DataCell.scss
yaguzmang: Yes, since this component is in a DataRow, the DataCell got the lastCol prop and the styles in the `rtl` class were not updated accordingly (overlooked by me, it passed this time me that Arabic has some different styles 😅). Will consider that when testing from now on. 

I've tested now for Arabic too, but I'll double check 👍

src/client/components/DataGrid/DataCell/DataCell.scss
yaguzmang: I hadn't checked the Arabic version that has slightly different styles that also needed to be updated. I will take that into account now when testing in general

src/client/components/DataGrid/DataCell/DataCell.scss
yaguzmang: Tested in Arabic and other languages UI styling and functionality in the mentioned sections today, and it works as expected. I couldn't think of any other special cases. If you think there might be something else please let me know, otherwise I think this is good to go 👍

minotogna (approved):

PR #3775, opened 2024-04-26
3645 - Dashboard: Support only countries

sorja:
Currently region data is fetched from value_aggregate table.

Todo:

- For 2025 data ask OJ
- For 2020 data, check how we can fetch without adding unnecessary complexity


sorja:
Next pr will support regions for FRA 2020

minotogna (approved):

PR #3780, opened 2024-04-30
3772 - Fix Navigating between data tables and GEO

### LINKED ISSUE #3772: Navigating between data tables and GEO
When navigating from a data table to GEO, it would be nice to return back to the same table instead of the dashboard, when switching between the views.

minotogna:
https://github.com/openforis/fra-platform/assets/1138156/cb7819a0-0637-4e08-8de5-14d51620a964

### DIFF
--- a/src/client/pages/Geo/Geo.tsx
+++ b/src/client/pages/Geo/Geo.tsx
@@ -3,7 +3,7 @@ import React, { Suspense } from 'react'
 import { Navigate } from 'react-router-dom'
 
 import { Routes } from 'meta/routes'
-import { Users } from 'meta/user'
+import { Authorizer } from 'meta/user'
 
 import { useCycle } from 'client/store/assessment'
 import { useUser } from 'client/store/user'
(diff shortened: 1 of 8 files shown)


src/client/pages/Geo/Geo.tsx
```diff
@@ -17,9 +17,7 @@ const Geo: React.FC = () => {
   const cycle = useCycle()
   const user = useUser()
 
-  const hasRoleInCountry = Users.hasRoleInCountry({ cycle, countryIso, user })
-
-  if (!hasRoleInCountry) {
+  if (!Authorizer.canViewGeo({ cycle, countryIso, user })) {
```
sorja: should we use hook useCanViewGeo

sorja (dismissed):

src/client/pages/Geo/Geo.tsx
minotogna: sure.

yaguzmang (approved):

PR #3786, opened 2024-05-02
3785 - Delete total land area node and union node_ext

sorja:
- [x] TODO: Resolve how to avoid showing all years for land area
- [x] TODO: When creating view, check cycle uuid for rows and tables and cols, verify

image

_Note:
Separate PR for refreshing the view
The above screenshot is generated by manually running the view update_

### DIFF
--- a/src/test/migrations/steps/20240502105432-step-delete-node-total-land-area.ts
+++ b/src/test/migrations/steps/20240502105432-step-delete-node-total-land-area.ts
@@ -1,0 +1,23 @@
+import { Promises } from 'utils/promises'
+
+import { AssessmentController } from 'server/controller/assessment'
+import { BaseProtocol, Schemas } from 'server/db'
+
+export default async (client: BaseProtocol) => {
+  const assessments = await AssessmentController.getAll({}, client)
+  await Promises.each(assessments, async (assessment) => {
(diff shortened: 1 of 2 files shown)


src/test/migrations/steps/20240502105432-step-delete-node-total-land-area.ts
```diff
@@ -0,0 +1,23 @@
...
+      await client.none(`
+          delete
+          from ${schemaCycle}.node
+          where row_uuid in (select r.uuid
+                             from ${schemaCycle}.node n
+                                      left join ${schemaAssessment}.row r on r.uuid = n.row_uuid
+                             where r.props ->> 'variableName' = 'totalLandArea')
+      `)
+    })
+  })
```
minotogna, line 22: We might need to update update data cache . What do you think ?

sorja: Sure, but I was thinking of doing it in the last migration step, as they will run anyway sequentally, this way we can avoid running it multiple times

Let me know

minotogna: Sure

minotogna (approved):

PR #3791, opened 2024-05-03
3743 - Broken Links - Add cycle link tables

yaguzmang:


src/meta/cycleData/link/link.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export type LinkLocation = {
+  column: string
+  entityProps: Record<string, string>
+  id: number
+  table: string
+}
+
+export type LinkProps = {
```
sorja, line 12: Should `LinkProps` be defined before usage at line `9`?

src/meta/cycleData/link/link.ts
```diff
@@ -0,0 +1,37 @@
...
+  readonly id: number
+  readonly uuid: string
+  countryIso: CountryIso
+  link: string
+  location: LinkLocation
+  props: LinkProps
+  target: string
+}
+
+export type LinkLocation = {
```
sorja: Should `LinkLocation` be defined before usage at line `8`?

src/meta/cycleData/link/link.ts
```diff
@@ -0,0 +1,37 @@
...
+  visit: {
+    code: LinkValidationStatusCode
+  }
+  visits: Array<{
+    code: LinkValidationStatusCode
+    timestamp: string
+  }>
+}
+
+export enum LinkValidationStatusCode {
```
sorja: As above

src/meta/cycleData/link/link.ts
yaguzmang: Reordered it 👍

src/meta/cycleData/link/link.ts
```diff
@@ -0,0 +1,37 @@
+import { CountryIso } from 'meta/area'
+
+export type LinkLocation = {
+  column: string
+  entityProps: Record<string, string>
+  id: number
+  table: string
+}
```
minotogna: I would leave only id and entityProps.
The other can be entityProps .
And I would call them tableName colName 
What do you think ?

src/meta/cycleData/link/link.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export enum LinkValidationStatusCode {
+  empty = 'empty',
+  enotfound = 'enotfound',
+  success = 'success',
+  urlParsing = 'urlParsing',
```
minotogna: UrlParsingError?

src/meta/cycleData/link/link.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export type LinkProps = {
+  approved?: boolean
+  name: string
+  visit: {
```
minotogna: I think we only need visits .
What is this visit ?
Also, I would add visits in a different column for fetching and updating reasons

src/meta/cycleData/link/link.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export type Link = {
+  readonly id: number
+  readonly uuid: string
+  countryIso: CountryIso
+  link: string
+  location: LinkLocation
+  props: LinkProps
+  target: string
```
minotogna: What if the same link is referenced with two different targets? 
@sorja @yaguzmang 
Do we really want to store this property ?

src/meta/cycleData/link/link.ts
minotogna: > I would leave only id and entityProps.
> The other can be entityProps .
> And I would call them tableName colName 
> What do you think ?

Actually , LinkLocation is itself the Record.
What do you think @yaguzmang @sorja ?

src/meta/cycleData/link/link.ts
yaguzmang: During our first discussion I think we said we could have this extra column for the latest visit so we don't need to query the visits array. I can remove it though.

src/meta/cycleData/link/link.ts
yaguzmang: Yes, that works. It would be the Record but it should always have at least id, tableName and colName.

src/meta/cycleData/link/link.ts
yaguzmang: We had it in our first notes but I agree we don't need it 👍

yaguzmang: @minotogna @sorja 
I am writing the query to upsert a link. It goes like:
```
insert into ${schemaCycle}.link 
         (country_iso, link, location, props, target)
         values ($1, $2, $3::jsonb, $4::jsonb, $5)
  on conflict (link, country, ???) do update
  props = ...
```
If the same link (url) exists in different locations, do we want to treat it as if it was actually two different links?
If so, we would need to have the 'location_id' as a new column to add a (link, country, location_id) unique constraint.

src/meta/cycleData/link/link.ts
minotogna: Why? Aren't these property of the record ?

src/meta/cycleData/link/link.ts
minotogna: I don't remember that . Last visit is the last element of the array .

src/meta/cycleData/link/link.ts
minotogna: Good point . Either we treat it as tow different links and am we can keep target or even use an array of targets .
Maybe better to use the first solution ?

yaguzmang: I have an idea, let me know what you think:
We make each link/country unique, and we handle different locations with an array of link locations.
The hard part of having an array of locations is to update them when a link is moved/deleted. What we can do there is: when we run the validation for all links, first we set all link locations to null. Then we run the validation, which will add the updated link locations. At the end, any link that still has a null location can be deleted or ignored because it means there are no instances of it anymore.

src/meta/cycleData/link/link.ts
yaguzmang: I mean having it like this so those properties are mandatory in the type:
```
export type LinkLocation = {
  colName: string
  id: number
  tableName: string
} & Record<string, string | number>
```

minotogna: I meant only Record . tableName is for example the key and the value the correspondent table … etc … 
Does it make sense ?

src/meta/cycleData/link/link.ts
minotogna: I think let’s not over complicate things .
Let’s just get rid of target , even because in our editor is alway _blank .
What do you think ?

yaguzmang: Yes. We won't have target. 
My idea there is related to the same link being used in different locations.
For example, "https://forest-reference.com/123" being used in two different data sources. We would need all locations of that link to be able to show the user where to edit them, no?

yaguzmang: Some examples of links that are used multiple times in different places:
image

minotogna: Would it be complicated to have a jsonb array column locations then ?
What do you think ?

yaguzmang: It will require a bit more pre-processing before inserting, but I think this is the way to go 👍

minotogna: okkkk 👍

minotogna (commented):
this looks good to me. @sorja you want to take a look ?

sorja (approved):

PR #3792, opened 2024-05-06
value aggragate

### LINKED ISSUE #3785: Migrate  Value Aggregate --> NodeExt
- [x] generate database views union node_ext and delete/remove total land area from node table
- [x] migrate value_aggregate to node_ext { type: node }
- [x] Check NDP total land area
- [x] remove all aggregate stuff from repository/controller
- [x] remove value_aggregate table
- [x] (use only node_ext)
- [x] type node.unit? = Unit
- [x] `nodeValue.FAOEstimate` and the valueAggregate should be only for 2020
- [x] For 2025 keep totalLandArea

For regions and global:
- [x] Include region data in data views (eg. `country_iso = 'AS'`)
- [x] Check WO data is correct

For refreshing:

- [x] controller.cycleData.data.refreshTableDataViews { assessment, cycle }
- [x] get tables with TableRepository
- [x] createOrReplaceTableDataView({ table, assessment, cycle })
- [x] add migration step to call it and regenerate metacache

see: src/server/repository/assessmentCycle/data/createOrReplaceTableDataView.ts

sorja:


sorja:
remove query for global and region
and remove fao estimates from the view

ignore redis, query DB directly

sorja:
1. remove region and global data from views (refreshView query remove region / global)
2. when in cycleData controller if (countryIso = global | region) use aggregated data repository or similar
3. in that repository/controller join with node ext as current the currnet slwo view is generated with sub queries
4. ?

PR #3803, opened 2024-05-10
[deprecated] 3785 value aggregate fix cache

sorja:
- Fix api request
- Introduce new param: 'faoEstimates'
- Fix redis caching data for regions
- Filter cached data: remove faoEstimates by default


src/server/repository/assessmentCycle/data/getTableData.ts
```diff
@@ -69,8 +70,13 @@ export const getTableData = (props: Props, client: BaseProtocol = DB): Promise<R
...
               and e.col_name is not null
-              -- exclude fao estimated values
-              and e.value ->> 'faoEstimate' is null
+              ${
+                faoEstimates
+                  ? ''
+                  : `
+               -- exclude fao estimated values
+              and e.value ->> 'faoEstimate' is null`
+              }
```
sorja, line 79: Not sure if this is needed?
This is only used for caching the data, right @minotogna ?

We never access the "database data" directly, we always access through redis?

PR #3813, opened 2024-05-17
3743 - Broken Links - Add visit cycle links worker

yaguzmang:


src/server/controller/cycleData/link/visitCycleLinks/visitCycleLinks.ts
```diff
@@ -0,0 +1,23 @@
...
+  removeOnComplete: true,
+  removeOnFail: true,
+}
+
+export const visitCycleLinks = (props: VisitCycleLinksProps): Promise<Job<VisitCycleLinksProps>> | undefined => {
+  const { assessment, cycle } = props
+
+  Logger.debug(`[visitCycleLinks] added visit all links job for ${assessment.props.name} ${cycle.name} to the queue`)
+
+  // TODO: persist when this function is already running, and if so, return early.
```
yaguzmang, line 19: How do we want to persist when a job is running for a given cycle so that we can display it in the FE? is there a table where we have this kind of information? or should we just use a socket event for that

minotogna: Let's talk in the daily tomorrow ? 
Can u work in parallel on the ui in the meantime ?

yaguzmang: Sure. I'm doing so at the moment 👍

minotogna (approved):

PR #3819, opened 2024-05-22
3690 - User pictures to public.file

### LINKED ISSUE #3690: performance issue: move public.users.profile_picture_file to public.file
Refactor db `users`:
- Move profile_picture_file to public.file
- Investigate if we want to keep profile_picture_filename or would it be enough to pull this information from public.file
- Add new columns file_uuid or profile_picture_file_uuid

sorja:
https://github.com/openforis/fra-platform/assets/5508251/37538cd7-b27b-46cd-bc2b-d1b721871f3b

### DIFF
--- a/src/server/repository/public/user/update.ts
+++ b/src/server/repository/public/user/update.ts
@@ -1,6 +1,7 @@
 import { User } from 'meta/user'
 
 import { BaseProtocol, DB } from 'server/db'
+import { FileRepository } from 'server/repository/public/file'
 
 import { getOne } from './getOne'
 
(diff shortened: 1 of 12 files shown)


src/server/repository/public/user/update.ts
```diff
@@ -26,18 +27,15 @@ export const update = async (
   )
 
   if (profilePicture) {
-    const {
-      profilePicture: { originalname, buffer },
-    } = props
+    const createdFile = await FileRepository.create({ file: profilePicture }, client)
```
minotogna: Repositories should be used inside controllers only . This should be moved at the controller level

sorja: Sure, updated here and cleaned the logic a bit to avoid double query:

https://github.com/openforis/fra-platform/pull/3819/commits/44159ced058a85581fe9686905dbee828046a49f

minotogna (approved):

PR #3826, opened 2024-05-27
fix migration

sorja:
Tested:

0. Reset DB
1. Run migration-steps
2. Run migrations

--

0. Reset DB
1. Run migrations
2. Run migration-steps


minotogna (approved):
👌

PR #3830, opened 2024-05-29
3743 - Broken Links - Add sockets events and Get verification job status

yaguzmang:
Adding sockets events and adding an endpoint to get the verification job status.
The idea is to get the verify job status when the UI page loads to potentially disable the verify button, and then listen to socket events.

### DIFF
--- a/src/server/controller/cycleData/links/visitCycleLinks/workerFactory.ts
+++ b/src/server/controller/cycleData/links/visitCycleLinks/workerFactory.ts
@@ -1,7 +1,7 @@
-import { Worker, WorkerOptions } from 'bullmq'
+import { Worker, WorkerListener, WorkerOptions } from 'bullmq'
 import IORedis from 'ioredis'
 
-import { ActivityLogMessage } from 'meta/assessment'
+import { ActivityLogMessage, Assessment, Cycle } from 'meta/assessment'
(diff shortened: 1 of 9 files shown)


src/server/controller/cycleData/links/visitCycleLinks/workerFactory.ts
```diff
@@ -45,6 +53,12 @@ const newInstance = (props: { key: string }) => {
   worker.on('failed', async (job, error) => {
     const { assessment, cycle, user } = job.data
 
+    const linksVerificationEvent = Sockets.getLinksVerificationEvent({
+      assessmentName: assessment.props.name,
+      cycleName: cycle.name,
+    })
+    SocketServer.emit(linksVerificationEvent, { event: 'failed' })
```
minotogna: what about on started ? Also, I think this code can be grouped in a common function

yaguzmang: I was handling the disable state of the button in the frontend in a way that it didn't need the 'on started' socket event, so I didn't add it. But, it's better to have it 👍  
Also added a common function to emit events in the worker factory.

minotogna (approved):

PR #3834, opened 2024-05-31
3743 - Broken Links - Delete unused links and disable approve button

yaguzmang:
(UPDATED) Soft deleting links that are no longer present:

https://github.com/openforis/fra-platform/assets/41337901/411c5c7e-4623-44a2-b86c-afe839dc1b01

UI without visits column:
image


src/server/controller/cycleData/links/visitCycleLinks/worker.ts
```diff
@@ -42,12 +42,16 @@ export default async (job: VisitCycleLinksJob): Promise<void> => {
...
-    await LinkRepository.upsertMany({
+    const updatedLinks = await LinkRepository.upsertMany({
       assessment,
       cycle,
       linkVisits,
     })
 
+    const excludedIds = updatedLinks.map((link) => link.id)
+
+    await LinkRepository.deleteMany({ assessment, cycle, excludeApproved: true, excludedIds })
```
yaguzmang: Do we want to remove links regardless if they are approved? if so, when they are deleted and maybe added again later by a user, they would need to be approved again.

minotogna: Should we mark them as deleted instead of hard delete ? So the approval remains, but in the ui we show the non delete only .
What do you think ?

yaguzmang: If we don't want to delete them then I think the best approach would be to update the `locations` field to an empty object. Then in the UI we show the links that are at least in one location. That would be more consistent with the current approach. What do you think?

sorja:
No of visits refers to our backend visitng the links?
Seems a bit misleading to show this info there?

yaguzmang:
> No of visits refers to our backend visitng the links? Seems a bit misleading to show this info there?

@sorja misleading in what sense? maybe we should rename it to No of validations or something similar?

sorja:
> > No of visits refers to our backend visitng the links? Seems a bit misleading to show this info there?
> 
> @sorja misleading in what sense? maybe we should rename it to No of validations or something similar?

I think on first view with no further explanatin it seems to refer how many times user has visited this link?
Is it relevant information to user (admin) how many times the link has been validated?

minotogna:
> > > No of visits refers to our backend visitng the links? Seems a bit misleading to show this info there?
> 
> > 
> 
> > @sorja misleading in what sense? maybe we should rename it to No of validations or something similar?
> 
> 
> 
> I think on first view with no further explanatin it seems to refer how many times user has visited this link?
> 
> Is it relevant information to user (admin) how many times the link has been validated?

Yes. @yaguzmang @sorja  let's remove that column ?

src/server/controller/cycleData/links/visitCycleLinks/worker.ts
minotogna: It might lead to confusion . It is better to explicitly save the deleted state in a prop maybe . What do you think ? @yaguzmang @sorja

src/server/repository/assessmentCycle/links/getCount.ts
```diff
@@ -8,17 +8,24 @@ import { BaseProtocol, DB, Schemas } from 'server/db'
 type Props = {
   assessment: Assessment
   cycle: Cycle
+  excludeDeleted?: boolean
```
minotogna, line 11: This prop , here and in the get can by by default true , so we don't pass it down. What do you think ?

yaguzmang: Sure. Added it 👍

minotogna (approved):

PR #3839, opened 2024-06-13
3645 - Dashboard: Add hidden variables for carbon stock

sorja:
Screenshot taken after running migrations and enabling hiding logic in UI:

image

Same table when hiding enabled:

image


minotogna (approved):
👍

PR #3847, opened 2024-06-25
3645 - Dashboard: Show real values for pie chart

sorja:
image

Note: 
The tooltip api doesn't support percentage.
https://recharts.org/en-US/api/Tooltip

It is possible to spend a bit more time to investigate this if needed

Current work around is:

Pie label shows percentage
Tooltip shows the real values

### DIFF
--- a/src/client/pages/CountryHome/Overview/meta/forestAreaPercentOfLandArea/index.ts
+++ b/src/client/pages/CountryHome/Overview/meta/forestAreaPercentOfLandArea/index.ts
@@ -19,7 +19,8 @@ const rowMetadata: RowsMetadata = [
     id: 1,
     variableName: 'forestArea',
     label: `statisticalFactsheets.rowName.forestArea`,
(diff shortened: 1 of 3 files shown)


src/client/pages/CountryHome/Overview/meta/forestAreaPercentOfLandArea/index.ts
```diff
@@ -19,7 +19,8 @@ const rowMetadata: RowsMetadata = [
     id: 1,
     variableName: 'forestArea',
     label: `statisticalFactsheets.rowName.forestArea`,
-    calculateFn: `100 * (${tableName}.forestArea / ${tableName}.totalLandArea)`,
+    calculateFn: `${tableName}.forestArea`,
+    // calculateFn: `100 * (${tableName}.forestArea / ${tableName}.totalLandArea)`,
```
sorja, line 23: Left as reference if we need to use this formula again

minotogna (approved):

PR #3873, opened 2024-08-09
3645 - Dashboard: Use only value aggregate

sorja:
Comparing table data with our script results in:
```
debug: No differences found
```


src/test/migrations/steps/20240503121725-step-insert-value_aggragate-into-node_ext.ts
```diff
@@ -1,14 +1,14 @@
 import { Promises } from 'utils/promises'
```
sorja, line 1: @minotogna 
Github might not show changes for this file, but this has also changed!

src/server/repository/assessmentCycle/data/getFaoEstimateData.ts
```diff
@@ -7,27 +7,30 @@ import { BaseProtocol, DB, Schemas } from 'server/db'
...
-             , ne.value                    as value
+        select 
+               '${regionCode}'                                                                      as country_iso
+             , ne.props ->> 'tableName'                                                             as table_name
+             , ne.props ->> 'variableName'                                                          as variable_name
+             , ne.props ->> 'colName'                                                               as col_name
+             , jsonb_build_object('raw', sum((ne.value ->> 'raw')::numeric), 'faoEstimate', true)   as value
           from ${schemaCycle}.node_ext ne
-          where ne.type = 'node' and ne.value ->> 'faoEstimate' = 'true'
+          where ((ne.type = 'node' and ne.value ->> 'faoEstimate' = 'true') or (ne.props ->> 'variableName' = 'totalLandArea'))
```
minotogna, line 30: Why hard coding totalLandArea?

minotogna: I think total land area is not a fao estimate . So it shouldn't be there ?

sorja: We want to include total land area, as it's used in the regions dashboard view.
But total land area is not marked as `faoEstimate`, so we work around this by adding this `or` clause.
Previously, total land area was populated through country data always containig total land area

Let me know

minotogna: According to the logic we discussed last day, also total land area should be part of the dataset we imported from OJ ? 

If yes, then we have duplicate total land area sources ?

sorja: It is not duplicated - it is already in nodeExt (previously imported), but the difference is that it doesn't have the flag `faoEstimate`


Total land area:
image

Forest Area:
image


Let me know

minotogna: I understood your point the first time you said it ...
I meant duplicate sources , not data .
Never mind 😅

minotogna (approved):

src/server/repository/assessmentCycle/data/getFaoEstimateData.ts
sorja: >  I meant duplicate sources , not data .

Ok, I think it should be fine!

PR #3874, opened 2024-08-12
3645 - Dashboard: Hide overview for unpublished regions

sorja:
Note:
Remove unused file left in main PR: src/.../getFaoEstimateViewName.ts

### DIFF
--- a/src/client/pages/CountryHome/FraHome/hooks/useSections.tsx
+++ b/src/client/pages/CountryHome/FraHome/hooks/useSections.tsx
@@ -1,10 +1,11 @@
 import React, { useMemo } from 'react'
 
 import { Areas } from 'meta/area'
+import { AssessmentNames } from 'meta/assessment'
 import { SectionNames } from 'meta/routes'
 import { Users } from 'meta/user'
 
-import { useCycle } from 'client/store/assessment'
+import { useAssessment, useCycle } from 'client/store/assessment'
 import { useUser } from 'client/store/user'
 import { useCountryRouteParams } from 'client/hooks/useRouteParams'
 import Collaborators from 'client/pages/CountryHome/FraHome/Collaborators'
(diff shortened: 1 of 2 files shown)


src/client/pages/CountryHome/FraHome/hooks/useSections.tsx
```diff
@@ -27,7 +28,11 @@ export const useSections = (): Array<Section> => {
 
     if (!cycle) return null
 
-    sections.push({ name: SectionNames.Country.Home.overview, component: Overview })
+    const showOverview = cycle.published || Areas.isISOCountry(countryIso)
```
sorja: would it be better to check that it's fra2025 vs published cycle?



```suggestion
const assessmentFra2025 = assessment.props.name = 'fra' && cycle.name = '2025'
const showOverview = assessmentFra2025  || Areas.isISOCountry(countryIso)
```

sorja: I am not sure if it matters, only case is that is it possible to publish the data before the dashboard?

minotogna: > would it be better to check that it's fra2025 vs published cycle?
> 
> 
> 
> 
> 
> 
> 
> ```suggestion
> 
> const assessmentFra2025 = assessment.props.name = 'fra' && cycle.name = '2025'
> 
> const showOverview = assessmentFra2025  || Areas.isISOCountry(countryIso)
> 
> ```

I think it's good published or country iso

minotogna: > I am not sure if it matters, only case is that is it possible to publish the data before the dashboard?

Good point . It shouldn't be possible, but the reality might be different .
It is safer to check fra 2025 explicitly, disregard my previous comment

sorja: [1521a9d](https://github.com/openforis/fra-platform/pull/3874/commits/1521a9d32dcd024fd342f2abb4d81828d3de104c)

minotogna (approved):

PR #3879, opened 2024-08-15
Add assessmentName/cycleName structure to data download

sorja:
Files are now same for 2020 and 2025
We need to get new files for 2025

Note:
This is pre-emptive PR to simplify upcoming file change for 2025 data download

image


src/server/service/file/index.ts
```diff
@@ -26,9 +26,9 @@ export const fileTypes = {
     downloadName: `Statistical Factsheets (${levelIso})`,
     fileType: 'ods',
   }),
-  dataDownload: (key: string, fileType: string) => ({
+  dataDownload: (assessmentName: string, cycleName: string, key: string, fileType: string) => ({
     key,
-    folder: 'dataDownload',
+    folder: `${assessmentName}/${cycleName}/dataDownload`,
```
sorja, line 31: the url was already in correct format, but the path was never used

simply moved the files to `assessmentName/cycleName/dataDownload/fileName`

src/server/api/file/getDataDownloadFile.ts
```diff
@@ -3,15 +3,17 @@ import { Request, Response } from 'express'
...
 
 type Query = {
+  assessmentName: string
+  cycleName: string
   fileName: string
   fileType: string
   language: string
 }
 
 export const getDataDownloadFile = async (req: Request, res: Response) => {
```
minotogna, line 13: Check if you can use cycle request or similar  to avoid casting req.query

sorja: [d9603be](https://github.com/openforis/fra-platform/pull/3879/commits/d9603be6273c2b87b3f5108ca371dbcc280f15f9)

minotogna (approved):

PR #3880, opened 2024-08-16
3645 - Dashboard: Update tooltip style

sorja:
image

https://github.com/user-attachments/assets/8f370c0d-8ff2-441e-afa3-a9f3299025e7

### DIFF
--- a/src/client/components/Chart/Bar/Bar.tsx
+++ b/src/client/components/Chart/Bar/Bar.tsx
@@ -16,8 +16,8 @@ import {
 import { Labels } from 'meta/assessment'
 import { BarChart as BarChartType, BarChartData } from 'meta/chart'
 
-import Tooltip from 'client/components/Chart/Bar/Tooltip/Tooltip'
-import { cursor } from 'client/components/Chart/utils/cursor'
+import { cursor } from '../utils/cursor'
+import Tooltip from './Tooltip'
 
 type Props = {
   data: BarChartData
--- a/src/client/components/Chart/Bar/Bar.tsx
+++ b/src/client/components/Chart/Bar/Bar.tsx
@@ -9,6 +9,7 @@ import {
   Legend,
   ResponsiveContainer,
   Tooltip,
+  TooltipProps,
   XAxis,
   YAxis,
 } from 'recharts'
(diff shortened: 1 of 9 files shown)


src/client/components/Chart/Bar/Bar.tsx
```diff
@@ -8,15 +8,16 @@ import {
...
+  Tooltip as TooltipRecharts,
   XAxis,
   YAxis,
 } from 'recharts'
 
 import { Labels } from 'meta/assessment'
 import { BarChart as BarChartType, BarChartData } from 'meta/chart'
 
-import TooltipContent from 'client/components/Chart/TooltipContent'
+import Tooltip from 'client/components/Chart/Bar/Tooltip/Tooltip'
```
minotogna: Check the import

src/client/components/Chart/Bar/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+
+import { TooltipProps } from 'recharts'
+
+import TooltipContent from 'client/components/Chart/TooltipContent'
+
+const Tooltip: React.FC<TooltipProps<never, never>> = (props) => {
```
minotogna, line 7: Is this component really needed ?
Can't the bar just get the data from a hook and pass it to the TooltipContent ?

src/client/components/Chart/Pie/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const content = payload.map((item) => ({
+    color: item.payload.color,
+    label: Labels.getLabel({ label: item.payload.label, t }),
+    name: item.payload.name,
+    percent: ((item.value as number) / totalValue) * 100,
+    unit: t(item.payload.unit),
+    value: item.value as number,
+  }))
+
+  return <TooltipContent content={content} />
```
minotogna, line 27: Same as above

src/client/components/Chart/Bar/Tooltip/Tooltip.tsx
sorja: image

I tried this kind of approach, some issues:

image
https://recharts.org/en-US/api/Tooltip#content

- Tooltip content expects a `ReactElement | Function`
- The rendering of the tooltip content slightly differs for Pie and Bar chart:

```
// Pie: (note item.payload)
  const content = payload.map((item) => ({
    color: item.payload.color,
    label: Labels.getLabel({ label: item.payload.label, t }),
    name: item.payload.name,
    percent: ((item.value as number) / totalValue) * 100,
    unit: t(item.payload.unit),
    value: item.value as number,
  }))

// Bar:
  const content = payload.map((item) => ({
    color: item.color,
    label: item.name,
    name: item.name,
    unit: item.unit,
    value: item.value as number,
  }))

```

For internal hook (eg `Bar/useTooltipContent`) the issue would be that the payload is passed as prop to the element:

```
        <TooltipRecharts content={<Tooltip totalValue={totalValue} />} cursor={cursor} />
```

Here the tooltip receives also `TooltipProps<never, never>` on top of the extra `totalValue`


Let me know what you suggest

minotogna: The total value is used to format the data .
You don't need another component that just wraps the data .
You can use directly tooltipContent and pass the data ?

src/client/components/Chart/Bar/Bar.tsx
minotogna: @sorja Can u fix the import ?

minotogna (approved):

PR #3900, opened 2024-09-02
3899: PanEuropean redirect from old url

### LINKED ISSUE #3899: PanEuropean: Redirect from old URL
Possible traces left for old URLs

> address has changed as our current website has always used  https://fra-data.fao.org/FE/panEuropean/home/. Same link was used also at https://foresteurope.org/state-of-europes-forests/

Investigate and discuss if we need redirect

minotogna:
https://github.com/user-attachments/assets/ab270633-c610-47fb-88fa-1292b33e0dad


sorja (approved):
Not sure if we check for old FRA links also, but maybe lets check later on if broken links are reported

👍

PR #3903, opened 2024-09-04
2480 - Add new command: migration-public:create

sorja:
image


src/tools/migrations/public/index.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Logger } from 'server/utils/logger'
+
+const client = DB
+let migrationSteps: Array<string>
+let previousMigrations: Array<string> = []
+const executedMigrations: Array<string> = []
+
+// TODO: Move to migration step / schema initialisation
+const tableDDL = `
+    create table if not exists public.migrations (
```
sorja, line 18: what should we call the db table? is `migrations` fine?

src/tools/migrations/public/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  previousMigrations = await client.map('select * from migrations', [], (row) => row.name)
+  migrationSteps = fs
+    .readdirSync(path.join(__dirname, `steps`))
+    .filter((file) => file !== 'template.ts' && file.endsWith('.ts') && !previousMigrations.includes(file))
+}
+
+const close = async () => {
+  // Note: Omitted REDIS related code. Add them here if needed.
+  await DB.$pool.end()
+}
```
sorja, line 37: for now i omit the redis related code, eg. connection closing etc as i don't think we need them in DDL manipulation
can be added as needed

src/tools/migrations/public/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const exec = async () => {
+  await init()
+
+  // eslint-disable-next-line no-restricted-syntax
+  for await (const file of migrationSteps) {
+    try {
+      Logger.info(`Running migration ${file}`)
+      // eslint-disable-next-line @typescript-eslint/no-var-requires,global-require,import/no-dynamic-require
+      await require(`./steps/${file}`).default(client)
```
sorja: transcation has been removed here

it is also possible to omit passing client as we initialise the variable in the template or update template to take param

src/tools/migrations/public/index.ts
minotogna: Sure 👍🏾

src/tools/migrations/public/index.ts
```diff
@@ -0,0 +1,65 @@
...
+const close = async () => {
+  // Note: Omitted REDIS related code. Add them here if needed.
+  await DB.$pool.end()
+}
+
+const exec = async () => {
+  await init()
+
+  // eslint-disable-next-line no-restricted-syntax
+  for await (const file of migrationSteps) {
```
minotogna: Is order by date 100% guaranteed?

src/tools/migrations/public/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const close = async () => {
+  // Note: Omitted REDIS related code. Add them here if needed.
+  await DB.$pool.end()
+}
+
+const exec = async () => {
+  await init()
+
+  // eslint-disable-next-line no-restricted-syntax
```
minotogna: To avoid this, what about Promises.each ?

src/tools/migrations/public/index.ts
minotogna: Client actually should not be passed , because ddl cannot be executed in transaction

minotogna (commented):
Check deep scan as well

sorja:
> Check deep scan as well 

Deepscan gives error on unused variable in the template file, what so you suggest?

minotogna:
> > Check deep scan as well
> 
> Deepscan gives error on unused variable in the template file, what so you suggest?

remove the variable. we'll use DB directly in our migrations ?

src/tools/migrations/public/index.ts
minotogna: @sorja is order guaranteed?

minotogna (approved):

src/tools/migrations/public/index.ts
sorja: Should be fine!

image

image

PR #3906, opened 2024-09-06
2480 - Init admin user

sorja:
kuva


src/tools/migrations/public/steps/20240906093649-step-init-admin-user.ts
```diff
@@ -0,0 +1,33 @@
+import { passwordHash } from 'server/api/auth/utils/passwordUtils'
+import { DB } from 'server/db'
+
+const ADMIN = {
+  name: 'Admin',
```
sorja, line 5: It is possible to read these from .env var eg
```
// env
ADMIN_EMAIL=''
ADMIN_PASSWORD=''

// migrations/other
ProcessEnv.Admin.Email
```

minotogna (approved):

PR #3911, opened 2024-09-09
3910 - Fix memory issues in broken links worker

### LINKED ISSUE #3910: Fix memory issues in Broken Links worker
The Broken Links feature generates a memory spike of over 2000MB upon calling the visitCycleLinks worker. This can crash the prod server.

yaguzmang:
I tracked down the spikes in memory usage in the Visit Cycle Links Job, and it turned out to be in the function `links/getAllLinksToVisit -> _processLinks -> getLinksFromHtml`

`getLinksFromHtml` uses the JSDOM library to parse the html text and get the links from the `a` tags: 
`const dom = new JSDOM(html) ....`.
It seems the memory is not freed after that function is done running due to how the internals of JSDOM work, causing the heap to grow until after all links in all locations are processed. 

I tried manually deleting the references to the JSDOM instance, but it didn't help:
```
dom.window.close();
document = null;
dom = null;
```
More on the issue: https://github.com/jsdom/jsdom/issues/1682

Solution: using `htmlparser2` instead, which is a module we already have (it is a dependency of a library we are using).

Memory usage before (peak: 1856 MB): 
Pasted Graphic 7
 
 Memory usage after (peak: 140 MB): 
 
Pasted Graphic 6

### DIFF
--- a/package.json
+++ b/package.json
@@ -197,6 +197,7 @@
     "express-validator": "^6.14.2",
     "extract-zip": "^2.0.1",
     "fast-csv": "^4.3.6",
+    "htmlparser2": "^9.1.0",
     "http-proxy-middleware": "^2.0.4",
     "i18next": "^21.6.14",
     "ioredis": "^5.2.4",
--- a/src/client/pages/Admin/Admin.tsx
+++ b/src/client/pages/Admin/Admin.tsx
@@ -30,10 +30,10 @@ const sections: Array<Section> = [
     name: SectionNames.Admin.userManagement,
     labelKey: 'landing.sections.userManagement',
   },
-  // {
-  //   name: SectionNames.Admin.links,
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -197,6 +197,7 @@
     "express-validator": "^6.14.2",
     "extract-zip": "^2.0.1",
     "fast-csv": "^4.3.6",
+    "htmlparser2": "^9.1.0",
```
yaguzmang, line 200: Added to comply with commit hooks. We already had it, as it is a depency of "html-webpack-plugin":

into Reasons this module exists

minotogna: Yes , if used directly in our code, it must be explicitly added

minotogna (approved):

PR #3921, opened 2024-09-12
2480 - Run migrations on app start

sorja:


package.json
```diff
@@ -24,7 +24,7 @@
     "run:dev:server": "ts-node-dev --exit-child src/server/start",
     "run:prod": "run-s build start",
     "run:prod:server": "node dist/server/start",
-    "start": "run:prod:server",
+    "start": "run-s migration-public:run run:prod:server",
```
minotogna, line 27: Also in dev mode ?

sorja: Sure

package.json
```diff
@@ -21,10 +21,11 @@
     "migration-steps:watch": "ts-node-dev --respawn --transpile-only --exit-child --watch src/test/migrations/ src/test/migrations/ -- --watch",
     "run:dev": "run-p run:dev:client run:dev:server",
     "run:dev:client": "cross-env NODE_ENV=development webpack serve --config webpack.config.babel.js",
-    "run:dev:server": "ts-node-dev --exit-child src/server/start",
+    "dev:server": "ts-node-dev --exit-child src/server/start",
+    "run:dev:server": "run-s migration-public:run dev:server",
```
minotogna: Why double script ? One is enough ?

sorja: https://github.com/mysticatea/npm-run-all/blob/master/docs/run-s.md

`npm-run-all` expects NPM script as param, i tried different combinations, but all give error if not written as in this PR

One error for example when using quotation marks:

image

sorja: Would you prefer `&&` syntax over `run-s` then?

minotogna: Up to you

sorja: We can keep as it is in this PR then 👍

minotogna: Can we please have one script only ?

sorja: We cannot using the `npm-run-all`, so we must use `&&` which then defeats the purpose of having the package `npm-run-all`

Changed to `&&` to close PR

minotogna: let's talk about that later on. but the run-p command is still good to use?

minotogna (approved):

package.json
sorja: parallel (p) and sequential (s)
but only work on package.json scripts

PR #3922, opened 2024-09-12
2480 - Fix public.migrations column timestamp and delete old migration entries from table

sorja:
Add by default timestamp so we don't have to add it from outside
It is now consistent with migration_steps


src/test/migrations/steps/20240912102424-step-fix-migrations-table-ddl.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseProtocol, DB } from 'server/db'
+
+const client: BaseProtocol = DB
+export default async () => {
+  await client.query(`
+    alter table migrations
+    alter column run_on set default now(),
+    alter column run_on type timestamp without time zone
```
minotogna: Since we haven't merged this yet, I wouldn't add a step , but update only initial step . What do u think ?

sorja: Unfortunately we have this table already in public schema
image


We reuse the same table that we had for the old migrations system, but we will empty it

What do you suggest?

minotogna: Update the initial step ? Didn't you do this already ?

minotogna: oh I get it now. 
this step should run before inserting the migrations in this table for current applications? or, can we do it in the same migration step (event better)?

sorja: Current status:
We have `public.migrations` table, that has no default value when inserting new rows
The idea of this step is to update the existing `migrations` table to be up to date with our init public schema, which has updated version of the `migrations` table.

So we are only updating what we already have in DB, hope this makes sense!

Let me know

minotogna: yes, we need one migration step that operates on current applications. ok ?

sorja: It is this one introduced in this PR?
`src/test/migrations/steps/20240912102424-step-fix-migrations-table-ddl.ts`

Sorry, maybe I am not following!

sorja:
Note:
Single migration step to:
- update `public.migrations`
- insert lines in `public.migrations` for migrations we dont wanna run in our existing instances

.github/workflows/test.js.yml
```diff
@@ -55,14 +55,14 @@ jobs:
             NODE_AUTH_TOKEN: ${{ secrets.ACCESS_TOKEN }}
```
sorja, line 55: Restore the original order we had.
First run migrations and then build the application

I don't think it matters in practice, but it makes more sense to be in this original order.

minotogna (approved):

PR #3928, opened 2024-09-13
3897 - Admin manage collaborators refactor - Clean up code

yaguzmang:


src/client/store/ui/userManagement/stateType.ts
```diff
@@ -1,9 +1,6 @@
-import { RoleName, User } from 'meta/user'
+import { User } from 'meta/user'
 
 export interface UserManagementState {
-  count: {
-    [key in keyof typeof RoleName | 'totals']?: number
-  }
   user?: User
```
yaguzmang, line 4: This state `user` is used as user to edit in a couple different places. Now that this slice doesn't have anything related to filters, count, and so on, I think keeping this and the `users` var makes sense.

minotogna: I wouldn't touch this now

minotogna (approved):

PR #3929, opened 2024-09-16
3914 -  Cycle Status

### LINKED ISSUE #3914: Cycle Status
- [x] Add CycleStatus 
- [x] add dates of status (dateXX) 
- [x] remove published

```typescript
enum CycleStatus {
  draft = 'draft',
  editing = 'editing',
  published = 'published',
}

export type CycleProps = {
  status: CycleStatus
  dateCreated: string
  dateDraft: string
  dateEditing: string
  datePublished: string
}

export type Cycle {
  id: number
  name: CycleName
  uuid: CycleUuid
  // published: boolean 
  props: CycleProps
}
```

Official dates for :
  dateCreated: 1 year before 1st of January
  dateEditing: 1st workshop
  datePublished: official report publish date

sorja:
image

- [x] PanEuropean redirect
- [x] Assessment create (cycle create default props)
- [x] migrations public alter ddl
- [x] migration step populate data


src/client/components/AreaSelector/CountryList/hooks/useUserCountryISOs.tsx
```diff
@@ -33,7 +33,7 @@ export const useUserCountryISOs = (): Record<CycleUuid, Record<string, Array<Cou
     if (!userCountryISOs[UserRoles.noRole.role]) userCountryISOs[UserRoles.noRole.role] = {}
 
     // Only show countries of cycle if published when no role
-    if (cycle.published) {
+    if (cycle.props.status === CycleStatus.published) {
```
minotogna: Should we have a utility Cycles.isPublished ?

minotogna:
@sorja can we also update src/client/pages/PanEuropeanRedirect/PanEuropeanRedirect.tsx to get last published cycle ?

minotogna:
@sorja also, the AssessmentController.createCycle should be updated, keeping in mind the new props status and dateCreated

src/client/components/AreaSelector/CountryList/hooks/useUserCountryISOs.tsx
sorja: https://github.com/openforis/fra-platform/pull/3929/commits/d4016148824f6cb2c43be2111b69575d7e7eaeaf

sorja: https://github.com/openforis/fra-platform/pull/3929/commits/fc9bbaea83a371749aad1c47064ec087b0b5d9e6

src/meta/assessment/index.ts
```diff
@@ -27,6 +27,8 @@ export type {
 } from './commentableDescription'
 export { CommentableDescriptionName } from './commentableDescription'
 export type { Cycle, CycledPropsObject, CycleName, CycleUuid } from './cycle'
```
minotogna: I would also export CycleProps

src/server/repository/assessmentCycle/cycle/create.ts
```diff
@@ -10,14 +10,23 @@ const defaultMetaCache: AssessmentMetaCache = {
   variablesByTable: {},
 }
 
+const defaultProps: CycleProps = {
+  status: CycleStatus.draft,
+  dateCreated: new Date().toISOString(),
+  dateDraft: undefined,
```
minotogna: dateDraft should be the same as dateCreated (when creating a new cycle)

src/server/repository/assessmentCycle/cycle/create.ts
```diff
@@ -10,14 +10,23 @@ const defaultMetaCache: AssessmentMetaCache = {
...
+  dateEditing: undefined,
+  datePublished: undefined,
+}
+
 export const create = async (
   params: {
     assessment: Assessment
     name: string
+    props?: CycleProps
```
minotogna: maybe , when creating a new cycle, we don't need to pass any props

src/test/migrations/steps/20240909101347-step-exclude-public-migrations.ts
```diff
@@ -27,7 +27,7 @@ export default async () => {
     table: { table: 'public', schema: 'migrations' },
   })
 
-  const query = `${pgp.helpers.insert(values, cs)}`
+  const query = `${pgp.helpers.insert(values, cs)} on conflict do nothing;`
```
minotogna, line 30: why is this necessary ?

src/test/migrations/steps/20240919123658-step-update-cycle-props-published-date-status.ts
```diff
@@ -0,0 +1,69 @@
...
+export default async (client: BaseProtocol) => {
+  const exists = await client.oneOrNone(`
+    select 1 as exists
+    from information_schema.columns
+    where table_name='assessment_cycle' and column_name='published'
+  `)
+
+  if (exists) {
+    throw new Error('Column public.assessment_cycle.published exists. Check that migration-public ran correctly')
+  }
```
minotogna: should we run first migration steps and then migrations ?
If yes, I think the migration step can alter the cycle table and update the init public schema only. what do you think?

src/test/migrations/steps/20240909101347-step-exclude-public-migrations.ts
sorja: In case we double run migrations, I like to be safe
In my case, I ran public migrations and then steps as test, which caused `key already exists` error.

We can omit if you prefer

src/test/migrations/steps/20240919123658-step-update-cycle-props-published-date-status.ts
sorja: I think it's better currently:
Let `public` migrations alter the DDL of public schema on application start
If then, we still want to manipulate the state of DB we can do so with `steps`

With this approach:
- We will have consistent DB from empty DB without steps
- We might cause data loss if we are not careful, but this shouldn't be the case

If we do everything with steps, do we need `migrations-public`?
Should we just update `getPublicDDL` with the most up to date and do everything else in steps?

Let me know

minotogna: In theory yes .
The problem is another .
Current production hasn't been updated yet and in current production we should first run migration steps to populate the migrations tables and then we'll run migrations .

What will happen in this case ?
Will it throw this exception ? My guess is yes.

That's why I proposed to do in migration steps the current update .

Please test the flow with a prod db

src/server/repository/assessmentCycle/cycle/create.ts
```diff
@@ -10,6 +10,14 @@ const defaultMetaCache: AssessmentMetaCache = {
   variablesByTable: {},
 }
 
+const defaultProps: CycleProps = {
+  status: CycleStatus.draft,
+  dateCreated: new Date().toISOString(),
+  dateDraft: new Date().toISOString(),
```
minotogna: Are those two identical ? If not , maybe add a variable ?

src/server/repository/assessmentCycle/cycle/update.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { BaseProtocol, DB } from 'server/db'
+
+type Props = {
+  cycle: Cycle
+}
+
+const validateCycleProps = (cycle: Cycle): void => {
+  const { props } = cycle
+  const requiredFields: Array<keyof CycleProps> = ['status', 'dateCreated', 'dateDraft', 'dateEditing']
```
minotogna: Why should dateEditing and all the other dates be required ?

sorja: What are the required props, just status, dateCreated, dateDraft?

sorja: Or should we remove the check?

minotogna: let's remove the check for now

minotogna (approved):
👍

PR #3937, opened 2024-09-17
2480 - Introduce migrations schema

sorja:
Flow:

Running migration-steps:
- drop `public.migrations`
- init `migrations.public` table
- populate `migrations.public` with given rows (we dont want to run)

Running migration-public (after steps) (local instance)
- no migrations are ran as expected

Running migration-public (clean/new instance)
- public is initialized as expected

image

### DIFF
--- a/src/test/migrations/index.ts
+++ b/src/test/migrations/index.ts
@@ -17,8 +17,27 @@ let migrationSteps: Array<string>
 let previousMigrations: Array<string> = []
 const executedMigrations: Array<string> = []
 
+const tableDDL = `
+    create schema if not exists migrations;
+
+    do $$ 
+    begin
+      if exists (select 1 from information_schema.tables where table_schema = 'public' and table_name = 'migration_steps') then
+        alter table public.migration_steps rename to steps;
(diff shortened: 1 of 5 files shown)


src/test/migrations/index.ts
```diff
@@ -17,8 +17,27 @@ let migrationSteps: Array<string>
 let previousMigrations: Array<string> = []
 const executedMigrations: Array<string> = []
 
+const tableDDL = `
```
sorja, line 20: TODO: For some reasons some already ran migrations are running. Need to resolve this to open from draft

src/test/migrations/steps/20240909101347-step-exclude-public-migrations.ts
```diff
@@ -15,19 +16,15 @@ export default async () => {
   const pgp = pgPromise()
   // DROP CONTENT
   await client.query(`
-    truncate table public.migrations;
+    drop table if exists public.migrations;
```
sorja, line 19: We can safely drop the table now

minotogna (approved):

PR #3898, opened 2024-08-30
3855 move files to s3 bucket

sorja:
implementation:
- create a bucket to s3
- implement new service `FileStorage` to write and read files
- remove column `file` from `public.file`

todo:
- [ ] src/server/repository/public/ddl/getCreatePublicSchemaDDL.ts update `public.file` with public migration
- investigate if we want to remove `public.file`

todo/check:
- [x] upload and view avatar
- [x] upload and view file from Links and repository
- [x] upload and view file from EditorWYSIWYG

PR #3938, opened 2024-09-17
3798 - Admin Manage collaborators page - csv download (backend)

yaguzmang:
Example output:
image

Example output (user language = es):
image

To test locally: http://fra-data.local:9000/api/admin/users/export?assessmentName=fra&cycleName=2025


package.json
```diff
@@ -164,6 +164,7 @@
     "@openforis/arena-core": "^0.0.79",
     "@reduxjs/toolkit": "^1.8.1",
     "@socket.io/redis-streams-adapter": "^0.1.0",
+    "@types/csv-write-stream": "^2.0.3",
```
minotogna: Types dependencies are better added to dev dependencies

src/server/api/admin/exportUsers.ts
```diff
@@ -0,0 +1,25 @@
...
+    const { assessmentName, cycleName } = req.query
+
+    const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName })
+
+    const fileName = `users-${assessment.props.name}-${cycle.name}.csv`
+    const user = Requests.getUser(req)
+    const { lang } = user.props
+    const fileBuffer = await UserController.exportToCsv({ assessment, cycle, lang })
+
+    Responses.sendFile(res, fileName, fileBuffer)
```
minotogna: Aren't we piping to response stream directly?

src/server/controller/user/exportToCsv.ts
```diff
@@ -0,0 +1,98 @@
...
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+  lang: Lang
+}
+
+export const exportToCsv = async (props: Props): Promise<Buffer> => {
+  const { assessment, cycle, lang } = props
+
+  const query = `
```
minotogna: The query to fetch and export users must be generated only once . Especially when adding filters and sorting

src/server/controller/user/exportToCsv.ts
```diff
@@ -0,0 +1,98 @@
...
+    [RoleName.REVIEWER]: i18n.t(Users.getI18nRoleLabelKey(RoleName.REVIEWER)),
+    [RoleName.NATIONAL_CORRESPONDENT]: i18n.t(Users.getI18nRoleLabelKey(RoleName.NATIONAL_CORRESPONDENT)),
+    [RoleName.ALTERNATE_NATIONAL_CORRESPONDENT]: i18n.t(
+      Users.getI18nRoleLabelKey(RoleName.ALTERNATE_NATIONAL_CORRESPONDENT)
+    ),
+    [RoleName.COLLABORATOR]: i18n.t(Users.getI18nRoleLabelKey(RoleName.COLLABORATOR)),
+    [RoleName.VIEWER]: i18n.t(Users.getI18nRoleLabelKey(RoleName.VIEWER)),
+  }
+
+  const rowTransformer = (rawUser: any): Record<string, string> => {
```
minotogna: any should not be used (nowhere in the pr please) add generics if needed

src/server/api/admin/exportUsers.ts
minotogna: https://stackoverflow.com/questions/38788721/how-do-i-stream-response-in-express

sorja: I implemented the file stream util in another branch, you can take a look here if it is relevant also here 🥇 

https://github.com/openforis/fra-platform/pull/3898/files#diff-107de2a11951ae0c4f535746d9850c127f8dc9a360323f8efb4a5d3a66e1c739R32

minotogna: my guess is we could use directly res.write (as in the stackoverflow link )

package.json
minotogna: @yaguzmang have a look at the module 'csv' https://csv.js.org/transform/api/mixed/ . 

this csv-write-stream has been last updated 8 years ago, and I wouldn't suggest to use it
https://npmtrends.com/csv-vs-csv-write-stream

src/server/controller/user/exportToCsv.ts
yaguzmang: I was wondering about this. Should we expose a function `buildQuery` somehow in the users `getMany`? or did you have any other ideas on how we could centralize the logic for creating the query? Since we will run into this for any table we want to export

package.json
yaguzmang: We are using [fast-csv](https://www.npmjs.com/package/fast-csv) in a couple of places in the codebase. Since it works similarly, I'll switch to it here instead of csv-write-stream. What do you think?

src/server/controller/user/exportToCsv.ts
minotogna: yes, just a method that exports the query. centralise it just for the users

package.json
minotogna: isn't that just for reading csv ? or if it works, yes use that one . just don't use abandoned packages

src/server/controller/user/exportToCsvStream.ts
```diff
@@ -0,0 +1,70 @@
...
+    [RoleName.REVIEWER]: i18n.t(Users.getI18nRoleLabelKey(RoleName.REVIEWER)),
+    [RoleName.NATIONAL_CORRESPONDENT]: i18n.t(Users.getI18nRoleLabelKey(RoleName.NATIONAL_CORRESPONDENT)),
+    [RoleName.ALTERNATE_NATIONAL_CORRESPONDENT]: i18n.t(
+      Users.getI18nRoleLabelKey(RoleName.ALTERNATE_NATIONAL_CORRESPONDENT)
+    ),
+    [RoleName.COLLABORATOR]: i18n.t(Users.getI18nRoleLabelKey(RoleName.COLLABORATOR)),
+    [RoleName.VIEWER]: i18n.t(Users.getI18nRoleLabelKey(RoleName.VIEWER)),
+  }
+
+  const rowTransformer = (rawUser: User): Record<string, string> => {
```
yaguzmang: I typed rawUser here as `User` for now, but it would actually be a snake case version of it. If we want to add the snake cased version, an option would be introducing [SnakeCasedProperties](https://github.com/sindresorhus/type-fest/blob/main/source/snake-cased-properties.d.ts) from [type-fest](https://github.com/sindresorhus/type-fest/tree/main):
`type SnakeCaseUser = SnakeCasedProperties<User>;`
and we could re-use it elsewhere.

src/server/repository/public/user/index.ts
```diff
@@ -21,3 +21,6 @@ export const UserRepository = {
   remove,
   readCountryUsersByRole,
 }
+
+export type { UsersGetManyProps }
+export { usersBuildGetManyQuery }
```
yaguzmang: Verbose names because these are exported on their own (unlike for example `UserRepository.getMany` where it doesn't make sense to be verbose)

package.json
```diff
@@ -287,4 +288,4 @@
       "prettier --write"
     ]
   }
-}
+}
```
minotogna: let's keep the new line

src/server/repository/public/user/index.ts
minotogna: what about UserRepository.buildGetManyQuery ?

src/server/service/export/queryToCsvStream.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { BaseProtocol, DB } from 'server/db'
+
+type Props<QueryResultRow> = {
+  query: string
+  queryParams?: Parameters<Client['query']>[1]
+  rowTransformer?: (row: QueryResultRow) => ParserRow
+}
+
+export const queryToCsvStream = <QueryResultRow>(
```
minotogna, line 15: this is nice 👍

src/server/api/admin/exportUsers.ts
```diff
@@ -0,0 +1,61 @@
...
+      fullName: fullName || '',
+      lang,
+      limit: limit && Number(limit),
+      offset: offset && Number(offset),
+      orderBy,
+      orderByDirection,
+      roles: roles || [],
+      statuses: [UserStatus.active, UserStatus.disabled, UserStatus.invitationPending],
+    })
+
```
minotogna, line 46: I would actually slightly update the ExportService.queryToCsvStream to 
```typescript 
type Props = {
  fileName:string
  query: string
  queryParams?: Parameters<Client['query']>[1]
  res:Response
  rowTransformer?: (row: QueryResultRow) => ParserRow
}

ExportService.queryToCsvResponseStream(props)
```

So that the line 31-32 and 49-57 in this file are generically handled in the export service .
The UserController here, can have a method UserController.getManyExport that returns the query, the queryParams and the transformer function.

what do you think?

yaguzmang: My reasoning behind returning the stream itself was that the `ExportService.queryToCsvStream` could be a generic util service to be used for any purposes. Like piping the csv to other streams such as file writer, in memory, or even uploading to s3. 
What would you think of adding `ExportService.queryToCsvResponseStream` as a new function but it uses `ExportService.queryToCsvStream` internally? that way we keep the generic and encapsulate the response stream logic. 
Or if you think we won't really need the generic `queryToCsvStream`, I can make the changes you suggested to it.

src/server/repository/public/user/index.ts
yaguzmang: Sure. Will change it. At first I thought these `<Entity>Repository....` like `UserRepository` were reserved for actual DB operations only. 👍

src/server/api/admin/exportUsers.ts
minotogna: yes @yaguzmang sounds good 👍

minotogna (commented):
@sorja for me this looks really good. You Want to have a final check and approval ?

src/server/api/admin/exportUsers.ts
```diff
@@ -0,0 +1,51 @@
...
+    await ExportService.queryToCsvResponseStream<User>({ fileName, query, queryParams, res, rowTransformer })
```
sorja, line 47: For API I would use `Requests.` and `Responses.` but this is fine

sorja (approved):

PR #3941, opened 2024-09-19
3916 - Cycle clone: data

minotogna:


sorja (approved):
👍 


Check also:

`CREATE TABLE _schemaA.table_name AS TABLE schemaB.table_name;`

PR #3946, opened 2024-09-20
3942 - Bulk download: Forest restoration

sorja:


src/server/controller/cycleData/getBulkDownload/index.ts
```diff
@@ -164,12 +165,18 @@ export const getBulkDownload = async (props: { assessment: Assessment; cycle: Cy
 
   // Tier data only available for 2025
   if (cycle.name === '2025') {
```
minotogna, line 167: We will have problems with new cycles .
We must avoid hard cording cycle names from now on .

sorja: I know this, but I think this bulk download code will be replaced, so it should be safe for now, what do you think?

What do you suggest?

sorja:
@minotogna PR Updated with `htmlparser2`

minotogna (approved):

PR #3956, opened 2024-09-24
3955 - Dashboard: Global data not updated on filter

### LINKED ISSUE #3955: Dashboard: Global data not updated on filter
Global dashboard data should be updated by filtering countries selected in the filter.

https://github.com/user-attachments/assets/99379212-50e9-455f-a0b3-10e0a0a05acc

sorja:
Support filtering countries for global area code in Dashboard

## Non logged in user

https://github.com/user-attachments/assets/99b54c52-3edc-4a05-b33b-2da6995163eb

## Logged in user

https://github.com/user-attachments/assets/4c7ec30d-3b3d-4e7f-9372-a4736b00ceb4

- Refactor useGetTableData hook to handle multiple country ISOs
- Use useHomeCountriesFilter hook for country selection
- Update getTableData action and API to support multiple country ISOs

### DIFF
--- a/src/client/store/ui/home/hooks/index.ts
+++ b/src/client/store/ui/home/hooks/index.ts
@@ -1,0 +1,5 @@
+import { useAppSelector } from 'client/store'
+
+import { selectHomeCountriesFilter } from '../selectors'
+
+export const useHomeCountriesFilter = () => useAppSelector(selectHomeCountriesFilter)
--- a/src/client/store/ui/home/hooks/index.ts
+++ b/src/client/store/ui/home/hooks/index.ts
@@ -1,5 +1,5 @@
 import { useAppSelector } from 'client/store'
 
-import { selectHomeCountriesFilter } from '../selectors'
+import { HomeSelector } from '../selectors'
 
-export const useHomeCountriesFilter = () => useAppSelector(selectHomeCountriesFilter)
+export const useHomeCountriesFilter = () => useAppSelector(HomeSelector.getCountriesFilter)
--- a/src/server/api/cycleData/table/getTableData.ts
+++ b/src/server/api/cycleData/table/getTableData.ts
@@ -1,7 +1,7 @@
 import { Response } from 'express'
 
 import { CycleDataRequest } from 'meta/api/request'
-import { AreaCode, Areas, CountryIso } from 'meta/area'
+import { Areas, CountryIso } from 'meta/area'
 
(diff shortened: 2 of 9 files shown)


src/client/store/ui/home/hooks/index.ts
```diff
@@ -0,0 +1,5 @@
+import { useAppSelector } from 'client/store'
+
+import { selectHomeCountriesFilter } from '../selectors'
+
+export const useHomeCountriesFilter = () => useAppSelector(selectHomeCountriesFilter)
```
minotogna: better to use the convention HomeSelectors.getCountriesFilter ?

src/server/api/cycleData/table/getTableData.ts
```diff
@@ -1,14 +1,15 @@
...
-import { Areas, CountryIso } from 'meta/area'
+import { AreaCode, Areas, CountryIso } from 'meta/area'
 
 import { AssessmentController } from 'server/controller/assessment'
 import { CycleDataController } from 'server/controller/cycleData'
 import Requests from 'server/utils/requests'
 
 type GetTableDataRequest = CycleDataRequest<{
   tableNames: Array<string>
+  countryIso: AreaCode
```
minotogna: isn't this already in the CycleDataRequest ?

src/server/controller/cycleData/tableData/getAggregatedTableData.ts
```diff
@@ -9,14 +9,27 @@ import { DataRepository } from 'server/repository/assessmentCycle/data'
 import { Props } from './props'
 
 export const getAggregatedTableData = async (
-  props: Props,
+  props: Props & { countryIso: AreaCode },
```
minotogna, line 12: I think the controller, instead of getting a generic countryIso, should getter either an array of countryISOs or a regionCode. (The api would pass the right parameters based on request params).
This way it's more clear how to call this controller and what to expect. what do u think ?

sorja: image

The problem lies here, we want to use the region_code for the `countryIso` parameter after aggregating the values
And for `WO` with filter enabled, we would need both `regionCode` and a list of selected `countryISOs`

Let me know what you think

minotogna: yes, then pass ```regionCode?``` and ```countryISOs```. right?

PR #3957, opened 2024-09-24
Fix typo in user redirect

### LINKED ISSUE #3958: Fix redirect and migration-step for assessment_cycle

sorja:
Refactoring code missed one exclamation mark, caused redirect to break

### DIFF
--- a/src/test/migrations/steps/20240919123658-step-update-cycle-props-published-date-status.ts
+++ b/src/test/migrations/steps/20240919123658-step-update-cycle-props-published-date-status.ts
@@ -59,18 +59,12 @@ export default async () => {
   }
 
   await client.query(`
-      do $$
-      begin
-        if exists (
-          select 1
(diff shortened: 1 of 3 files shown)


src/test/migrations/steps/20240919123658-step-update-cycle-props-published-date-status.ts
```diff
@@ -59,18 +59,12 @@ export default async () => {
...
   await client.query(`
-      do $$
-      begin
-        if exists (
-          select 1
-          from information_schema.columns
-          where table_name='assessment_cycle' and column_name='published'
-        ) then
+        do $$
+        begin
```
sorja, line 63: Simplify the migration by removing double check for `published` column
We already check for `published` above, so no need for double check here.

It was required previously when running this under `public`-migration

minotogna (approved):

PR #3962, opened 2024-09-24
3949 - TablePaginated: Filter management + Admin users -> name - initial frontend

yaguzmang:
image


src/client/store/ui/tablePaginated/hooks/index.ts
```diff
@@ -7,6 +7,12 @@ export const useTablePaginatedCount = (path: string): { total: number } =>
 export const useTablePaginatedData = <Datum>(path: string): Array<Datum> | undefined =>
   useAppSelector((state) => TablePaginatedSelectors.getData(state, path) as Array<Datum> | undefined)
 
+export const useTablePaginatedFilters = (path: string) =>
+  useAppSelector((state) => TablePaginatedSelectors.getFilters(state, path))
+
+export const useTablePaginatedFilterValue = (path: string, fieldName: string) =>
+  useAppSelector((state) => TablePaginatedSelectors.getFilterValue(state, path, fieldName))
+
```
minotogna, line 22: Let's add return type to the hooks . In case use generic if needed

src/client/store/ui/tablePaginated/slice/index.ts
```diff
@@ -20,6 +24,10 @@ export const TablePaginatedSlice = createSlice({
       const { path, page } = action.payload
       Objects.setInPath({ obj: state, path: [path, 'page'], value: page })
     },
+    resetFilters: (state, action: PayloadAction<{ path: string }>) => {
```
minotogna: Where is this used ? Atm only one reset filter on text (x on top of text input)

sorja:
Looks good to me

src/client/store/ui/tablePaginated/hooks/index.ts
yaguzmang: Added those and a couple other here that were missing 👍

src/client/store/ui/tablePaginated/slice/index.ts
yaguzmang: I had misunderstood at first where you wanted the x. Now added it on the text input:
image

src/client/components/TablePaginated/Filters/Text/Text.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setLocalInputValue('')
+    dispatch(TablePaginatedActions.resetFilter({ fieldName, path }))
+  }
+
+  return (
+    <div className="table-paginated-filters__input-text-container">
+      <InputText
+        onChange={(e) => {
+          setLocalInputValue(e.target.value)
+          handleChangeThrottle(e)
```
minotogna: I think the throttle should not be handled here. the on change, should only update the filter value. 

The throttle will be done later, when filter updates throttled, then run query and count (this should be added where table data is fetched)

What do you think?

src/client/components/TablePaginated/types.ts
```diff
@@ -27,6 +28,12 @@ export type TablePaginatedCounter = {
   Component?: TablePaginatedCounterComponent
 }
 
+export type TablePaginatedFilter = {
```
minotogna, line 31: this maybe should go under meta ?

src/client/components/TablePaginated/Filters/Text/Text.tsx
yaguzmang: Yes, then the throttle can be handled in only one place :+1:
I will remove it here and add it in he next PR.

src/client/components/TablePaginated/types.ts
yaguzmang: This one has the "label" property and for now it's used only for the TablePaginated component in the FE. That's why I kept it in its types. If in the next PR I need it for the BE also, then I'll move it. What do you think?

minotogna: sure 👍

minotogna (approved):

PR #3969, opened 2024-09-25
Heroku Deploy fixes

minotogna:


package.json
```diff
@@ -270,6 +270,10 @@
     "webpack-node-externals": "^3.0.0",
     "winston": "^3.8.2"
   },
+  "overrides": {
```
yaguzmang, line 273: Does our yarn version support this? from the docs I see they support "resolutions":
https://classic.yarnpkg.com/en/docs/package-json#toc-resolutions

npm does support "overrides" though:
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides

yaguzmang (approved):

PR #3972, opened 2024-09-26
3967 - BulkDownload:getContentVariables->Get metadata from existing controller

### LINKED ISSUE #3967: BulkDownload:getContentVariables->Get metadata from existing controller
BulkDownload:getContentVariables->Get metadata from existing controller
@sorja we can address this in a separate issue ?

_Originally posted by @minotogna in https://github.com/openforis/fra-platform/pull/3965#discussion_r1775502033_

sorja:


minotogna (dismissed):

src/server/controller/cycleData/getBulkDownload/utils/getMetadata.ts
```diff
@@ -33,17 +33,29 @@ const getUnitLabelPath = (tableName: string, cycle: Cycle): string[] => {
...
+ *
+ * @returns {Promise<Table|undefined>} A promise that resolves to the table metadata or undefined if not found.
  */
-const getRowType = (tableName: string): RowType => {
-  const rowTypeMap: Record<string, RowType> = {
-    carbonStockSoilDepth: RowType.data,
-  }
 
-  return rowTypeMap[tableName] || RowType.header
+const getTable = async (props: Props): Promise<Table | undefined> => {
```
minotogna: controllers cannot use other controllers. that's our convention.
I'd rather push you to think generically , what's already in the platform? can this be needed genetically ?

In this case, I would suggest to use Redis repositories, and adding a TableRedisRepository.getOne(). internally you can use SectionRedisRespository.getSubsection

what do u think?

sorja: Sounds good, I will implement this

Thanks for reminding

sorja: So, in my implementation now I am making use of `SectionRedisRepository.getManyMetadata` as it is the lowest level in our Redis DB we can find the tables without knowing about the sectionName

The implementation is the same, we pass `{ assessment, cycle, tableName }` and we get `table: Table` returned from the func.

https://github.com/openforis/fra-platform/pull/3972/commits/8691d147a66abeace748990474b407931fddff6b

minotogna (approved):

PR #3976, opened 2024-09-26
3949 - TablePaginated: Filter management + Admin users -> name - Integration

yaguzmang:
https://github.com/user-attachments/assets/86a3babf-3f9c-4aa4-8d30-54ad73dc4f98

TODO:
 - [x] Export CSV taking into account filters and orderBys.
      - [x] Disable Export CSV button when count is 0.
 - [x] Add the logic to handle the filters which have a default value (In this case, the administrators filter). 
 - [x] Hide x button when filter input is empty.
 - [x] Make filters div sticky.
 - [x] Remove flickering of the table when typing


src/client/components/TablePaginated/Filters/Filters.tsx
```diff
@@ -7,12 +7,16 @@ import Icon from 'client/components/Icon'
 import Text from 'client/components/TablePaginated/Filters/Text/Text'
 import { TablePaginatedFilter } from 'client/components/TablePaginated/types'
 
-const componentsByFilterType: Record<TablePaginatedFilterType, React.FC<TablePaginatedFilter & { path: string }>> = {
+const componentsByFilterType: Record<
+  TablePaginatedFilterType,
+  React.FC<TablePaginatedFilter<TablePaginatedFilterType> & { path: string }>
+> = {
   [TablePaginatedFilterType.TEXT]: Text,
+  [TablePaginatedFilterType.SWITCH]: () => null,
```
yaguzmang, line 15: We can implement this once we need to add a visible boolean filter. For now it's not necessary because the "administrators" filter is hidden.

src/meta/api/request/admin/users.ts
```diff
@@ -10,11 +10,17 @@ export type UsersRequest<QueryParams = any, Body = any> = Request<
...
+    orderByDirection?: string
   }
 >
+
+export type DecodedUserFilters = {
+  administrators?: boolean
+  countries?: Array<CountryIso>
+  fullName?: string
+  roles?: Array<RoleName>
+}
```
yaguzmang: Any ideas what would be a better place for this type? `meta/user` didn't seem quite right to me since this type has to do with table paginated and users.

src/server/repository/public/user/count.ts
```diff
@@ -41,12 +45,28 @@ export const count = async (props: Props, client: BaseProtocol = DB): Promise<Re
...
+    roles,
+    statuses,
+  })
+
+  const queryTotals = `
+  select count(*) as total
+  from (
+      ${subQueryTotals}
+  ) as users;
+  `
```
yaguzmang, line 59: The query that calculates the `total` must be exactly the same as the one that gets the data (except for the limit and offset), otherwise any incongruences can result in the table paginator not working correctly. e.g it showing that there are 5 pages but in reality there are only 3. This happened to me, so I introduced this change.

src/client/components/TablePaginated/EmptyList/EmptyList.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+
+import { TablePaginatedEmptyListComponent } from 'client/components/TablePaginated/types'
+
+const DefaultEmptyList: React.FC = () => {
+  return <div />
+}
+
+type Props = {
+  Component?: TablePaginatedEmptyListComponent | undefined
```
minotogna: when using ? , | undefined is redundant

src/client/components/TablePaginated/EmptyList/EmptyList.tsx
```diff
@@ -0,0 +1,25 @@
...
+const EmptyList: React.FC<Props> = (props) => {
+  const { Component } = props
+
+  const EmptyListComponent = Component ?? DefaultEmptyList
+
+  return <EmptyListComponent />
+}
+
+EmptyList.defaultProps = {
+  Component: undefined,
```
minotogna: can the DefaultEmptyList be added to the default props instead ?

src/client/components/TablePaginated/Filters/Filters.tsx
```diff
@@ -23,6 +27,7 @@ const Filters: React.FC<Props> = (props: Props) => {
     <div className="table-paginated-filters-container">
       <Icon name="filter" />
       {filters.map((filter) => {
+        if (filter.isHidden) return null
```
minotogna: the parent component should render Filters if length of non hidden filers is > 0

src/client/components/TablePaginated/TablePaginated.tsx
```diff
@@ -77,8 +81,10 @@ const TablePaginated = <Datum extends object>(props: Props<Datum>) => {
 }
 
 TablePaginated.defaultProps = {
-  EmptyListComponent: () => <div />,
   counter: { show: true },
+  emptyList: {
+    showInTable: false,
+  },
```
minotogna: I'm not a big fan of this props. we should always render headers at this point. remove this prop and handle the empty list only in one way

src/client/components/TablePaginated/hooks/useFetchData.tsx
```diff
@@ -12,20 +17,56 @@ type Props = {
...
+        (newFilters: Record<string, TablePaginatedFilterValues>) => setFiltersThrottled(newFilters),
+        500,
+        { trailing: true }
+      ),
+    []
+  )
+
+  useEffect(() => {
+    updateThrottledFilters(filters)
+  }, [filters, updateThrottledFilters])
```
minotogna: the update of the filters should not be throttled. the fetch (on filers update) should be throttled

src/client/pages/Admin/UserManagement/UserManagement.tsx
```diff
@@ -19,6 +20,7 @@ const UserManagement: React.FC = () => {
     <TablePaginated
       columns={columns}
       counter={{ show: true, Component: UsersCount }}
+      emptyList={{ Component: UsersEmptyList, showInTable: true }}
```
minotogna: I think for now. let's not add this Users custom empty list, but customize the default table paginated  empty list component that says 'No items found'

src/client/pages/Admin/UserManagement/hooks/useFilters.tsx
```diff
@@ -5,15 +5,26 @@ import { TablePaginatedFilterType } from 'meta/tablePaginated'
...
       {
-        type: TablePaginatedFilterType.TEXT,
         fieldName: 'fullName',
         label: t('common.name'),
+        type: TablePaginatedFilterType.TEXT,
+      },
+      {
+        defaultValue: true,
+        fieldName: 'administrators',
+        isHidden: true,
```
minotogna: use the name hidden

src/client/store/ui/tablePaginated/actions/getCount.ts
```diff
@@ -3,23 +3,28 @@ import axios from 'axios'
...
 
 type Returned = {
   total: number
 }
 
 export const getCount = createAsyncThunk<Returned, Props>('tablePaginated/count/get', async (props) => {
-  const { assessmentName, cycleName, countryIso, sectionName, path } = props
+  const { assessmentName, countryIso, cycleName, filters, path, sectionName } = props
+
+  const encodedFilters = encodeFilters(filters)
```
minotogna: create utility TablePaginateds with methods encodeFilters and decodeFIlter<Returned>(string:string)=> Returned

src/client/store/ui/tablePaginated/actions/getData.ts
```diff
@@ -4,31 +4,36 @@ import axios from 'axios'
...
-  sectionName?: SectionName
+  cycleName: CycleName
+  filters?: Record<string, TablePaginatedFilterValues>
+  limit: number
   orderBy?: TablePaginatedOrderBy
   page: number
   path: string
-  limit: number
+  sectionName?: SectionName
 }
```
minotogna: check if you can extend TablePaginatedDataRequestParams

src/meta/api/request/admin/users.ts
minotogna: let's export this as meta/tablePaginated/user and call this just UsersFilter

src/server/repository/public/user/count.ts
```diff
@@ -41,12 +45,28 @@ export const count = async (props: Props, client: BaseProtocol = DB): Promise<Re
...
+  from counts`
+
+  const { query: subQueryTotals, queryParams: queryTotalsParams } = buildGetManyQuery({
+    administrators,
+    assessment,
+    countries,
+    cycle,
+    fullName,
+    roles,
+    statuses,
```
minotogna: can you pass directly props ?

src/client/store/ui/tablePaginated/actions/getData.ts
yaguzmang: I would have had to do something like this because some types weren't the same:
```
type Props = Omit<TablePaginatedDataRequestParams, 'limit' | 'filters' | 'offset' | 'orderBy' | 'orderByDirection'> & {
  filters?: Record<string, TablePaginatedFilterValues>
  limit: number
  orderBy?: TablePaginatedOrderBy
  page: number
  path: string
}
```
So I better exposed `TablePaginatedBaseParams` and used it here and in `getCount`. Let me know what you think.

src/client/components/TablePaginated/Filters/Filters.tsx
yaguzmang: Good catch 👍

minotogna (approved):
👍

PR #3978, opened 2024-09-27
3855 - Move files to s3 bucket

sorja:
- [x] check migration step is `exists` query
- [x] Check `public` migration OR `migration-step` to alter DDL correctly

implementation:
- create a bucket to s3
- implement new service `FileStorage` to write and read files
- remove column `file` from `public.file`

todo:
- [x] src/server/repository/public/ddl/getCreatePublicSchemaDDL.ts update `public.file` with public migration
- investigate if we want to remove `public.file`

todo/check:
- [x] upload and view avatar
- [x] upload and view file from Links and repository
- [x] upload and view file from EditorWYSIWYG


sorja:
- [  ] Write a script to test all files

src/server/api/file/get.ts
```diff
@@ -1,6 +1,6 @@
 import { Request, Response } from 'express'
 
-import { FileRepository, fileTypes } from 'server/service/file'
+import { FileRepository, fileTypes } from 'server/service/file_deprecated'
```
minotogna, line 3: how come we have some deprecated code?

src/server/controller/cycleData/repository/getManyFiles.ts
```diff
@@ -27,6 +30,12 @@ export const getManyFiles = async (props: Props): Promise<Returned> => {
   const repositoryProps = { fileUuids: repositoryItems.map((item) => item.fileUuid) }
   const files = await FileRepository.getMany(repositoryProps)
 
+  // eslint-disable-next-line no-restricted-syntax
```
minotogna: what about Promises.each ?

src/tools/uploadToS3/index.ts
```diff
@@ -0,0 +1,42 @@
...
+if (!BUCKET_NAME) {
+  Logger.error('S3_BUCKET_NAME must be defined in the .env file.\nCheck .env.template for example')
+  process.exit(1)
+}
+
+/**
+ * Uploads a file to an S3 bucket.
+ * @param filePath - The path to the file to upload.
+ */
+const uploadFileToS3 = async (filePath: string): Promise<void> => {
```
minotogna: do we need this script ? if yes, should we move it under tools/s3Files ?

src/server/api/file/get.ts
sorja: https://github.com/openforis/fra-platform/issues/4214

It's pending this - it's just serving static files

What do you suggest? Keep filename as `file`?

image

minotogna: Aaaa ok . Let's leave as it is then . We'll refactor later . No worries now

minotogna (changes requested):
Removing a repository item doesn't remove the file from the s3 bucket. it should.

minotogna (approved):

PR #3979, opened 2024-09-30
3931 - yarn migration-steps:reset

### LINKED ISSUE #3931: migrations-steps:reset
- add to package.json script migration-steps:reset that:
  - [ ] remove all migration-steps file
  - [ ] truncate migration-steps table 
  - [ ] reset pkey migration-steps table  ?

sorja:
Developers note:
Updated `create-migration-step.sh` to `createMigrationStep.ts`

--

- [ ] move migration to tools/ (another PR)
- [ ] execute first migration-steps reset (another PR)

---

- In this video we first inspect the `migrations.steps` table, showing our old/ran migrations
- Then we run the reset step
  - We check the git status showing the removed files
  - We check again `migrations.steps` table, showing only a single row for the new reset step

https://github.com/user-attachments/assets/113753a2-4d38-4ad2-9f4c-858347907f01


src/test/migrations/reset.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import { DB } from 'server/db'
+import { Logger } from 'server/utils/logger'
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+})
+
+const _confirmReset = (): Promise<boolean> => {
```
sorja: Added a user confirm:
After running command, user needs to write `reset` in the terminal to avoid running this by mistake (mainly to avoid resetting the db if the user didn't want to run this command)

minotogna: sure 👍

minotogna:
@sorja does this pr needs more work ?

sorja:
> @sorja does this pr needs more work ?

Yes, tryint to find a good way for userA and userB be in sync after migrations reset (see pr desc)

src/test/migrations/createMigrationStep.ts
```diff
@@ -0,0 +1,66 @@
...
+    Logger.info(`Created migration step ${filePath}`)
+  } catch (error) {
+    Logger.error('Error creating migration step:', error)
+    process.exit(1)
+  }
+
+  return { fileName, filePath }
+}
+
+if (require.main === module) {
```
sorja, line 63: Only execute the `createMigrationStep` when called from terminal (so when importing it doesn't automatically call the function)

src/test/migrations/utils/index.ts
```diff
@@ -0,0 +1,50 @@
+import * as fs from 'fs'
+import * as path from 'path'
+import * as readline from 'readline'
+
+/**
+ * Get user input from the terminal
+ * @param question - The question to ask the user
+ * @returns The user input
+ */
+export const getUserInput = (question: string): Promise<string> => {
```
minotogna, line 10: I suggest to forgot about test folder for the tools .

Also , I suggest to use our utility conventions (object that ends with s and have static methods )

What do you think ?

minotogna (approved):

PR #3984, opened 2024-09-30
3949 - TablePaginated: Filter management + Admin users -> name - Export with Hidden filters

yaguzmang:
https://github.com/user-attachments/assets/d3b92c7a-ff13-46c9-b1b0-1b99fee4fa9a


src/client/components/TablePaginated/hooks/useFilterValues.tsx
```diff
@@ -0,0 +1,35 @@
...
+      return acc
+    }, {} as Returned)
+  }, [filters])
+
+  const allFilterValues = useMemo<Returned>(() => {
+    return { ...hiddenFilterValues, ...activeFilterValues }
+  }, [activeFilterValues, hiddenFilterValues])
+
+  return allFilterValues
+}
```
yaguzmang: The last thing missing would be to handle filters that are not hidden and have a default value (like having a prefilled name for the text input, for instance). However, that's not necessary for this issue.

src/client/components/TablePaginated/TablePaginated.tsx
```diff
@@ -61,7 +61,7 @@ const TablePaginated = <Datum extends object>(props: Props<Datum>) => {
       <div>
         {(exportTable || withFilters) && (
           <div className="table-paginated-actions">
-            {exportTable && <ExportButton path={path} />}
+            {exportTable && <ExportButton counterActive={counter.show} filters={filters} path={path} />}
```
minotogna: Why should counter.show matter here ? The export csv can be disabled if there's no data ?

src/client/components/TablePaginated/hooks/useFilterValues.tsx
minotogna: Aren't all filters model stored in redux ? Making this hook redundant ?

minotogna (commented):
Is the links and repository page working ?

src/client/components/TablePaginated/TablePaginated.tsx
yaguzmang: I can do it by just looking if data is empty, but it will flicker a bit since we immediately reset the data before every fetch: 

https://github.com/user-attachments/assets/79deabd9-6897-40e0-bba5-30fa6dd9fe26

I can add the change if that's fine

src/client/components/TablePaginated/hooks/useFilterValues.tsx
yaguzmang: The hidden filters are not stored in redux. My first approach to handle those was setting them in redux but it would cause a re fetch. Something like this:

https://github.com/user-attachments/assets/648cecb5-f23d-40f6-a65f-8ee2c59bbda8

So I decided to keep them out of the store. And also because those don't have inputs.

We can keep them out or we could add logic that prevents that refetch. Either works. What do you think?

src/client/components/TablePaginated/TablePaginated.tsx
yaguzmang: And yes, the repository page is working. I also checked recent activity, history, and the other places TablePaginated is used. 
image

src/client/components/TablePaginated/hooks/useFilterValues.tsx
minotogna: I think we should have a consistent way of storing models. if filters are in redux, then all filter values must be in redux. one source of the data is important. otherwise it will be impossible to know where is the source of the filters.

src/client/components/TablePaginated/TablePaginated.tsx
minotogna: > I can do it by just looking if data is empty, but it will flicker a bit since we immediately reset the data before every fetch:
> 
>  2024-10-01.08-18-58.mp4 
> 
> I can add the change if that's fine

yes ok, but what about data is empty and not fetching ?

yaguzmang: Okay, we don't have the loading status on redux so I'll add it and use it here 👍

minotogna: sure 👍

minotogna: actually, let's just enable the csv button if count?.total > 0. let's see later other improvements . would that work now?

yaguzmang: We don't fetch the count if `counter.show` is false, so the export would be disabled for tables that don't display the counter, as count would be undefined (that's why I passed the counterActive prop).

minotogna: for now it's ok. let's see later if and when needed . what do you think ?

yaguzmang: Sure 👍

src/client/components/TablePaginated/hooks/useSetHiddenFilters.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useAppDispatch } from 'client/store'
+import { TablePaginatedActions } from 'client/store/ui/tablePaginated'
+import { TablePaginatedFilter } from 'client/components/TablePaginated/types'
+
+type Props = {
+  filters: Array<TablePaginatedFilter<TablePaginatedFilterType>>
+  path: string
+}
+
+export const useSetHiddenFilters = (props: Props): boolean => {
```
minotogna: Can we have a TablePaginated.init method instead of this ? What do you think ?

yaguzmang: Sure, sounds good. You mean something like this:
```
export const useInitTablePaginated = ({ filters, path }: Props): void => {
  const isInitialized = useIsTablePaginatedInitialized(path)

  useEffect(() => {
    if (!isInitialized) {
      dispatch(TablePaginatedActions.init({ filters, path }))
    }
  }, [dispatch, filters, isInitialized, path])
}
```
And handle the other logic in `TablePaginatedActions.init`

minotogna: yes. but what do u mean by 'other logic' ?

yaguzmang: Added, much cleaner now, thanks 👍 
By other logic I meant setting the defaultValues in state.

minotogna (approved):

src/client/components/TablePaginated/hooks/useSetHiddenFilters.tsx
minotogna: Nice 👍

PR #4008, opened 2024-10-08
3988 -  Data Table: CSV export file name

### LINKED ISSUE #3988: Data Table: CSV export file name
Data table csv exported file name should be:
assessmentName-cycleName-[current name]-dateISOstring

Image

sorja:
image

### DIFF
--- a/src/utils/dates.ts
+++ b/src/utils/dates.ts
@@ -25,17 +25,12 @@ export const getRelativeDate = (rawDate: any, i18n: any) => {
   return i18n.t('time.aMomentAgo')
 }
 
-const getCurrentDateISOString = () => {
-  return new Date().toISOString().split('T')[0]
-}
-
 export const Dates = {
   addDays,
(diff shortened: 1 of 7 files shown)


src/utils/dates.ts
```diff
@@ -25,13 +25,18 @@ export const getRelativeDate = (rawDate: any, i18n: any) => {
   return i18n.t('time.aMomentAgo')
 }
 
+const getCurrentDateISOString = () => {
+  return new Date().toISOString().split('T')[0]
```
minotogna: This is not iso string . I wouldn't add this method , but use Dates.format ?

sorja: You are right, I updated the code here [59ba320](https://github.com/openforis/fra-platform/pull/4008/commits/59ba32007c31d582f675266c13ccdbb2bbaa56c5)

minotogna (approved):

PR #4015, opened 2024-10-10
4000 - Dashboard metadata migration to node ext

sorja:
- Metadata copied from `Overview/`
- Next pr will remove old metadata from `Overview/`
- migration metadata only change is updated imports and removed function call(s) (eg. util `unit()`)

kuva

### DIFF
--- a/src/meta/nodeExt/nodeExt.ts
+++ b/src/meta/nodeExt/nodeExt.ts
@@ -11,6 +11,7 @@ export enum NodeExtCellType {
 export enum NodeExtType {
   contact = 'contact',
   node = 'node',
+  dataQuery = 'data-query',
 }
 
 export type NodeExt<Props, Value extends NodeValue | null = NodeValue> = {
--- a/src/meta/nodeExt/nodeExt.ts
+++ b/src/meta/nodeExt/nodeExt.ts
@@ -11,7 +11,7 @@ export enum NodeExtCellType {
 export enum NodeExtType {
   contact = 'contact',
   node = 'node',
-  dataQuery = 'data-query',
+  dataQuery = 'dataQuery',
 }
 
 export type NodeExt<Props, Value extends NodeValue | null = NodeValue> = {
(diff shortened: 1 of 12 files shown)


src/meta/nodeExt/nodeExt.ts
```diff
@@ -11,6 +11,7 @@ export enum NodeExtCellType {
 export enum NodeExtType {
   contact = 'contact',
   node = 'node',
+  dataQuery = 'data-query',
```
minotogna: I would use camel case

minotogna (commented):
We need to identify that this data query is a dashboard . Maybe add some property in the data query value

src/test/migrations/steps/metadata/dashboard/naturallyRegeneratingForestArea/index.ts
```diff
@@ -0,0 +1,58 @@
...
+    {
+      variableName: 'naturalForestArea',
+      color: ChartColor.green,
+      unit: `unit.${Unit.haThousand}`,
+      label: { key: 'statisticalFactsheets.rowName.naturalForestArea' },
+    },
+    {
+      variableName: 'plantedForest',
+      color: ChartColor.forestPlanted,
+      unit: `unit.${Unit.haThousand}`,
```
minotogna, line 28: I just realized that the unit is not a unit , but a string . Why ?

sorja:
> Maybe add some property in the data query value

@minotogna 
I suggest:

type: `dataQuery/dashboard`

or

props: `{ type: daashboard }`

minotogna:
> > Maybe add some property in the data query value
> 
> 
> 
> @minotogna 
> 
> I suggest:
> 
> 
> 
> type: `dataQuery/dashboard`
> 
> 
> 
> or
> 
> 
> 
> props: `{ type: daashboard }`

@sorja I don't know . Just add a type dashboard then

src/test/migrations/steps/metadata/dashboard/naturallyRegeneratingForestArea/index.ts
sorja: In the component we translate the string directly:

```
unit={`${cell.unit ? `${t(cell.unit)}` : ''}`}
```

minotogna (approved):

PR #4018, opened 2024-10-10
4000 - Dashboard store

sorja:
image


src/server/repository/assessmentCycle/nodeExt/getManyDashboardItems.ts
```diff
@@ -13,7 +13,10 @@ export const getManyDashboardItems = async (
   const { assessment, cycle } = props
   const schemaCycle = Schemas.getNameCycle(assessment, cycle)
 
-  return client.many<DashboardItem<DashboardItemType>>(`select value from ${schemaCycle}.node_ext where type = $1`, [
-    NodeExtType.dashboard,
-  ])
+  const result = await client.one<{ value: Array<DashboardItem<DashboardItemType>> }>(
+    `select value from ${schemaCycle}.node_ext where type = $1`,
+    [NodeExtType.dashboard]
```
minotogna: You could potentially still leave return and the callback as third arg to the client.one call .
Reaaaallly minor .😅

sorja: This is actually much better comment than you might realise, as this is what I was trying to do - I was trying to find a good way to use only the return

Thanks, I will update

minotogna (approved):

PR #4020, opened 2024-10-11
4000 - Utils for object manipulation

sorja:
Introduce two new object utils:

- mergePartial
- getDiffAsPartialObject

These will be used in upcoming PRs


src/utils/objects/getDiffAsPartialObject.ts
```diff
@@ -0,0 +1,62 @@
+import { isEmpty } from './isEmpty'
+import { setInPath } from './setInPath'
+
+const _traverseAndCompare = <T extends Record<string, unknown>>(
+  a: unknown,
+  b: unknown,
+  path: Array<string> = [],
+  keysToIgnore: Array<string> = [],
```
sorja: for example, we might want to ignore the 'uuid' or 'rowId' as it is not relevant for us (in this specific case)

src/utils/objects/mergePartial.ts
```diff
@@ -0,0 +1,33 @@
...
+ * const resultObj = mergePartial(baseObj, partialObj);
+ * // resultObj: { a: 1, b: { c: 2, d: 3 } }
+ *
+ * @example
+ * const baseArr = [1, 2, { a: 3 }];
+ * const partialArr = [4, undefined, { b: 5 }];
+ * const resultArr = mergePartial(baseArr, partialArr);
+ * // resultArr: [4, 2, { a: 3, b: 5 }]
+ */
+export const mergePartial = (obj1: any, partial: any): any => {
```
sorja: Merge the output of eg. `getDiffAsPartialObject` to a base object.

Our use case is:
base object = array of country dashboard items
partial object = region level items (eg unit is different in some deeply nested row)

minotogna: What about lodash.merge ?

src/utils/objects/getDiffAsPartialObject.ts
minotogna: Can this help : https://gist.github.com/Yimiprod/7ee176597fef230d1451 ?
Shorter version maybe 🤔?

sorja: Seems to have problem with arrays, would keep current version
let me know

src/utils/objects/mergePartial.ts
sorja: Replaced with lodash.merge

Seems to work as expected

src/utils/objects/index.ts
```diff
@@ -7,24 +7,29 @@ import * as isFunction from 'lodash.isfunction'
...
 import { camelize } from './camelize'
+import { getDiffAsPartialObject } from './getDiffAsPartialObject'
 import { isEmpty } from './isEmpty'
 import { propertyOf } from './propertyOf'
 import { setInPath } from './setInPath'
 
 export const Objects = {
   camelize,
   cloneDeep,
+  getDiffAsPartialObject,
```
minotogna: Should we call this simply getDiffs ?

minotogna (approved):

PR #4026, opened 2024-10-14
2661 - Translation System - JSON files

yaguzmang:
Translation working as before:

https://github.com/user-attachments/assets/1a437aeb-9bd3-4bf0-9a55-5dd34345e829


src/i18n/resources/ar.js
```diff
@@ -1,22 +1,22 @@
...
+const fra = require('./ar/fra.json')
+const generalValidation = require('./ar/generalValidation.json')
+const geo = require('./ar/geo.json')
+const history = require('./ar/history.json')
+const landing = require('./ar/landing.json')
+const login = require('./ar/login.json')
+const nationalDataPoint = require('./ar/nationalDataPoint.json')
+const other = require('./ar/other.json')
+const statisticalFactsheets = require('./ar/statisticalFactsheets.json')
+const uc = require('./ar/uc.json')
```
minotogna: I wonder now, can we use import vs require and maybe convert the en , es... to ts files ?

yaguzmang: Yes, I needed to add `"resolveJsonModule": true` to tsconfig, and was able to convert the files to ts.
Now we can access the properties:
 
image

minotogna (approved):
👍

PR #4027, opened 2024-10-15
4000 - Remove unused files

sorja:
Concludes #4000


src/client/store/metadata/hooks/useGetDashboard.ts
```diff
@@ -4,16 +4,12 @@ import { useAppDispatch } from 'client/store'
...
   const dispatch = useAppDispatch()
   const { assessmentName, cycleName, countryIso } = useCountryRouteParams()
-  const dashboardItems = useDashboardItems()
 
   useEffect(() => {
-    if (!dashboardItems) {
-      dispatch(MetadataActions.getDashboard({ assessmentName, cycleName, countryIso }))
-    }
-  }, [assessmentName, cycleName, countryIso, dispatch, dashboardItems])
+    dispatch(MetadataActions.getDashboard({ assessmentName, cycleName, countryIso }))
```
sorja: Only change in PR:
Refetch items (eg. changing country -> region -> country)

minotogna: If they aren't in store ? Let's fetch them once only ?

sorja: so i will introduce .region and .country ?

minotogna: yes

sorja: https://github.com/openforis/fra-platform/pull/4027/commits/e60d9187d1eac1dab9faa46021a1eb5a607077c3

introduced here

this can later be expanded very easily to

[AreaCode]: dashboarditems

So we can think about each country / area having their own dashboard :tada:

src/client/store/metadata/extraReducers/getDashboardReducer.ts
```diff
@@ -1,14 +1,23 @@
...
     }
-    state.dashboard[assessmentName][cycleName] = action.payload
+
+    if (!state.dashboard[assessmentName][cycleName]) {
+      state.dashboard[assessmentName][cycleName] = { region: undefined, country: undefined }
+    }
+
+    const key = Areas.isISOCountry(countryIso) ? 'country' : 'region'
+
+    state.dashboard[assessmentName][cycleName][key] = action.payload
```
minotogna: To avoid the if above , even Objects.set would work ?

src/client/store/metadata/extraReducers/getDashboardReducer.ts
```diff
@@ -1,14 +1,23 @@
...
     if (!state.dashboard[assessmentName]) {
       state.dashboard[assessmentName] = {}
     }
-    state.dashboard[assessmentName][cycleName] = action.payload
+
+    if (!state.dashboard[assessmentName][cycleName]) {
+      state.dashboard[assessmentName][cycleName] = { region: undefined, country: undefined }
+    }
+
+    const key = Areas.isISOCountry(countryIso) ? 'country' : 'region'
```
minotogna: Better to have an enum for those two keys ?

src/client/store/metadata/state.ts
```diff
@@ -2,10 +2,15 @@ import { AssessmentName, CycleName, Section, TableSection } from 'meta/assessmen
 import { SectionName } from 'meta/assessment/section'
 import { DashboardItem } from 'meta/dashboard'
 
+type DashboardState = Record<
+  AssessmentName,
+  Record<CycleName, { region: Array<DashboardItem>; country: Array<DashboardItem> }>
```
minotogna: the values for region and country are optional ?

src/client/store/metadata/hooks/useDashboardItems.ts
```diff
@@ -1,10 +1,12 @@
...
 import { useAppSelector } from 'client/store/store'
-import { useCycleRouteParams } from 'client/hooks/useRouteParams'
+import { useCountryRouteParams } from 'client/hooks/useRouteParams'
 
 export const useDashboardItems = (): Array<DashboardItem> => {
-  const { assessmentName, cycleName } = useCycleRouteParams()
-  return useAppSelector((state) => MetadataSelectors.getDashboard(state, assessmentName, cycleName))
+  const { assessmentName, cycleName, countryIso } = useCountryRouteParams()
+  const key = Areas.isISOCountry(countryIso) ? 'country' : 'region'
+  return useAppSelector((state) => MetadataSelectors.getDashboard(state, assessmentName, cycleName)?.[key])
```
minotogna: The key , can be passed to the selector ?

minotogna (approved):
👍

PR #4031, opened 2024-10-15
2661 - Translation System - Crowdin integration

yaguzmang:
Adding crowdin workflows and sorting the translation keys in all files except area.json, where we have the regions at the top. The reason to sort the keys is that once crowdin is connected and creates a PR, it sorts the resulting keys in the translated files based on the order of the source file. So, for consistency, I am sorting in this PR.


src/i18n/resources/ar/other.json
```diff
@@ -1,8 +1,4 @@
 {
-  "fileDrop": {
-    "dropFilesHere": "ضع الملف هنا",
-    "dragAndDropOrClick": "اسحب و ضع الملفات هنا, او اضغط لاختيار الملف "
-  },
   "anchors": {
     "fra": {
       "2025": {
```
yaguzmang: As a note, we have a special case here. This "anchors" property is only defined for Arabic. Crowdin only exports properties that are present in the source language file, all other properties are cleaned up. Since `en/other.json` doesn't have it, Crowdin would delete it for `ar/other.json` . To avoid that, I will extract "anchors" into its own file for Arabic only. 

This is an example of the Crowdin PR in the dummy project I created to test:
image

.github/workflows/translations-download-workflow.yml
```diff
@@ -0,0 +1,34 @@
+name: New Crowdin translations PR
+
+on:
+  workflow_dispatch:
```
yaguzmang, line 4: When triggered, this workflow creates a PR with the current approved translations in Crowdin. 

Unfortunately I wasn't able to add the automatic workflow we wanted to set up
(New translations pushed to dev -> Crowdin automatically translates -> A PR is created automatically)
because the feature that triggers translations when the source files are updated is behind a paywall, so I wasn't able to even test it:

image

image


Minimum plan to enable it would be "Team".
image

Same for webhooks:
image

src/i18n/resources/ar/other.json
minotogna: Is it possible we don't use this anchor ? Maybe it should be removed from this file ?

.github/workflows/translations-download-workflow.yml
minotogna: That's fine . But if it works when we auto translate from crowdin , then it's fine . 
@yaguzmang @sorja something to keep in mind

minotogna (approved):
🤩

PR #4034, opened 2024-10-16
4033 - NationalData-DataSources Cycle 2029: data/metadata clone

minotogna:
pr 1/2

- change to note: Objects.isEmpty + unit test

```typescript
export const isEmpty = (value: any): boolean =>
  isNil(value) ||
  Number.isNaN(value) ||
  (typeof value === 'object' && Object.keys(value).length === 0) ||
  (typeof value === 'string' && value.trim().length === 0)
```

image

### 2020
image

### 2025
image

### 2029
image

### DIFF
--- a/src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/DataSources.tsx
+++ b/src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/DataSources.tsx
@@ -41,7 +41,7 @@ export const DataSources: React.FC<Props> = (props: Props) => {
   const editable = useIsDescriptionEditable({ sectionName, name })
   const { empty } = useDescriptionErrorState({ name, sectionName })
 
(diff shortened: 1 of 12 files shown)


sorja:
image

Noticed this, is this still relevant?

minotogna:
> image
> 
> Noticed this, is this still relevant?

it's because of the toString of these two cases:
Screenshot 2024-10-16 at 11 38 05

sorja:
> > image
> > Noticed this, is this still relevant?
> 
> it's because of the toString of these two cases: Screenshot 2024-10-16 at 11 38 05

It is possible to pass formatter, eg:

```
const stringify = (value) => JSON.stringify(value,   function(k, v) { return v === undefined ? 'undefined' : v; }) 
```

sorja (approved):

PR #4038, opened 2024-10-17
4002 - Tablepaginated filter by user role/admin users role - Integration

yaguzmang:
This PR is just the integration of the new filter type. It uses the current `Select` component. The new generic  `MultiSelect` component will be in the next PR. 

(Updated)

https://github.com/user-attachments/assets/81796e0d-da59-4818-975d-e3624850eecc

### DIFF
--- a/src/client/pages/Admin/UserManagement/hooks/useColumns.tsx
+++ b/src/client/pages/Admin/UserManagement/hooks/useColumns.tsx
@@ -1,38 +1,23 @@
 import React, { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { Objects } from 'utils/objects'
-
 import { RoleName, User, Users } from 'meta/user'
 
-import { useTablePaginatedFilterValue } from 'client/store/ui/tablePaginated/hooks'
 import { Column } from 'client/components/TablePaginated'
 import NameField from 'client/pages/Admin/UserManagement/NameField'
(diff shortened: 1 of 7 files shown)


src/client/pages/Admin/UserManagement/hooks/useColumns.tsx
```diff
@@ -1,21 +1,38 @@
...
+  const { path } = props
   const { t } = useTranslation()
 
-  return useMemo<Array<Column<User>>>(() => {
-    const roleColumns: Array<Column<User>> = Object.values(RoleName).map((roleName) => ({
-      component: ({ datum }) => <RoleField roleName={roleName} user={datum} />,
-      header: t(Users.getI18nRoleLabelKey(roleName)),
-      key: roleName,
-    }))
+  const selectedRoles = useTablePaginatedFilterValue<Array<string>>(path, 'roles')
```
minotogna: Why this selectedRoles ? Should we include in the options just the role independently to the selection ?

minotogna:
@yaguzmang the video doesn't show when there are no roles selected. can u post a comprehensive video please ?

src/client/pages/Admin/UserManagement/hooks/useColumns.tsx
yaguzmang: The previous implementation had this feature, so I thought we might to do something similar. I will remove it though 👍

minotogna (approved):

PR #4039, opened 2024-10-17
4033 - Dashboard cycle clone

sorja:


src/client/pages/CountryHome/FraHome/hooks/useSections.tsx
```diff
@@ -30,8 +30,8 @@ export const useSections = (): Array<Section> => {
 
     if (!cycle) return null
 
-    const isFra2025 = assessment.props.name === AssessmentNames.fra && cycle.name === '2025'
-    const showOverview = !isFra2025 || Areas.isISOCountry(countryIso)
+    const isFra2020 = assessment.props.name === AssessmentNames.fra && cycle.name === '2020'
+    const showOverview = isFra2020 || Areas.isISOCountry(countryIso)
```
sorja, line 35: Not sure if there is (now) a better way to fix this

Maybe cycle.published and FRA?

But it's not directly related to publishing of cycles, it's more related IF we have faoEstimates in db

minotogna: We don't know at the moment.
These data sent to us , just for the dashboard, is very tricky to handle .
Let's see later on if we find a solution

sorja: Sounds good, I added proposal here so we can keep track/remember:
https://github.com/orgs/openforis/projects/8/views/1?pane=issue&itemId=83732694

minotogna (approved):

PR #4049, opened 2024-10-18
4002 - TablePaginated: Filter by User Role + Admin users 'Role' - Multiselect UI

yaguzmang:
https://github.com/user-attachments/assets/93119499-5652-4cb9-bea9-07aa4ec3d041

(After scss updates based on comments:)
image


src/client/components/Inputs/Select/Indicators/MultiValue/MultiValue.tsx
```diff
@@ -0,0 +1,24 @@
+import './MultiValue.scss'
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { MultiValueProps as OriginalMultiValueProps } from 'react-select'
+
+import { SelectProps } from 'client/components/Inputs/Select/types'
+
+type MultiValueProps = OriginalMultiValueProps & Pick<SelectProps, 'multiLabelKey'>
+
+export const MultiValue: React.FC<MultiValueProps> = (props) => {
```
minotogna: I would call this MultiValueSummary ? (if in the future we'll need to customize the MultiValue). what do u think?

src/client/components/Inputs/Select/hooks/useComponents.tsx
```diff
@@ -22,7 +25,13 @@ export const useComponents = (props: SelectProps): Returned => {
       IndicatorSeparator: null,
     }
     if (isMulti) components.Option = MultiSelectOption
+    if (isMulti && !Objects.isEmpty(multiLabelKey)) {
```
minotogna: I would call this prop multiLabelSummaryKey

src/client/components/Inputs/Select/hooks/useComponents.tsx
```diff
@@ -1,18 +1,21 @@
-import { useMemo } from 'react'
-import { Props as ReactSelectProps } from 'react-select'
+import React, { useMemo } from 'react'
+import { MultiValueProps as OriginalMultiValueProps, Props as ReactSelectProps } from 'react-select'
```
minotogna: is there a real need to override the name MultiValueProps ? if not, I would leave it as MultiValueProps

src/client/components/TablePaginated/Filters/Filters.scss
```diff
@@ -4,6 +4,8 @@
   align-items: center;
   display: flex;
   gap: $spacing-xxs;
+  height: $spacing-m;
+  padding: $spacing-xxxxs 0;
```
minotogna: why the vertical padding ? aren't the items vertically centered?

src/client/components/TablePaginated/Filters/MultiSelect/MultiSelect.scss
```diff
@@ -0,0 +1,57 @@
...
+div.filter-multiselect__container {
+  border-radius: 2px;
+  border: 1px solid $ui-border;
+  font-size: $font-xxs;
+  min-width: #{3 * $spacing-xl};
+  padding: 0;
+  transition: all 0.2s ease;
+
+  div.select__control {
+    min-height: $spacing-s;
```
minotogna: why the min-height ? should we use flex and vertically center ?

src/client/components/TablePaginated/Filters/MultiSelect/MultiSelect.scss
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        &:hover {
+          background-color: unset;
+
+          svg.icon {
+            color: $ui-destructive;
+          }
+        }
+      }
```
minotogna, line 39: should the select clear indicator style be applied generically to all select instances ?
What do u think @sorja @yaguzmang ?

src/client/components/TablePaginated/Filters/MultiSelect/MultiSelect.scss
```diff
@@ -0,0 +1,57 @@
...
+      width: $icon-size;
+    }
+  }
+
+  &.active {
+    border: 1px solid $ui-accent;
+  }
+
+  &.select__container {
+    height: calc(100% + 1px);
```
minotogna: this is strange. why the + 1px ?

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
```diff
@@ -0,0 +1,58 @@
...
+export const useTooltipContent = (props: Props): Returned => {
+  const { fieldName, options, path } = props
+  const filterValue = useTablePaginatedFilterValue<Array<string>>(path, fieldName)
+  const [canDisplayTooltip, setCanDisplayTooltip] = useState(true)
+
+  const valueToLabelMap = useMemo(() => {
+    return options.reduce((acc, { value, label }) => {
+      // eslint-disable-next-line no-param-reassign
+      acc[value] = label
+      return acc
```
minotogna: to avoid eslint here, return {...acc,[value]:label}

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
```diff
@@ -0,0 +1,58 @@
...
+    }, {} as Record<string, string>)
```
minotogna: u don't need the cast, add the type to useMemo and reduce

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
```diff
@@ -0,0 +1,58 @@
...
+  }, [options])
+
+  const tooltipContent = useMemo(() => {
```
minotogna: add the type

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
```diff
@@ -0,0 +1,58 @@
...
+    return selectedLabels.join(', ')
+  }, [filterValue, valueToLabelMap])
+
+  const hideTooltip = useCallback(() => setCanDisplayTooltip(false), [])
+  const showTooltip = useCallback(() => setCanDisplayTooltip(true), [])
+
+  return {
+    hideTooltip,
+    showTooltip,
+    tooltipContent: canDisplayTooltip ? tooltipContent : null,
```
minotogna: you can do this in the useMemo

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
minotogna: and I would remove the useMemo above, you can declare the variable within this useMemo

src/client/components/TablePaginated/Filters/Text/Text.scss
```diff
@@ -35,7 +35,7 @@
     }
 
     &.disabled {
-      opacity: 0.6;
+      opacity: 0;
```
minotogna: I wouldn't do this. I would just not render the X

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const useTooltipContent = (props: Props): Returned => {
+  const { fieldName, options, path } = props
+  const filterValue = useTablePaginatedFilterValue<Array<string>>(path, fieldName)
+  const [canDisplayTooltip, setCanDisplayTooltip] = useState(true)
+
+  const valueToLabelMap = useMemo(() => {
+    return options.reduce((acc, { value, label }) => {
+      // eslint-disable-next-line no-param-reassign
+      acc[value] = label
```
sorja: is it possible to add in eslint a rule where we can assign in certain functions eg reduce?

minotogna: We should spread in this case .

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
yaguzmang: I separated the `valueToLabelMap` because it doesn't depend on the filter selected values. It is calculated based on the select options. If I declare it in this memo, it will be recalculated every time the user changes selection. That's not an issue for this specific filter since it only has a few items, but should we keep it separate in case we have larger selection lists in the future? like countries, although that one will probably be in its own component.

src/client/components/TablePaginated/Filters/MultiSelect/MultiSelect.scss
yaguzmang: It would look like this:

https://github.com/user-attachments/assets/ecd33085-f324-4c6a-bc31-d79e1b5dd23b

If we add it then we should probably change the red color of x button on the selected options, for consistency. 
image



Current for reference:

https://github.com/user-attachments/assets/03e6b50d-38db-4760-8a3a-a690d9f2f489

Personally, in the selects that are in tables, I like the extra red squared background color. And since clicking it in a table would affect data, it makes sense that the "x" is red by default.

minotogna: ok!

minotogna: lets leave it as it is

src/client/components/TablePaginated/Filters/MultiSelect/hooks/useTooltipContent.ts
minotogna: ok

src/client/components/TablePaginated/Filters/Filters.scss
yaguzmang: It was there to get the inputs to be the same height as before, but now just changed the height directly to 30px.

src/client/components/TablePaginated/Filters/MultiSelect/MultiSelect.scss
yaguzmang: The default control styles have a min-height set (end of list): 
image
So I changed now ` min-height: $spacing-s;` to `min-height: unset;` as a specific value for it is not necessary 👍

minotogna (approved):

PR #4052, opened 2024-10-23
4042 - TablePaginated: Filter by Country + Admin users 'Country' - Initial version

yaguzmang:
Updated vid:

https://github.com/user-attachments/assets/d4fb3082-2a46-46e2-8b0b-39274c354321

TODO:
- [x] Integrate CountryMultiSelect with TablePaginated filters
- [ ] (Next PR) Add group selection.

### DIFF
--- a/src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
+++ b/src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
@@ -1,29 +1,23 @@
 import React from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { Country } from 'meta/area'
-
 import Select, { SelectProps } from 'client/components/Inputs/Select'
 
 import { useCountriesByRegionOptions } from './hooks/useCountriesByRegionOptions'
 
-type Props = Omit<SelectProps, 'options'> & {
-  countries: Array<Country>
-  excludedRegions: Array<string>
-}
+type Props = Omit<SelectProps, 'options'>
 
 const CountryMultiSelect: React.FC<Props> = (props: Props) => {
-  const { countries, excludedRegions, ...selectProps } = props
-  const { multiLabelSummaryKey, placeholder } = selectProps
+  const { multiLabelSummaryKey, placeholder } = props
 
   const { t } = useTranslation()
 
-  const optionGroups = useCountriesByRegionOptions({ countries, excludedRegions })
+  const optionGroups = useCountriesByRegionOptions()
 
   return (
     <Select
       // eslint-disable-next-line react/jsx-props-no-spreading
-      {...selectProps}
+      {...props}
       multiLabelSummaryKey={multiLabelSummaryKey ?? 'admin.country'}
       options={optionGroups}
       placeholder={placeholder ?? t('common.countries')}
--- a/src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
+++ b/src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
@@ -1,0 +1,34 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+
+import { Country } from 'meta/area'
+
+import Select, { SelectProps } from 'client/components/Inputs/Select'
+
+import { useCountriesByRegionOptions } from './hooks/useCountriesByRegionOptions'
+
+type Props = Omit<SelectProps, 'options'> & {
+  countries: Array<Country>
+  excludedRegions: Array<string>
+}
+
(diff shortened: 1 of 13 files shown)


src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
```diff
@@ -0,0 +1,34 @@
...
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      {...selectProps}
+      multiLabelSummaryKey={multiLabelSummaryKey ?? 'admin.country'}
+      options={optionGroups}
+      placeholder={placeholder ?? t('common.countries')}
+    />
+  )
+}
+
+export default CountryMultiSelect
```
yaguzmang, line 34: To be able to use this generic component as a filter I see two options:
1. Adding a new filter type "custom" that allows passing your own filter component which handles the state changes. The drawback of it is that for example in this case, the logic to handle the tooltip would need to be repeated, as that logic is currently part of the filter -> MultiSelect.

2. Keep using MultiSelect and allow passing a custom component to it. I tried this option but passing the props `{
  countries: Array<Country>
  excludedRegions: Array<string>
}` to the CountryMultiSelect made the ts typing too convoluted. 

Any ideas for a clean way to integrate this to the table paginated filters? @sorja @minotogna

src/client/components/TablePaginated/Filters/Country/hooks/useTooltipContent.ts
```diff
@@ -0,0 +1,52 @@
...
+        return t(`area.${countryIso}.listName`)
+      })
+      .sort((a, b) => a.localeCompare(b))
+      .slice(0, MAX_VISIBLE_COUNTRY_LABELS)
+
+    if (filterValue.length > MAX_VISIBLE_COUNTRY_LABELS) {
+      selectedCountryLabels.push('...')
+    }
+
+    return selectedCountryLabels.join(', ')
```
yaguzmang, line 41: Possible improvements for the countries tooltip:
1. If all countries in a region are selected, then display the region name instead of its selected countries.
2. Instead of showing "..." at the end when too many are selected, show "+{{remainingCount}} more". For example: "... Somalia, South Africa, +78 more."

Let me know and I can include either in the next PR.

src/client/components/Inputs/Select/Indicators/MultiValueSummary/MultiValueSummary.tsx
```diff
@@ -3,18 +3,18 @@ import React from 'react'
...
 
 export const MultiValueSummary: React.FC<MultiValueProps> = (props) => {
   const { getValue, index, multiLabelSummaryKey } = props
-  const count = getValue().length
   const { t } = useTranslation()
 
   const displayCountLabel = index === 0
 
   if (displayCountLabel) {
+    const count = Array.from(new Set(getValue().map((v) => (v as Option).value))).length
```
minotogna: Is this necessary ? Isn't vale an array of selected options already ?

src/client/components/TablePaginated/Filters/Country/hooks/useTooltipContent.ts
minotogna: Yes , both are valid improvements. Next pr is fine .

Also , the regions should be collapsed by default , and have an arrow to expand and see all regions .
Also , the regions should have a checkbox that toggle the selection of their countries .
And the font should be bold , to highlight that it's a group .
A separate pr just for this would work in case .
What do you think ?

src/client/components/Inputs/Select/hooks/useOnChange.ts
```diff
@@ -23,13 +23,15 @@ export const useOnChange = (props: SelectProps): Returned => {
 
         // If "Select All" is toggled with no selection, select all original options
         if (selectedValues.length === 1) {
-          const allValues = selectOptions.flatMap((optionOrGroup) => {
+          const uniqueValues = new Set<string>()
```
minotogna: Should we assume an option exists only once ?

yaguzmang: Yes. I removed the changes to handle that case now :+1:

src/client/components/TablePaginated/Filters/Country/hooks/useTooltipContent.ts
yaguzmang: Yes, I agree a separate PR is better for this 👍

minotogna (approved):

PR #4054, opened 2024-10-23
4032 - UserInvitation and UserRole types and integration test

sorja:
Build issues etc can be ignored in this PR

Add unit test for getLastRole:

image

### DIFF
--- a/src/meta/user/userInvitation.ts
+++ b/src/meta/user/userInvitation.ts
@@ -4,7 +4,7 @@ import { RoleName } from 'meta/user/userRole'
 export interface UserInvitation {
   acceptedAt?: string
   assessmentUuid: string
-  countryIso?: CountryIso
+  countryIso: CountryIso
   cycleUuid: string
   id: number
   invitedAt: string
--- a/src/meta/user/userInvitation.ts
+++ b/src/meta/user/userInvitation.ts
@@ -3,7 +3,7 @@ import { RoleName } from 'meta/user/userRole'
 
 export interface UserInvitation {
   acceptedAt?: string
-  assessmentUuid?: string
+  assessmentUuid: string
   countryIso?: CountryIso
   cycleUuid: string
   id: number
--- a/src/meta/user/userInvitation.ts
+++ b/src/meta/user/userInvitation.ts
@@ -1,0 +1,16 @@
+import { CountryIso } from 'meta/area'
+import { RoleName } from 'meta/user/userRole'
+
+export interface UserInvitation {
+  acceptedAt?: string
+  assessmentUuid?: string
+  countryIso?: CountryIso
+  cycleUuid: string
+  id: number
+  invitedAt: string
+  invitedByUserUuid?: string
+  props: never // default {}
+  role: RoleName
+  userUuid: string
+  uuid: string
+}
--- a/src/meta/user/userInvitation.ts
+++ b/src/meta/user/userInvitation.ts
@@ -8,7 +8,7 @@ export interface UserInvitation {
   cycleUuid: string
   id: number
   invitedAt: string
-  invitedByUserUuid?: string
+  invitedByUserUuid: string
   props: never // default {}
   role: RoleName
   userUuid: string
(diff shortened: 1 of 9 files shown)


minotogna (changes requested):
the comment in the previous PR should be updated.

src/meta/user/userRoles.ts
```diff
@@ -50,10 +43,10 @@ const getLastRole = (params: { assessment?: Assessment; user: User }) => {
...
 
   const _roles = [...roles].sort((roleA, roleB) => {
-    if (!roleA.acceptedAt && !roleB.acceptedAt) return 0
-    if (!roleA.acceptedAt) return 1
-    if (!roleB.acceptedAt) return -1
-    return roleB.acceptedAt.localeCompare(roleA.acceptedAt)
+    if (!roleA.createdAt && !roleB.createdAt) return 0
+    if (!roleA.createdAt) return 1
+    if (!roleB.createdAt) return -1
+    return roleB.createdAt.localeCompare(roleA.createdAt)
```
minotogna: should we use Dates instead ?

src/meta/user/userInvitations.ts
```diff
@@ -0,0 +1,8 @@
+import { UserInvitation } from 'meta/user/userInvitation'
+
+export const isInvitationExpired = (invitation: UserInvitation, expiryPeriod?: number) =>
+  new Date().getTime() - Date.parse(invitation.invitedAt) > (expiryPeriod || 7) * 86400000
+
+export const UserInvitations = {
+  isInvitationExpired,
```
minotogna: I would not duplicate the name invitation in the methods. this utility works on invitations objects only

src/meta/user/userInvitation.ts
```diff
@@ -0,0 +1,16 @@
+import { CountryIso } from 'meta/area'
+import { RoleName } from 'meta/user/userRole'
+
+export interface UserInvitation {
+  acceptedAt?: string
+  assessmentUuid?: string
+  countryIso?: CountryIso
+  cycleUuid: string
```
minotogna: how come cycle is required and assessment not ?

src/meta/user/userRole.ts
```diff
@@ -53,18 +53,17 @@ export type UserRoleExtendedProps = UserRoleBaseProps & {
 }
 
 export interface UserRole<Name extends RoleName, Props extends UserRoleBaseProps = undefined, Permissions = undefined> {
-  id: number
   assessmentId?: number
-  cycleUuid: string
   countryIso?: CountryIso
-  invitedByUserUuid?: string
+  createdAt: string
+  cycleUuid: string
```
minotogna: same here... and are we still using assessment id ? or uuid ?

sorja: Typo, also cycle should be optional (admin)

src/meta/user/userInvitation.ts
sorja: I will mark in the main PR that we should check if assessmentUuid is really needed as an invitation can be related only to a single cycle, same as roles

What do you think?

#discussion

minotogna: leave both, assessments and cycles for now. we'll talk in the near future about this. ok?

sorja: sure - added in main pr so we don't forget

also something to keep in mind in the discussion:

if role/invitation has either, it should maybe mean it's for all cycles (avoid role duplication?)

minotogna: an invitation is per cycle. same thing for the role. or did I misunderstood your comment ?

minotogna:
please check deepscan , the tests have some issues

src/meta/user/userInvitation.ts
```diff
@@ -0,0 +1,16 @@
...
+import { RoleName } from 'meta/user/userRole'
+
+export interface UserInvitation {
+  acceptedAt?: string
+  assessmentUuid: string
+  countryIso: CountryIso
+  cycleUuid: string
+  id: number
+  invitedAt: string
+  invitedByUserUuid?: string
```
minotogna: is it possible to have an invitation without the inviter ?

minotogna (approved):
👍

PR #4057, opened 2024-10-25
4032 - Accept and Remove invitation

sorja:
Purpose of PR:
- Fix building and tests (after this, tests and build should always pass per PR)
  -  image
- Update Accept and Remove invitation:
  - Remove them from under UserRole repository
  - Move them to UserInvitation
- Added some `// TODO` checks as reminders for frontend development
  - Eg. when removing an invitation, we used to return `UserRole`, now we return `UserInvitation`.


minotogna (approved):
🤩

PR #4066, opened 2024-10-29
4042 - TablePaginated: Filter by Country + Admin users 'Country' - Selectable and collapsible groups

yaguzmang:
image

FRA:

https://github.com/user-attachments/assets/c9a23307-1418-49ea-bf37-ca27caeb2367

panEuropean:

https://github.com/user-attachments/assets/cdb71551-1582-4249-b4f5-0630f4b89c53

rtl: 

https://github.com/user-attachments/assets/b5b68d3e-b3d6-4445-a532-7283da67bc77


src/client/components/Inputs/Select/Indicators/CollapsibleGroup/hooks/useExpandGroup.ts
```diff
@@ -0,0 +1,38 @@
...
+        setExpanded(true)
+        setManuallyExpanded(false)
+      }
+    } else if (!manuallyExpanded) {
+      // Collapse if not manually expanded
+      if (expanded) {
+        setExpanded(false)
+      }
+    }
+  }, [expanded, inputValue, manuallyExpanded])
```
yaguzmang, line 30: One caveat here is that when you manually open a region but the search string you type doesn't match any of its countries, react-select will internally remove this Group component (and its state). So, when the search is cleared and the component is rendered again, it has the default initial state. 

example:

https://github.com/user-attachments/assets/b5e4134a-b51e-4ef7-a56e-6ab43f1afef3

Unfortunately react-select doesn't have a prop or custom option to prevent this. One possible solution could be to add a Context provider to the CountrySelect and store the `expanded` state for each Group using a Group id.

minotogna: That is fine !! 😌

src/client/components/Inputs/Select/Indicators/CollapsibleGroup/hooks/useExpandGroup.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const toggleExpanded = useCallback(() => {
+    setExpanded((prev) => !prev)
+    setManuallyExpanded(true)
```
minotogna: When closing , manually expanded should be false ?

yaguzmang: It doesn't affect because `manuallyExpanded = false` only matters when expanded is true. 
Added it to the toggleExpanded for consistency though :+1:

minotogna: I see , nice 😊 thank you

minotogna (approved):

PR #4073, opened 2024-11-01
4032 - UI: User list export with invitations

sorja:


src/client/pages/CountryHome/Collaborators/UserList/Export/Export.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { useHeaders } from './hooks/useHeaders'
+
+const FILENAME = 'FRA-Users.csv'
+
+const Export: React.FC = () => {
+  const csvHeaders = useHeaders()
+  const data = useData()
+
+  return (
+    <CSVLink
```
minotogna, line 16: Do we have a generic csv link component ?

sorja: Do you mean the one in datatables?
Those export the table data, but in our case we export custom data from two tables

Let me know

minotogna: I meant just for the ui style .
A wrapper to CsvLink imported .

minotogna: If we don't , we'll think later on

sorja: Let's add a note for this, I did not find this

https://github.com/orgs/openforis/projects/8?pane=issue&itemId=85627575

src/client/pages/CountryHome/Collaborators/UserList/Export/Export.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { useData } from './hooks/useData'
+import { useHeaders } from './hooks/useHeaders'
+
+const FILENAME = 'FRA-Users.csv'
+
+const Export: React.FC = () => {
+  const csvHeaders = useHeaders()
+  const data = useData()
+
+  return (
```
minotogna, line 15: Wait . We'll need to use table paginated csv export generically.
We don't need it now this export

sorja: We have two tables, should we have two buttons?

minotogna: we'll do the export later, using generic db export. I would close this pr now

PR #4075, opened 2024-11-02
4042 - TablePaginated: Filter by Country + Admin users 'Country' - Tooltip

yaguzmang:
https://github.com/user-attachments/assets/87279371-22bc-466c-b2dc-e45dd7c71685

panEuropean

https://github.com/user-attachments/assets/ac7c9258-ce3a-4e4e-adee-858f8452e4fa

### DIFF
--- a/src/client/components/TablePaginated/Filters/Country/Country.scss
+++ b/src/client/components/TablePaginated/Filters/Country/Country.scss
@@ -3,3 +3,42 @@
 div.filter-country {
   width: 235px;
 }
+
+.tooltip-container {
+  .react-tooltip.filter-country__tooltip {
+    max-width: unset;
+
+    .regions-container {
+      column-gap: $spacing-xs;
+      display: grid;
+      font-size: $font-xs;
+      max-height: 500px;
+      max-width: 1200px;
+      overflow-y: scroll;
+      pointer-events: auto;
+      row-gap: $spacing-xxxs;
+
+      .region-title {
+        background-color: lighten($ui-accent, 55%);
+        font-weight: bold;
+        padding-bottom: $spacing-xxxs;
+        position: sticky;
+        top: 0;
+      }
+
+      .countries-container {
+        display: grid;
+        gap: $spacing-xxxs;
+        height: min-content;
+
+        .country {
+          height: $spacing-s;
+          max-width: 170px;
+          overflow: hidden;
+          text-overflow: ellipsis;
+          white-space: nowrap;
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


src/client/components/TablePaginated/Filters/Country/Country.scss
```diff
@@ -3,3 +3,42 @@
...
+      }
+
+      .countries-container {
+        display: grid;
+        gap: $spacing-xxxs;
+        height: min-content;
+
+        .country {
+          height: $spacing-s;
+          max-width: 170px;
```
yaguzmang, line 36: This makes the width more consistent when certain options are selected. 

For example, instead of this:

https://github.com/user-attachments/assets/fe4b184a-ce30-4a59-a10a-c9d8b685348e

It looks like this:


https://github.com/user-attachments/assets/5baf9f37-f23c-4eae-8962-1a4795b57cf7

src/client/components/TablePaginated/Filters/Country/Country.tsx
```diff
@@ -40,8 +40,11 @@ const Country: React.FC<Props> = (props: Props) => {
   return (
```
yaguzmang, line 40: @minotogna Currently in the TablePaginated we handle the redux state with the api `path`. 
image


When we switch cycles that `path` stays the same, so we use the same state in Redux for the same table but in different cycle. 
With that in mind, I noticed this when testing for this PR :

https://github.com/user-attachments/assets/a792f4c1-8437-476b-bc93-c5de02a311fd

Since the state to manage the table is the same, the filter values are kept. 

To fix it I already tried something like:
```
export const useResetFilter = (props: Props) => {
  const { fieldName, path } = props
  const { assessmentName, cycleName } = useCycleRouteParams()

  return useEffect(() => {
    dispatch(TablePaginatedActions.resetFilter({ fieldName, path }))
  }, [assessmentName, cycleName, dispatch, fieldName, path])
}
```
But that wouldn't really work as the filters would be reset also when the hook is mounted (e.g when navigating manage collaborators -> invitations -> manage collaborators).

Should we instead consider the assessment/cycle when we store the TablePaginated state?

Let me know what you think

minotogna: filters and data should be reset on unmount for every table paginated generically ?
What do you think @sorja @yaguzmang

yaguzmang: That would be the cleanest code-wise. Wouldn’t refetching the data every time it unmounts and then mounts again be a concern, like when navigating between tabs in the admin page?

yaguzmang: @minotogna @sorja Added `useResetOnUnmount` to TablePaginated. Let me know what you guys think

minotogna: That looks good to me . @sorja we need to keep this in mind . Gets reset on every unmount . Ok what do u think ?

minotogna (approved):

PR #4076, opened 2024-11-04
4032 -  Update country status change confirm pop up

sorja:


src/client/components/PageLayout/Toolbar/Status/StatusConfirm/UserList/UserList.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+interface Props {
+  users: Array<User>
+}
+
+interface CellProps {
+  user: User
+}
+
+const NameCell: React.FC<CellProps> = ({ user }) => <span>{Users.getFullName(user)}</span>
```
sorja, line 19: Is it preferred in this case separate file with `(props: Props)` for all Cell components?

minotogna (approved):

src/client/components/PageLayout/Toolbar/Status/StatusConfirm/UserList/UserList.tsx
minotogna: Let's see this at the end 😅

PR #4089, opened 2024-11-07
4032 - Update country summary

sorja:
image

With one pending user (new invitation/new user)


src/server/repository/assessmentCycle/countrySummary/createMaterializedView.ts
```diff
@@ -49,37 +49,56 @@ export const createMaterializedView = async (props: Props, client: BaseProtocol
...
+           user_invitations as (
+               select ui.country_iso,
+                      ui.user_uuid,
+                      ui.invited_at,
+                      ui.accepted_at
+               from public.users_invitation ui
+               where ui.cycle_uuid = '${cycle.uuid}'
+           ),
+           unique_users as (
+               select user_uuid, country_iso
```
minotogna, line 66: do we need distinct here? (since the same user can have both, role and invitation ?)

minotogna (approved):

PR #4090, opened 2024-11-07
Fix remove cycle (returning updated assessment) 

minotogna:


src/server/controller/cycleData/persistNodeValues/persistNode/persistNode.ts
```diff
@@ -21,10 +21,9 @@ export const persistNode = async (props: Props, client: BaseProtocol): Promise<N
   const { assessment, countryIso, activityLogMessage, user, cycle, sectionName } = props
   const node: Node = await NodeRepository.getOneOrNone(props, client)
 
-  const [nodeUpdated] = await Promise.all([
-    node ? NodeRepository.update(props, client) : NodeRepository.create(props, client),
-    DataRedisRepository.updateNode(props),
-  ])
+  const nodeUpdated = await (node ? NodeRepository.update(props, client) : NodeRepository.create(props, client))
```
sorja, line 24:
```suggestion
const upsert = node ? NodeRepository.update : NodeRepository.create
const nodeUpdated = await upsert(props, client)
```

sorja (approved):

src/server/controller/cycleData/persistNodeValues/persistNode/persistNode.ts
minotogna: We should have a real upsert method . We'll do this later on

PR #4097, opened 2024-11-08
migration steps reset: fix double insert

### LINKED ISSUE #4094: migration-steps reset - error
**Execute on review and production** 

Image

**by fixing the code, adding file at line 63 it fails anyway**

Image

sorja:
Changes:

main file (index.ts):
ignore files ending -step-reset (dont double insert the row in db `steps`)

reset template and existing reset step:
use `path` and `__filename` to get current file name

**reset.ts (yarn migration-steps:reset):
dont autorun the reset step - should we ? @minotogna 
this requires to run `migration-steps:run` after resetting
i think this approach is more failsafe (requires user to run the migration steps that would be run on other instances, eg review, other developer, prod..)**


src/test/migrations/reset.ts
```diff
@@ -58,17 +58,11 @@ const resetMigrationSteps = async () => {
...
 
     // === 3. Create reset step
-    const { filePath, fileName } = await createMigrationStep('reset')
+    const { fileName } = await createMigrationStep('reset')
     // eslint-disable-next-line @typescript-eslint/no-var-requires,global-require,import/no-dynamic-require
-    const resetStep = require(filePath).default
 
-    await DB.tx(async (t) => {
-      // === 4. Run reset step
-      await resetStep(t, fileName)
```
sorja, line 67: autorun removed @minotogna

sorja: meaning that `migration-steps:reset` will only delete the files and create the reset step

afterwards it is expected to align the DB with `migration-steps:run`

src/test/migrations/index.ts
```diff
@@ -67,11 +67,14 @@ const exec = async () => {
       }
     }
   })
+
   if (!process.argv.includes('--watch')) {
     // eslint-disable-next-line no-restricted-syntax
     for (const file of executedMigrations) {
-      // eslint-disable-next-line no-await-in-loop
-      await client.query('insert into migrations.steps (name) values ($1)', [file])
+      if (!file.endsWith('-step-reset.ts')) {
```
minotogna, line 74: How do other developers reset the migrations after one has run the reset command ?

minotogna (approved):

src/test/migrations/index.ts
sorja: Running migration-steps now should be ok

PR #4100, opened 2024-11-08
4001 - Add initial header styles

yaguzmang:
image

Coming in next PR:
- Tooltips fix. Currently DataCell doesn't show html tooltips.
- Calculated cells styling

### DIFF
--- a/src/client/pages/Section/DataTable/Table/Table.scss
+++ b/src/client/pages/Section/DataTable/Table/Table.scss
@@ -1,44 +1,10 @@
 @import 'src/client/style/partials';
 
-.table__grid {
+.table-grid {
   font-size: $font-s;
   line-height: 19px;
   overflow-x: auto;
   width: 100%;
-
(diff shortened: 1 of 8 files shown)


src/client/pages/Section/DataTable/Table/Table.scss
```diff
@@ -1,11 +1,44 @@
 @import 'src/client/style/partials';
 
-
 .table__grid {
   font-size: $font-s;
   line-height: 19px;
   overflow-x: auto;
   width: 100%;
+
+  .data-cell {
```
minotogna: Better to add their own css file and with prefix . Eg table-grid__ ... and rename the table__grid to table-grid .
We isolate the style and don't create a big file here .
What do u think ?

yaguzmang: Yes. Added 👍

minotogna (approved):

PR #4101, opened 2024-11-11
4098 - Metadata: Select Options per cycle + select years options

### LINKED ISSUE #4098: Metadata: Select Options per cycle + select years options

minotogna:
image

image

image


sorja (approved):
👍 👍

PR #4103, opened 2024-11-11
4032 - Users: Get Many through Country User Summary

sorja:
- [x] All features/filters work
- [x] Use params vs string concatenation
- [x] Investigate slow performance (maybe missing index)
- [x] Include users get count: https://github.com/openforis/fra-platform/pull/4103/commits/a5773c728271dcebf591ba3cf68988081b08f625

Upcoming PRs:

1 update admin UI
2. update country home UI

### DIFF
--- a/src/meta/user/countryUserSummary.ts
+++ b/src/meta/user/countryUserSummary.ts
@@ -1,0 +1,7 @@
+import { User } from 'meta/user/user'
+import { UserInvitation } from 'meta/user/userInvitation'
+
+export type CountryUserSummary = User & {
+  readonly fullName: string
+  readonly invitations: Array<UserInvitation>
+}
--- a/src/server/repository/public/user/getMany.ts
+++ b/src/server/repository/public/user/getMany.ts
@@ -5,12 +5,7 @@ import { Assessment, Cycle } from 'meta/assessment'
 import { TablePaginatedOrderByDirection } from 'meta/tablePaginated'
 import { RoleName, User, UserStatus } from 'meta/user'
 
-import { BaseProtocol, DB } from 'server/db'
-import { UserRoleAdapter } from 'server/repository/adapter'
-
-import { fields } from './fields'
-
-const selectFields = fields.map((f) => `u.${f}`).join(',')
+import { BaseProtocol, DB, Schemas } from 'server/db'
 
 export type UsersGetManyProps = {
   assessment?: Assessment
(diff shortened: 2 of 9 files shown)


src/meta/user/countryUserSummary.ts
```diff
@@ -0,0 +1,7 @@
+import { User } from 'meta/user/user'
+import { UserInvitation } from 'meta/user/userInvitation'
+
+export type CountryUserSummary = User & {
+  readonly fullName: string
+  readonly invitations: Array<UserInvitation>
```
minotogna, line 6: This object should have country iso and one optional invitation ?

src/server/repository/public/user/getMany.ts
```diff
@@ -67,76 +54,46 @@ export const buildGetManyQuery = (props: UsersGetManyProps): BuildQueryReturned
...
-      and
-      `)}
-      group by ${selectFields}
-                   ${order}
-                   ${limit ? `limit ${limit}` : ''}
-                   ${offset ? `offset ${offset}` : ''}
+  with filtered_users as (
+    select distinct id
+    from ${schemaName}.country_user_summary cus
+    where 1 = 1
```
minotogna: Why this ? The where clauses could be build in an array and joined with and .

Also , does this query works with current admin filters ?

sorja: This query works with current admin filters

It was added for readability and separating all where clauses, 1=1 always return true and we can concatenate with ` and <clause>`

I will update

src/meta/user/countryUserSummary.ts
sorja: This is something I am also investigating now that you mentioned it

In country collaborator view - yes, this is what we want: a single user, for single country with a single role

In admin though, we have one row with multiple country+role, so it makes sense to extend the user (although, we don't really extend all fields, but this is so we can use eg. Users.getRole)

src/server/repository/public/user/getMany.ts
minotogna: Ok I see 😉it looked strange . I'll check the full file , not the diffs here .

sorja:
@minotogna 

I have updated this PR, but I leave it in draft until tomorrow ( i want to double check it twice ).

The request takes now 108ms in local, down from 4s500ms (about 41x faster than initial)

The filtering including multiple filters is tested and working

minotogna (approved):

PR #4140, opened 2024-11-22
4133 - Add Kiosk routes

yaguzmang:
Bare-bone Kiosk routes:
image
image
image
image
image
image

Lazy loading (buttons are there just for the vid):
 Notice:
  - Non-kiosk route doesn't fetch the Kiosk bundle
  - Kiosk bundle is fetched only once when navigating in the Kiosk routes. 

https://github.com/user-attachments/assets/929c531f-3039-4e44-b241-b0b2b2a38f3b


src/client/pages/PageRoutes/hooks/useRoutes.tsx
```diff
@@ -4,6 +4,8 @@ import { createRoutesFromElements, Navigate, Route } from 'react-router-dom'
 import { RegionCode } from 'meta/area'
 import { Routes } from 'meta/routes/routes'
 
+import Kiosk from 'client/components/Kiosk/Kiosk'
+import KioskPageLayout from 'client/components/Kiosk/PageLayout'
```
yaguzmang: @minotogna I added the Kiosk related components at `client/components/Kiosk/` for now. Let me know if you think those should go in a different folder.

src/client/pages/PageRoutes/hooks/useRoutes.tsx
```diff
@@ -30,57 +32,68 @@ import User from 'client/pages/User'
...
+                <Route element={<LoginResetPassword />} path={Routes.LoginResetPassword.path.relative} />
               </Route>
-              <Route element={<LoginResetPassword />} path={Routes.LoginResetPassword.path.relative} />
             </Route>
           </Route>
+          <Route element={<PanEuropeanRedirect />} path={`/${RegionCode.FE}/*`} />
+          <Route element={<Navigate replace to={Routes.Root.path.relative} />} path="*" />
+        </Route>
+        {/* Kiosk */}
+        <Route element={<KioskPageLayout />} path={Routes.Kiosk.path.relative}>
```
minotogna: Maybe better to call KioskLayout , and even the component (not page layout, that's used for main app )

src/client/pages/PageRoutes/hooks/useRoutes.tsx
minotogna: I think it's ok all in there .

src/client/pages/PageRoutes/hooks/useRoutes.tsx
minotogna: I would also dynamically lazy load all the kiosk related js .

https://react.dev/reference/react/lazy

https://remix.run/blog/lazy-loading-routes

What do you think ?

minotogna: Check the Geo route . That's lazy loaded

src/client/components/Kiosk/PageLayout/PageLayout.scss
```diff
@@ -0,0 +1,8 @@
+@import 'src/client/style/partials';
+
+.kiosk-container {
+  display: flex;
```
minotogna: This can be a grid that defines the width of the components .
The side panel I would make fixed width .

src/client/pages/PageRoutes/hooks/useRoutes.tsx
minotogna: Rethinking , for consistency page components are under Pages . The Kiosk pages could go under pages/Kiosk

yaguzmang: I added the pages to `pages/Kiosk`

The KioskLayout is kept in `/components`, and the side panel is at `/components/KioskLayout/SidePanel`
similar to how we have PageLayout and Footer, Toolbar, etc.. within in, for consistency

src/client/pages/PageRoutes/hooks/useRoutes.tsx
yaguzmang: I added an inner file `_lazyLoad` that uses React.lazy and also wraps the component in suspense so we don't have to define it every time (see vid in description). Let me know what you think.

src/client/pages/PageRoutes/hooks/useRoutes.tsx
```diff
@@ -27,60 +27,81 @@ import SectionAreaSwitch from 'client/pages/SectionAreaSwitch'
...
 
+import { lazyLoad } from './_lazyLoad'
+
+const ForestKids = lazyLoad(() => import('client/pages/Kiosk/ForestKids'))
+const Fra2020DataPlatform = lazyLoad(() => import('client/pages/Kiosk/Fra2020DataPlatform'))
+const FraProcess = lazyLoad(() => import('client/pages/Kiosk/FraProcess'))
+const InteractiveStories = lazyLoad(() => import('client/pages/Kiosk/InteractiveStories'))
+const Kiosk = lazyLoad(() => import('client/pages/Kiosk/Kiosk'))
+const KioskLayout = lazyLoad(() => import('client/components/KioskLayout'))
+const RecentHighlights = lazyLoad(() => import('client/pages/Kiosk/RecentHighlights'))
```
minotogna: it would be nice if all these components could be part of 1 bundle. their own bundle. with this I think there's one bundle per import. can u check if it's somehow possible ?

yaguzmang: Added it with [webpackchunkname](https://webpack.js.org/api/module-methods/#webpackchunkname) (see new vid).
I also moved the kioks routes definition to `_KioskRoutes.tsx` to make it more organized

src/client/components/KioskLayout/KioskLayout.tsx
```diff
@@ -0,0 +1,16 @@
+import './KioskLayout.scss'
+import React from 'react'
+import { Outlet } from 'react-router-dom'
+
+import SidePanel from './SidePanel'
+
+const KioskLayout: React.FC = () => {
```
minotogna: Can we move this under kiosk pages as well ?

src/client/pages/PageRoutes/hooks/_KioskRoutes.tsx
```diff
@@ -0,0 +1,68 @@
...
+  () =>
+    import(
+      /* webpackChunkName: "kiosk" */
+      'client/pages/Kiosk/Kiosk'
+    )
+)
+const RecentHighlights = React.lazy(
+  () =>
+    import(
+      /* webpackChunkName: "kiosk" */
```
minotogna, line 54: We will migrate to vitejs soonish . 
Is there any more flexible way or we'll need to refactor when we'll use vite ?

yaguzmang: From what I'm seeing, to do this code splitting in vite we can use the [Chunking strategy](https://vite.dev/guide/build.html#chunking-strategy) "output.manualChunks": https://rollupjs.org/configuration-options/#output-manualchunks, and it seems it's all done in the config, so here in the code we will just need to delete the webpackChunkName comments.

minotogna: Nice 👍

minotogna (approved):

PR #4149, opened 2024-11-27
4133 - Kiosk v1 - Side panel

yaguzmang:
image


src/client/pages/Kiosk/Kiosk.scss
```diff
@@ -1,5 +1,3 @@
-@import 'src/client/style/partials';
```
yaguzmang, line 1: Removing these imports as the Kiosk doesn't really follow our established conventions fully. Also, in the future we could ideally have the Kiosk bundle be fetched on its own, and for that better have it as independent as possible from the platform. Maybe that can be accomplish when we migrate to Vite and we can avoid fetching the main bundle unnecessarily.

src/client/pages/Kiosk/KioskLayout/SidePanel/SidePanel.tsx
```diff
@@ -1,8 +1,45 @@
...
+        <img alt="FAO" src="/img/fao/FAOen_3_lines.svg" />
+        <button
+          className="kiosk-side-panel__home-button"
+          onClick={() => startTransition(() => navigate(Routes.Kiosk.path.absolute))}
+          type="button"
+        >
+          <Icon name="home-circle" />
+        </button>
+      </div>
+      <div className="kiosk-side-panel__quiz">
```
yaguzmang, line 25: If we want to hide the quiz for the first deployment we can simply comment it out and the side panel would look like this:
image

minotogna (approved):

PR #4151, opened 2024-11-28
4133 -  Kiosk v1 - Content

yaguzmang:
Content optimized for full HD resolution:

https://github.com/user-attachments/assets/001d36ae-017d-43a9-9ee4-b9aff2909dfc

Forest kids (didn't load on responsive mode):

https://github.com/user-attachments/assets/694060d2-6e5e-4f4e-89d0-b0d38a385855


src/client/pages/Kiosk/InteractiveStories/InteractiveStories.tsx
```diff
@@ -1,7 +1,49 @@
...
+    link: Routes.TheSecretsOfMangroves.path.relative,
+    title: 'The secrets \n of mangroves',
+  },
+]
+
 const InteractiveStories: React.FC = () => {
-  return <div>Interactive Stories</div>
+  return (
+    <div className="kiosk-content">
+      <h1 className="kiosk-content__title">Interactive Stories</h1>
```
yaguzmang, line 40: I reused some classes here, however I noticed the styles won't be applied if the page is refreshed in a nested route:

https://github.com/user-attachments/assets/d7732812-5067-410e-a7a3-c8a35190df79

This is because even though the kiosk bundle has already been fetched, the styles from the other components are not loaded in the browser until the components are rendered (probably because of the lazy loading). So, I can either import the styles directly here, or duplicate some code, or add something like  'kiosk-common.scss' styles.
@minotogna Let me know what you think

web-resources/img/kiosk/recent-highlights.jpg
yaguzmang, line 1: I couldn't find the exact image used in the mockup, so I used this one from the news letter. We can change it if provided.

src/client/pages/Kiosk/InteractiveStories/InteractiveStories.tsx
minotogna: Have you tried to import that same css file in this file?

Or the one bundle is working ? Have you checked what gets fetched ?

yaguzmang: Yes I checked what's fetched and there is nothing new besides the images:

https://github.com/user-attachments/assets/b975e712-1440-444d-b4fd-3d088ba6e238

The styles from lazy loaded components aren't available until imported directly, unlike the regular imports we use:

https://github.com/user-attachments/assets/28339fb7-9083-47f1-99a5-a151762c7a90

And, yes, I tried importing directly in the same file as my first approach, and it works as expected :+1:. I didn't see any examples of that in the rest of the platform, so I wanted to note this here.

src/client/pages/Kiosk/InteractiveStories/InteractiveStories.tsx
```diff
@@ -1,7 +1,50 @@
+import './InteractiveStories.scss'
+import 'client/pages/Kiosk/Kiosk.scss'
```
yaguzmang, line 2: Styles applied after refresh:

https://github.com/user-attachments/assets/2353435a-bf68-45f6-8c4a-a2185f0a9348

minotogna (approved):

PR #4160, opened 2024-11-28
4032 - Fix migrations

sorja:
Fix migrations:
- For somea reason using `for await (const file of migrationSteps) {` messes up running the migrations when one migration is not using transcation (eg 20241016104908-step-users_role-invitation-split.ts) and other ones are running under transcation.
- Create schema _legacy if it doesn't exist (partial db copy support)

### DIFF
--- a/src/test/migrations/steps/20241016104908-step-users_role-invitation-split.ts
+++ b/src/test/migrations/steps/20241016104908-step-users_role-invitation-split.ts
@@ -90,6 +90,7 @@ export default async () => {
   await client.query(`create unique index if not exists users_uuid_key on public.users (uuid);`)
 
   // ---- 2. Move users_role table to _legacy schema
+  await client.query(`create schema if not exists _legacy;`) // eg running vs partial db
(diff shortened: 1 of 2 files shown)


src/test/migrations/steps/20241016104908-step-users_role-invitation-split.ts
```diff
@@ -90,6 +90,7 @@ export default async () => {
   await client.query(`create unique index if not exists users_uuid_key on public.users (uuid);`)
 
   // ---- 2. Move users_role table to _legacy schema
+  await client.query(`create schema if not exists _legacy;`) // eg running vs partial db
```
minotogna, line 93: Why this ?

sorja: It says it right there:

Running vs partial db copy we don't have legacy schema

(Also soon we will remove the _legacy schema anyway)

minotogna (approved):

PR #4161, opened 2024-11-29
4133 - Kiosk v1 - Hide quiz & Add responsiveness 

yaguzmang:
Entire Kiosk working:

https://github.com/user-attachments/assets/d05c88f5-9e42-42b8-bda5-71fc6f52b3a7

After this PR, the main PR is ready to merge. Unless there is something new to add.

(UPDATED) Responsiveness: 

https://github.com/user-attachments/assets/c1c9f24f-5e5e-4f7f-9ca3-2c95c1cbb023

To make the embedded content look bigger for higher resolutions, we need to zoom in the browser as seen in the end of the video. This is the only way since it's not possible to manipulate the object content.

### DIFF
--- a/src/client/pages/Kiosk/InteractiveStories/InteractiveStories.scss
+++ b/src/client/pages/Kiosk/InteractiveStories/InteractiveStories.scss
@@ -1,5 +1,5 @@
 .kiosk-interactive-stories__grid-container {
   display: grid;
-  gap: 64px;
+  gap: min(3.33dvw, 5.925dvh);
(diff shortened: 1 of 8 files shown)


minotogna:
@yaguzmang can we leave that image in the side panel ?

minotogna:
> @yaguzmang can we leave that image in the side panel ?

We could just hide the text 'Fra quiz' for now . What do you think ?

minotogna (approved):

PR #4169, opened 2024-12-03
4001 - Datatable use Datagrid - Add review comments

yaguzmang:
Datatable row highlight:

https://github.com/user-attachments/assets/dcbede46-33f6-4652-8bfa-52f54e938bee

special cases:

https://github.com/user-attachments/assets/47b4b0f3-f12d-49e7-b10b-3f1a1861ac76

RTL:

https://github.com/user-attachments/assets/f2fa9d89-9b5d-4363-b39c-35bb499a0f07

ODP unaffected: 

https://github.com/user-attachments/assets/6261c649-544d-4c2a-984c-186d5f7673b2

### DIFF
--- a/src/client/components/DataGrid/DataRow/DataRow.tsx
+++ b/src/client/components/DataGrid/DataRow/DataRow.tsx
@@ -1,16 +1,20 @@
 import React, { PropsWithChildren, ReactElement } from 'react'
 
+import { Objects } from 'utils/objects'
+
 import { DataRowAction } from 'client/components/DataGrid'
 import Actions from 'client/components/DataGrid/DataRow/Actions'
 
 import { useHighlighted } from './hooks/useHighlighted'
+import { DataRowHighlightRange } from './types'
 
 type DataRowProps = PropsWithChildren<{
   actions?: Array<DataRowAction>
+  highlightRange?: DataRowHighlightRange
 }>
 
 const DataRow: React.FC<DataRowProps> = (props) => {
-  const { actions, children } = props
+  const { actions, children, highlightRange } = props
 
   const highlighted = useHighlighted({ actions })
 
@@ -18,9 +22,19 @@ const DataRow: React.FC<DataRowProps> = (props) => {
     <>
       {React.Children.map(children, (child, idx) => {
         if (!React.isValidElement(child)) return null
-        const firstCol = idx === 0
-        const lastCol = idx === React.Children.count(children) - 1
-        return React.cloneElement(child as ReactElement, { firstCol, highlighted, lastCol })
+
+        let firstCol = idx === 0
+        let lastCol = idx === React.Children.count(children) - 1
+        let shouldHighlightCell = highlighted
+        if (!Objects.isEmpty(highlightRange)) {
+          shouldHighlightCell &&= idx >= highlightRange.start && idx <= highlightRange.end
+          if (shouldHighlightCell) {
+            firstCol = idx === highlightRange.start
+            lastCol = idx === highlightRange.end
+          }
+        }
+
+        return React.cloneElement(child as ReactElement, { firstCol, highlighted: shouldHighlightCell, lastCol })
       })}
       <Actions actions={actions} />
     </>
@@ -29,6 +43,7 @@ const DataRow: React.FC<DataRowProps> = (props) => {
 
 DataRow.defaultProps = {
   actions: [],
+  highlightRange: undefined,
 }
 
 export default DataRow
--- a/src/client/components/DataGrid/DataRow/DataRow.tsx
+++ b/src/client/components/DataGrid/DataRow/DataRow.tsx
@@ -17,24 +17,30 @@ const DataRow: React.FC<DataRowProps> = (props) => {
   const { actions, children, highlightRange } = props
 
   const highlighted = useHighlighted({ actions })
+  const highlightRangeExists = !Objects.isEmpty(highlightRange)
 
   return (
     <>
       {React.Children.map(children, (child, idx) => {
         if (!React.isValidElement(child)) return null
 
-        let firstCol = idx === 0
-        let lastCol = idx === React.Children.count(children) - 1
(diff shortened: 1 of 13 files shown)


src/client/pages/Section/DataTable/Table/Table.tsx
```diff
@@ -56,8 +64,18 @@ const Table: React.FC<Props> = (props) => {
...
+          className="table-grid"
+          gridTemplateColumns={`minmax(auto, 400px) repeat(${headers.length},1fr)`}
+          withActions={withActions}
+        >
+          <GridHead
+            assessmentName={assessmentName}
+            data={data}
+            headers={headers}
+            table={table}
+            withActions={withActions}
```
minotogna, line 77: GridHead doesnt really do anything with withActions prop. it just renders an empty div.
After the GridHead you can simply add ```{withActions && <div />}```?

src/client/components/DataGrid/DataRow/DataRow.tsx
```diff
@@ -1,26 +1,40 @@
...
-        return React.cloneElement(child as ReactElement, { firstCol, highlighted, lastCol })
+        if (!React.isValidElement(child)) return null
+
+        let firstCol = idx === 0
+        let lastCol = idx === React.Children.count(children) - 1
+        let shouldHighlightCell = highlighted
+        if (!Objects.isEmpty(highlightRange)) {
+          shouldHighlightCell &&= idx >= highlightRange.start && idx <= highlightRange.end
+          if (shouldHighlightCell) {
+            firstCol = idx === highlightRange.start
```
minotogna: altering firstCol and lastCol might alter the intended use from outside. and should not be done

src/client/components/DataGrid/DataRow/DataRow.tsx
```diff
@@ -1,26 +1,40 @@
...
-        if (!child) return null
-        const firstCol = idx === 0
-        const lastCol = idx === React.Children.count(children) - 1
-        return React.cloneElement(child as ReactElement, { firstCol, highlighted, lastCol })
+        if (!React.isValidElement(child)) return null
+
+        let firstCol = idx === 0
+        let lastCol = idx === React.Children.count(children) - 1
+        let shouldHighlightCell = highlighted
+        if (!Objects.isEmpty(highlightRange)) {
```
minotogna: all this code is a little bit tricky to understand and maintain. 
The ideal situation would be: from outside one passes the highlight range, then it's highlighted if it's highlighted and (no highlight range is passed or the index of the cell is within the highlight range).
Can you check if you find a way to make it work like this?

src/client/pages/Section/DataTable/Table/GridRow/RowData/hooks/useHighlightRange.ts
```diff
@@ -0,0 +1,41 @@
...
+  return useMemo<Returned>(() => {
+    const rowSpans = cols.map((col) => Cols.getStyle({ col, cycle }).rowSpan ?? 1)
+    if (rowSpans.length === 0) {
+      return undefined
+    }
+
+    const end = rowSpans.length - 1
+    const targetValue = rowSpans[end]
+
+    // returns the range of consecutive equal row spans, starting from the last column.
```
minotogna: i'm wondering why the 'consecutive equal row spans'.

I was thinking more:
the start is the first col with colSpan = 1
and the end is the last col with colSpan = 1
Simpler maybe?

src/client/pages/Section/DataTable/Table/Table.tsx
yaguzmang: GridHead has a loop and it renders the empty div for each row, so that wouldn't work. 
However, in a coming PR I will get rid of `GridRow`, so in that same PR I'll remove `GridHead` and render directly in the main Table component.

src/client/components/DataGrid/DataRow/DataRow.tsx
yaguzmang: Added  firstHighlightCol & lastHighlightCol and simplified the logic 👍

src/client/components/DataGrid/DataRow/DataRow.tsx
```diff
@@ -1,26 +1,55 @@
...
+    [children]
+  )
 
   return (
     <>
-      {React.Children.map(children, (child, idx) => {
-        if (!child) return null
+      {validChildren.map((child, idx) => {
         const firstCol = idx === 0
-        const lastCol = idx === React.Children.count(children) - 1
```
yaguzmang, line 22: I came across this obscure bug while testing:
When you render conditionally in the DataRow component, like:
```
<DataRow>
   <DataCell>...<DataCell/>
    ...
   {someCondition && <DataCell>...<DataCell/>}
</DataRow>
```
When `someCondition` is false, the result of `{someCondition && <DataCell>...<DataCell/>}`  is actually considered a child with value =  `null`, which means that when looping with `React.Children.map(children, (child, idx)` this `null` child affects the idx value.
Since we use the idx value to render the borders, if a row has a first or last `null` child, something like this would happen (constructed example):
Pasted Graphic 12
the border would be missing.

The fix was to filter the valid children before the loop.

src/client/components/DataGrid/DataRow/DataRow.tsx
```diff
@@ -1,26 +1,55 @@
...
 const DataRow: React.FC<DataRowProps> = (props) => {
-  const { actions, children } = props
+  const { actions, children, highlightRange } = props
 
   const highlighted = useHighlighted({ actions })
+  const highlightRangeExists = !Objects.isEmpty(highlightRange)
+
+  const validChildren = useMemo<React.ReactElement[]>(
+    () =>
+      React.Children.toArray(children).reduce<Array<React.ReactElement>>((acc, child) => {
```
minotogna, line 24: Maybe here even an array.filter might work

minotogna (approved):

PR #4174, opened 2024-12-09
4032 - Dashboard UI

minotogna:
https://github.com/user-attachments/assets/19be3cbc-06c0-4f03-bcd1-15734d360d4c


src/meta/tablePaginated/sortFn.ts
```diff
@@ -0,0 +1 @@
+export type TablePaginatedCompareFn<Datum extends object> = (a: Datum, b: Datum) => number
```
sorja, line 1: should this file be compareFn.ts?

sorja (approved):

PR #4176, opened 2024-12-09
4001 - Datatable use Datagrid - Input number

yaguzmang:
Inputs (updated):

https://github.com/user-attachments/assets/97bd794d-c3ad-4c93-903f-dffb98001e73

Inputs in the Datagrid (updated):

https://github.com/user-attachments/assets/de817304-fd6a-4d72-ab74-14db7e4c337a


src/client/pages/Section/DataTable/Table/RowData/Cell/Number/Number.tsx
```diff
@@ -2,21 +2,28 @@ import React from 'react'
 
 import { ColType } from 'meta/assessment'
 
-import ThousandSeparatedDecimalInput from 'client/components/ThousandSeparatedDecimalInput'
-import ThousandSeparatedIntegerInput from 'client/components/ThousandSeparatedIntegerInput'
```
yaguzmang, line 6: We still use `PercentInput`, `ThousandSeparatedDecimalInput` or`ThousandSeparatedIntegerInput` in the components:
- `src/client/pages/OriginalDataPoint/components/ForestCharacteristics/ForestCharacteristicsRow.tsx`

- `src/client/pages/OriginalDataPoint/components/ForestCharacteristics/PrimaryForestPercent/PrimaryForestPercent.tsx`,

- `src/client/pages/OriginalDataPoint/components/ExtentOfForest/ExtentOfForestRow.tsx`

image

So I'll delete `PercentInput`, `ThousandSeparatedDecimalInput`, `ThousandSeparatedIntegerInput` once those tables are refactored as well.

minotogna: Sure @yaguzmang

src/client/components/Inputs/InputNumber/InputNumber.scss
```diff
@@ -0,0 +1,51 @@
+@import 'src/client/style/partials';
+
```
minotogna: It's a great improvement. I have a few questions / doubts I will post here 

Is there any chance to let number input use text input to avoid duplication of the code ? 

E.g. can number input just be a wrapper of text input ? 
If not , can number input use the same style ?

src/client/components/Inputs/InputNumber/InputNumber.tsx
```diff
@@ -0,0 +1,85 @@
...
+    disabled,
+    id,
+    isInteger,
+    maxLength,
+    onChange,
+    onPaste,
+    placeholder,
+    precision,
+    shouldRound,
+    thousandSeparated,
```
minotogna: Is this used only in formatting the read only value ? Isn't always the case ?

src/client/components/Inputs/InputNumber/InputNumber.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const {
+    disabled,
+    id,
+    isInteger,
+    maxLength,
+    onChange,
+    onPaste,
+    placeholder,
+    precision,
+    shouldRound,
```
minotogna: And is this always the case as well ?

src/client/components/Inputs/InputNumber/InputNumber.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+import { useOnBlur } from './hooks/useOnBlur'
+import { useOnChange } from './hooks/useOnChange'
+import { InputNumberProps } from './types'
+
+const InputNumber = forwardRef<HTMLInputElement, InputNumberProps>((props, outerRef) => {
+  const {
+    disabled,
+    id,
+    isInteger,
```
minotogna: Can we have only one prop between is integer and precision ?
Integer can be derived if precision =0?

src/client/components/Inputs/InputNumber/InputNumber.tsx
```diff
@@ -0,0 +1,85 @@
...
+    precision,
+    shouldRound,
+    thousandSeparated,
+    value,
+  } = props
+
+  const inputRef = useRef<HTMLInputElement>(null)
+  useImperativeHandle(outerRef, () => inputRef.current!, [])
+
+  const [localValue, setLocalValue] = useState<typeof value>(value)
```
minotogna: Can we avoid local value ? Only The read only value is different , right ?

src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
```diff
@@ -36,7 +36,7 @@
     justify-content: start;
   }
 
-  .number-input__container {
+  .input-number__container {
```
minotogna: Can we move this where it's used ?

src/client/components/Inputs/InputNumber/InputNumber.tsx
yaguzmang: When removing the local value, this happens when the `value` from the caller is async and formatted/sanitized:

https://github.com/user-attachments/assets/6fe4d055-1073-4428-9416-c672608ffb91

Issues in this vid:
- When you start typing you get the .00 format. 
- You can experience that you delete something but it comes back for a split second because of the async+formatting.
- Deleting from the end of the input is funky. If you go from, let's say `123.00` to `123.0` while deleting, the format applied by the component caller is two decimals so you go back to `123.00`. 

It's similar to the issues we had with the EditorWYSIWYG, since the value is sanitized/formatted and async. There the fix was to call the passed `onChange` only when the editor is out of focus.

We could remove the local value and some other logic but then this behavior would need to be handled by the caller. What do you think?

src/client/components/Inputs/InputNumber/InputNumber.tsx
yaguzmang: `shouldRound` I added to keep this feature from the old input:


https://github.com/user-attachments/assets/915d3c7f-7514-47ab-a568-ed52c2013d93



This prop and `thousandSeparated` I added so the component is more flexible, but if you think those won't be needed at all, I can remove them 👍

src/client/components/Inputs/InputNumber/InputNumber.tsx
yaguzmang: The core issue is separating when the formatting is applied from when the user is typing. We don't have these problems with the `InputText` because we don't format/sanitize the values where we use it.

src/client/components/Inputs/InputNumber/InputNumber.tsx
minotogna: the rounding shouldn't happen while editing data ?

src/client/components/Inputs/InputNumber/InputNumber.tsx
minotogna: the formatted value is only for read only

yaguzmang: Yes, I mean the  format applied to the `value` prop by the caller `onChange`. in this case, when updating the node value: 
image
At `src/client/pages/Section/DataTable/Table/RowData/Cell/hooks/useOnChange.ts` 
`Sanitizer.sanitize` also formats the value, which is why if you delete a 0 from `123.00` it stays as `123.00`. 
I could remove that for decimal and integer value types. 

Also, if we don't have local value, the caller needs to handle the input of the decimal period, eg `123.` would need to be considered valid so you can get to input `123.1` using the value prop. At the moment, `Sanitizer.isAcceptable` would return false for `123.`so the value update wouldn't run and the input would get stuck at `123`

src/client/components/Inputs/InputNumber/InputNumber.scss
yaguzmang: > Is there any chance to let number input use text input to avoid duplication of the code ?

Yes we can use, although it won't remove a lot of code. The custom `onBlur` we would still need to round the input value after it is unfocused. The custom `onChange` we would still need to accept only valid numbers. So, at the end it will remove only the logic for the cursor position, and the styles partially. I will still need to add some specific styles to the input number to place it correctly on top of the read only div. I can make this change :+1:

src/client/components/Inputs/InputNumber/InputNumber.tsx
yaguzmang: If it happens while you edit, the input itself will be editing the value as you input, that might change the cursor position and affect the UX. An alternative is to limit the decimal numbers according to the precision, but I think it's better to let the user input and have the site round the value .

src/client/components/Inputs/InputNumber/InputNumber.scss
minotogna: I don't get why the input number must be placed on top of the read only div.
don't we render either one or the other? 

Then, of course we might need to have custom props, otherwise we wouldn't have needed this component at all.
But onChange, onBlur, and maybe some other props is anyway better than duplicating all the component. what do you think?

src/client/components/Inputs/InputNumber/InputNumber.tsx
minotogna: My objective wasn't to open a long discussion.

The natural flow of controlled components:
value, onChange come from outside. 
In the inner  input onChange, if the value is correct, then it calls the onChange. Therefore the value to pass to the props.onChange can be already sanitized or it gets sanitized as it is now. the code you show above is used also by the onPaste

In addition, the value displayed in read only mode, is another story. and formatted according to its rules

Does this make sense @yaguzmang ? and can we do it?

yaguzmang: Okay, I will make the adjustments 👍

src/client/components/Inputs/InputNumber/InputNumber.scss
yaguzmang: it's on top the readonly div so you can click on it and trigger the input focus. Though something simple in the input props like:
```
<input
value={focused ? value : formattedValue}
```
might work.
I will try something like that and get back to you.

src/client/components/Inputs/InputNumber/InputNumber.tsx
minotogna: why do we round while we edit the data? do we do it now? if yes, where ?

yaguzmang: I've removed this prop now, so the rounding done by the InputNumber component is always done `onBlur`.

minotogna (approved):
👍👍👍

PR #4178, opened 2024-12-12
4001 - Datatable use Datagrid - Taxon

yaguzmang:
https://github.com/user-attachments/assets/ce5f3027-50d9-4076-b53b-6704286398d9

Using enter:

https://github.com/user-attachments/assets/660c6dcc-994a-42aa-92f5-dcf2c6ac2949


src/client/components/Inputs/Select/hooks/useComponents.tsx
```diff
@@ -35,13 +35,13 @@ export const useComponents = (props: SelectProps): Returned => {
     }
     if (isMulti && selectableGroups) components.GroupHeading = SelectableGroupHeading
     if (collapsibleGroups) components.Group = CollapsibleGroup
-    if (Objects.isFunction(onPaste) || !Objects.isEmpty(inputHidden)) {
+    if (!Objects.isEmpty(inputHidden)) {
       components.Input = (originalInputProps: InputProps) => {
         // eslint-disable-next-line react/jsx-props-no-spreading
-        return <originalComponents.Input {...originalInputProps} isHidden={inputHidden} onPaste={onPaste} />
```
yaguzmang: Having the onPaste function here caused bugs because if it changes, the Input component is replaced with a new instance and thus loses its state. That caused problems with the default behavior, like the cursor jumping or the select list not hidden onBlur.

react-select doesn't support another way to pass the onPaste event. The other option would have been to extend the props of the native select and pass the onPaste there, but that seemed too hacky. Instead I removed it and added the onPaste handler to the Taxon div container.

src/test/migrations/steps/20241114204645-step-fix-datatable-grid-layouts-and-taxon-codes.ts
```diff
@@ -482,7 +482,53 @@ const _fixPanEuropean2025GridLayouts = async (client: BaseProtocol) => {
...
+            set value = updated_values.new_value
+            from updated_values
+            where ${schemaCycle}.node.id = updated_values.id;
+          `)
+        })
+      )
+    )
+  )
+}
+
```
yaguzmang, line 347: The regexp_replace takes care of cases like:
image

Pasted Graphic 4

src/client/components/Inputs/Select/hooks/useComponents.tsx
```diff
@@ -17,7 +17,7 @@ import { SelectProps } from 'client/components/Inputs/Select/types'
 type Returned = ReactSelectProps['components']
 
 export const useComponents = (props: SelectProps): Returned => {
-  const { collapsibleGroups, isMulti, multiLabelSummaryKey, selectableGroups } = props
+  const { collapsibleGroups, inputHidden, isMulti, multiLabelSummaryKey, selectableGroups } = props
```
minotogna, line 20: What's the scope of this prop ?

If you intend to hide the input search , there should be a select prop that does it already

src/client/components/Inputs/Select/types.ts
```diff
@@ -14,17 +15,35 @@ export type OptionsOrGroups = readonly (Option | OptionsGroup)[]
...
 export type ValueInput = string | Array<string> | null
 
-type SelectBaseProps = Pick<
-  ReactSelectProps,
-  'isClearable' | 'isMulti' | 'maxMenuHeight' | 'onMenuOpen' | 'onMenuClose' | 'placeholder'
->
+type SelectBaseProps =
+  | Pick<
+      ReactSelectProps,
+      | 'inputValue'
```
minotogna, line 21: Why do we pass input value now ?

src/client/pages/Section/DataTable/Table/RowData/Cell/Taxon/Taxon.scss
```diff
@@ -0,0 +1,10 @@
+@import 'src/client/style/partials';
+
+.table-grid__taxon-cell-container {
+  height: 100%;
+  width: 100%;
+
+  &.disabled {
```
minotogna: If it's disabled , the layout should not be affected here . The content should adapt itself

src/client/pages/Section/DataTable/Table/RowData/Cell/Taxon/Taxon.tsx
```diff
@@ -1,71 +1,54 @@
...
-        disabled={disabled}
-        items={items.map((item) => ({ value: item, label: item.scientificName }))}
-        name="taxon"
-        onInputValueChange={onInputValueChange}
-        onPaste={onPaste}
-        onSave={onSave}
-        value={nodeValue.raw}
-      />
+    <div className={classNames('table-grid__taxon-cell-container', { disabled })} onPaste={onPaste}>
+      {disabled && inputValue}
```
minotogna: Oh I see . Render a div with its own class

src/client/pages/Section/DataTable/Table/RowData/Cell/Taxon/hooks/useIsValidNewOption.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { Option, SelectProps } from 'client/components/Inputs/Select'
+
+type Props = {
+  nodeValue: NodeValue
+}
+
+type Returned = SelectProps['isValidNewOption']
+
+export const useIsValidNewOption = (props: Props): Returned => {
```
minotogna, line 15: Do we really need this now ?

src/client/pages/Section/DataTable/Table/RowData/Cell/Taxon/Taxon.tsx
minotogna: even though, there should be no need for this. the Select can have a prop disabled.

src/client/pages/Section/DataTable/Table/RowData/Cell/Taxon/hooks/useOnChange.ts
```diff
@@ -0,0 +1,45 @@
...
+export const useOnChange = (props: Props): Returned => {
+  const { nodeValue, onChangeNodeValue, options } = props
+
+  return useCallback<Returned>(
+    (value: string | null) => {
+      const nodeValueUpdate = { ...nodeValue }
+
+      if (value === CURRENT_NODE_OPTION_VALUE) return
+
+      if (value === null || value === '') {
```
minotogna: you can use is Objects.isEmpty(value)

src/client/components/Autocomplete/Autocomplete.tsx
```diff
@@ -1,143 +0,0 @@
-import './Autocomplete.scss'
-import React, { useEffect, useState } from 'react'
-import { useTranslation } from 'react-i18next'
-
-import classNames from 'classnames'
-import { useCombobox, UseComboboxStateChange } from 'downshift'
```
minotogna, line 6: can we get rid of 'downshift' dependency ?

src/client/components/Inputs/Select/hooks/useComponents.tsx
minotogna: Better to handle the onPaste in the input component , for consistency with others.
You can add a div around the Select that handles the onPaste ? 

https://stackoverflow.com/questions/53310989/using-onpaste-with-react-select

src/client/components/Inputs/Select/hooks/useComponents.tsx
yaguzmang: This prop is passed to the selectComponents.Input `isHidden` prop. There isn't a Select prop that is available that passes it directly to the internal input. 

I added it because the default behavior of react-select is to hide the input when selecting an option:
 
https://github.com/user-attachments/assets/6d88a7ce-ffaf-4dd5-bbf9-ca5f2cd2d602



Passing this prop as false to the select input makes it visible al the time. 

(related issue for reference: https://github.com/JedWatson/react-select/issues/3068)

src/client/components/Inputs/Select/types.ts
yaguzmang: When it's not passed, the state is controlled by react-select, and it has some default behavior like clearing the input onChange. Also, we want the input to be the same as the current node value on display, so it needs to be a controlled input. 

It's not necessary to pass it in other places we use the select component. When it's not passed it works as usual:

https://github.com/user-attachments/assets/237998fd-980f-4b90-b24c-5470b8ac9447

src/client/pages/Section/DataTable/Table/RowData/Cell/Taxon/hooks/useIsValidNewOption.ts
yaguzmang: Yes, otherwise the default internal behavior doesn't account for trimming or casing and this would happen:
image

or this
image

src/i18n/resources/en/common.json
```diff
@@ -1,6 +1,7 @@
 {
   "accepted": "Accepted",
   "add": "Add",
+  "addValue": "Add \"{{value}}\"",
```
minotogna: Can we reuse the add ?

minotogna (approved):

PR #4179, opened 2024-12-12
4032 - Fix User status

minotogna:


src/test/migrations/steps/20241016104908-step-users_role-invitation-split.ts
```diff
@@ -89,6 +89,20 @@ export default async () => {
...
 
+  // ---- 1.5 Fix users inconsistencies
+  await client.query(`
+      update users u
+      set status = '${UserStatus.active}'
+      where u.id in
+            (select distinct u.id
+             from users u
+                      left join users_role ur on u.id = ur.user_id
+             where ((ur.invited_at is not null and ur.accepted_at is not null)
```
sorja, line 100: If I understand this correctly - we had some active user roles with user status invitationPending?

sorja (approved):

PR #4186, opened 2024-12-17
4185 - Update country names in english

### LINKED ISSUE #4185: Country names (English translation)
ensure that in the platform English version of country names:
 
French Guyana is French Guiana
Saint-Martin (French part) is Saint Martin (French part)
 
See also country names according to NOC in this web list:
https://www.fao.org/forest-resources-assessment/fra-2020/country-reports/en/

yaguzmang:
I went over all country names in English, and there are no mismatches with the list provided besides "French Guiana", "Saint Martin (French Part)" and "Netherlands". Except Türkiye as we have it (prev Turkey), but that's probably because the source site is not fully up to date.

### DIFF
--- a/src/i18n/resources/en/area.json
+++ b/src/i18n/resources/en/area.json
@@ -309,7 +309,7 @@
     "listName": "Guatemala"
   },
   "GUF": {
-    "listName": "French Guyana"
+    "listName": "French Guiana"
   },
   "GUM": {
     "listName": "Guam"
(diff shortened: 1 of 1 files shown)


src/i18n/resources/en/area.json
```diff
@@ -516,7 +516,7 @@
     "listName": "Niue"
   },
   "NLD": {
-    "listName": "Netherlands (Kingdom of the)"
+    "listName": "Netherlands"
```
minotogna: This should not change 

https://www.fao.org/nocs/en

minotogna (changes requested):

src/i18n/resources/en/area.json
yaguzmang: Done. The rest of the names match NOCS.

minotogna (approved):

PR #4187, opened 2024-12-17
4032 - Collaborators/User edit final fixes

minotogna:


src/client/components/EditUserForm/UserCountryRoleSelector/hooks/useOptions.tsx
```diff
@@ -3,15 +3,25 @@ import { useTranslation } from 'react-i18next'
...
+  const user = useUser()
 
   return useMemo<Returned>(() => {
-    const options = Object.keys(RoleName).reduce<Returned>((acc, key) => {
+    const roles = Users.isAdministrator(user)
+      ? Object.keys(RoleName)
+      : Users.getRolesAllowedToEdit({ user, countryIso, cycle })
+
+    const options = roles.reduce<Returned>((acc, key) => {
       if (key !== RoleName.ADMINISTRATOR) {
```
yaguzmang: It seems `User.getRolesAllowedToEdit` also handles the admin user. It returns every role but administrator. So you can use `User.getRolesAllowedToEdit` instead of Rolename and then the check at line 25 can be removed too

minotogna: good catch 😃. update pushed

yaguzmang (approved):

PR #4190, opened 2024-12-24
4001 - Datatable use Datagrid - Fix Growing Stock Composition 2025 header row & Minor fixes

yaguzmang:
Before:
Pasted Graphic 8

After:
Pasted Graphic 9

Minor fixes:
table_1_4b_2020 labels before:
image

now:
image

### DIFF
--- a/src/test/migrations/steps/20241114204645-step-fix-datatable-grid-layouts-and-taxon-codes.ts
+++ b/src/test/migrations/steps/20241114204645-step-fix-datatable-grid-layouts-and-taxon-codes.ts
@@ -512,9 +512,59 @@ const _fixFraTaxonCodes = async (client: BaseProtocol) => {
   )
 }
 
+const _fixGrowingStockComposition2025HeaderRows = async (client: BaseProtocol) => {
+  const assessment = await AssessmentController.getOne({ assessmentName: AssessmentNames.fra }, client)
+
+  const schemaAssessment = Schemas.getName(assessment)
+
+  // Move 'header_1' row cols to 'header_0' row
+  await client.query(
+    `with target_row as (
+      select 
+          r.id as target_row_id
+      from ${schemaAssessment}.row r
+      join ${schemaAssessment}.table t on r.table_id = t.id
+      where r.props ->> 'index' = 'header_0'
+          and t.props ->> 'name' = 'growingStockComposition2025'
+    ),
+    cols_to_update as (
+      select 
+          c.id as col_id, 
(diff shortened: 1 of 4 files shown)


src/test/migrations/steps/20241114204645-step-fix-datatable-grid-layouts-and-taxon-codes.ts
```diff
@@ -512,9 +512,59 @@ const _fixFraTaxonCodes = async (client: BaseProtocol) => {
...
+
+      delete from ${schemaAssessment}.row
+      where id in (
+        select 
+            r.id
+        from ${schemaAssessment}.row r
+        join ${schemaAssessment}.table t on r.table_id = t.id
+        where r.props ->> 'index' = 'header_1'
+            and t.props ->> 'name' = 'growingStockComposition2025'
+      );`
```
yaguzmang, line 747: Any alternatives to handle moving cols from one row to another? 🤔

minotogna: Having two cols ?

yaguzmang: Having two cols as in duplicating them? 

I was trying to simplify the query move the cols, but this was the most concise I could get it to.

minotogna: Not sure I understand then 😅
Why would you need to move a column ? A column is associated to a row through row_id .

yaguzmang: Ahh I see 😅 Allow me to explain:
This issue with the header:
image

was there because the cols "FRA 2025 categories", "Scientific name", "Common name" were associated to a `row_id`, and "Million m³" with "% of total" were associated to a different `row_id`.
 When the actions are visible, we render an empty div for each header row, and because in the data we had two different rows for the same header, an extra empty div was added unnecessarily. 

So, in the query I change the `row_id` of "Million m³" and "% of total" to the `row_id` of  "FRA 2025 categories", "Scientific name" and "Common name". That way those are all in the same row. I then delete the row that is left with no cols. 

My question is if any other alternatives to that query come to mind

minotogna (approved):

PR #4198, opened 2025-01-06
4196 - Show invited users in country collaborators

### LINKED ISSUE #4196: UI: Show invited users in country collaborators
Steps:

- Invite new user that doesn't exist in the platform
- Get redirected back to users view

Expected:
- New invitation is shown in the list with other users

Current:
- New invitation is **not** shown in the list

sorja:
- [ ] Admin collaborators should not show invitations (eg. A.P. X20)

### DIFF
--- a/src/server/api/user/getMany.ts
+++ b/src/server/api/user/getMany.ts
@@ -1,6 +1,8 @@
 import { Response } from 'express'
 
 import { CycleRequest } from 'meta/api/request'
+import { UserFilters } from 'meta/tablePaginated'
+import { UserStatus } from 'meta/user'
 
 import { AssessmentController } from 'server/controller/assessment'
 import { UserController } from 'server/controller/user'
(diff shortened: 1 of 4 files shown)


src/server/api/user/getMany.ts
```diff
@@ -13,7 +15,8 @@ export const getMany = async (req: CycleRequest<{ print: string }>, res: Respons
 
     const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName })
 
-    let users = await UserController.getMany({ assessment, cycle, countryIso })
+    const filters: UserFilters = { statuses: [UserStatus.active, UserStatus.invitationPending] }
```
minotogna, line 18: Does this show invitations in admin users ? In admin we only want to see users , not invitations . Is it the case ?

minotogna, line 18: Plus , default should be taken ONLY in one place , in the getQuertParams...

sorja: Yes, we set the filters.statuses only in api/user/getMany not in api/admin/getUsers

sorja: This is not the default, this is wanted behavior only for this endpoint

Should we return always inivtationPending?

The previous implementation including invitationPending users as default was removed in favor of 

src/server/repository/public/user/utils/getPropsToQueryParams.ts
`const statuses = disabled ? [UserStatus.disabled] : [UserStatus.active]`

minotogna (approved):

PR #4199, opened 2025-01-07
4001 - Datatable use Datagrid - Cell borders

yaguzmang:
image

image

#

image

image

#

image
image

## ODP Header in print:
image


src/client/pages/Section/DataTable/Table/RowData/RowData.tsx
```diff
@@ -11,33 +14,52 @@ import Cell from './Cell'
...
-          lastCol={index === cols.length - 1}
-          lastRow={lastRow}
-          row={row}
-          rowIndex={Number(row.props.index)}
-          sectionName={sectionName}
-          table={table}
-        />
-      ))}
+      {cols.map((col, colIndex) => {
+        const lastCol = colIndex === cols.length - 1
```
minotogna, line 28: Can column count passed as prop be different than cols.length-1?(can we use one variable only ?)

Also , are lastSpanningCol and lastSpanningRow needed ?
With colSpan taken from Cols.getStyle and be default =0:
Lastcol = (colIndex + colSpan)===colsLength (see first point)
Can this work ?

src/client/pages/Section/DataTable/Table/hooks/useGridColumnCount.ts
```diff
@@ -0,0 +1,15 @@
...
+type Props = {
+  gridRef: MutableRefObject<HTMLDivElement>
+}
+
+export const useGridColumnCount = (props: Props): number => {
+  const { gridRef } = props
+  if (!gridRef?.current) return 0
+
+  const gridStyle = getComputedStyle(gridRef?.current)
+  const columnCount = gridStyle.getPropertyValue('grid-template-columns').split(' ').length
```
minotogna: Also , this might not work if there's a repeat for example ?

src/client/pages/Section/DataTable/Table/RowData/RowData.tsx
yaguzmang: Thinking about it not even `Lastcol = (colIndex + colSpan)===colsLength` is needed, because really whichever col is last in the data row array should always be `lastCol`, regardless of the span. The only way that wouldn't apply is if we have a row that has columns with spans that don't add up to the col count of the table, but that would be a data issue. 
It's similar when ` lastRow= index === rowsData.length - 1` , there is no need to check the spans for the last data row, as the cells there should always have lastRow=true. 

`lastSpanningRow` is necessary though because you could have a cell in the middle of the table that spans down to the last row. Unlike `lastSpanningCol` where a cell in the middle can't span to the last column.

src/client/pages/Section/DataTable/Table/hooks/useGridColumnCount.ts
yaguzmang: The computedStyle for this property is the pixel width value of each column, so for example here:
image
`gridTemplateColumns = 'repeat(3, 1fr)'`
computed style: `192.67px 192.67px 192.67px`

I removed this hook though as it wasn't necessary 👍

minotogna: Right 😅👌

minotogna (approved):

PR #4202, opened 2025-01-09
4001 - Datatable use Datagrid - Add first col sticky

yaguzmang:
TODO:
- [x] Finish testing on all tables

https://github.com/user-attachments/assets/44e62d36-7d89-4c4f-90db-400773c37b40

### DIFF
--- a/src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
+++ b/src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
@@ -5,6 +5,12 @@
   display: flex;
   justify-content: end;
 
+  &.firstCol {
+    left: 0;
+    position: sticky;
+    z-index: 1;
+  }
+
   &.readonly,
   &.calculated-input {
     cursor: default;
--- a/src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
+++ b/src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
@@ -8,11 +8,7 @@
   &.firstCol {
     left: 0;
     position: sticky;
-    z-index: 1;
-  }
-
-  .select__wrapper > .select__container > div:not(.select__control) {
-    z-index: 2;
+    z-index: $z-index-table-grid-sticky-cell;
   }
 
   &.readonly,
--- a/src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
+++ b/src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
@@ -11,6 +11,10 @@
     z-index: 1;
   }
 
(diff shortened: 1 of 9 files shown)


src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
```diff
@@ -11,6 +11,10 @@
     z-index: 1;
   }
 
+  .select__wrapper > .select__container > div:not(.select__control) {
+    z-index: 2;
+  }
+
```
yaguzmang: Fixes this for cells using Select:
#9 Ranked

to: 
#9 Ranked

minotogna: Maybe better to add it to the Select directly ? 
And we should use variables for the z-index .(see in partials ) 
What do you think ?

src/client/pages/Section/DataTable/Table/Table.scss
```diff
@@ -14,6 +14,7 @@
 
 .table-grid-container {
   margin-bottom: 28px;
+  overflow-x: auto;
```
minotogna, line 17: Didn't we add it already some place
Else ?

yaguzmang: Yes, we have it in `table-grid-container` so it doesn't stretch the page and in table-grid, so that it's only the grid moving and not the buttons:

https://github.com/user-attachments/assets/f244f408-9864-4022-ad41-00d96fa9df59

It's a bit different now that I removed some of the fra table wrapper divs.

src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
yaguzmang: Sure. I will add it to the Select component and then see nothing else is affected where it's used :+1:

src/client/pages/Section/DataTable/Table/hooks/useParsedTable.ts
```diff
@@ -57,7 +60,15 @@ export const useParsedTable = (props: Props): Returned => {
...
 
+    const firstColHeader = rowsHeader[0]?.cols[0]
+    let firstHeaderRowSpan = 0
+    if (!Objects.isEmpty(firstColHeader)) {
+      const { rowSpan } = Cols.getStyle({ col: firstColHeader, cycle })
+      firstHeaderRowSpan = rowSpan ?? 1
+    }
+
     return {
+      firstHeaderRowSpan,
```
minotogna, line 71: I understand this, it's a fix just for the first column . Let's talk later about this

minotogna (approved):

PR #4207, opened 2025-01-13
3855 - Check all files exist in S3

sorja:
dev note: Avoid run many times back to back - 
`// Note: Running this script many times may cause runners IP to timeout as protection mechanism`


src/test/s3Files/index.ts
```diff
@@ -0,0 +1,119 @@
...
+// eslint-disable-next-line no-promise-executor-return
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
+
+const _allFilesExist = async (files: Array<FileSummary>): Promise<boolean> => {
+  if (!TEST_S3_FILES) return true
+  const batches = Arrays.chunk(files, BATCH_SIZE)
+  const batchResults = await Promises.each(batches, async (batch) => {
+    return Promises.each(batch, async ({ uuid: key }) => {
+      const exists = await FileStorage.fileExists({ key })
+      if (!exists) Logger.info(`Missing file in S3: ${key}`)
```
sorja, line 39: Note: I deleted one file from S3 manually to see output from this command
All other files are there (and the file was there before manual deletion ;) )

`0077fe28-....-....-....-5809eb1d364c`

src/test/s3Files/index.ts
```diff
@@ -0,0 +1,119 @@
...
+    await sleep(1000)
+    const response = await axios({
+      method: 'GET',
+      url,
+      responseType: 'stream',
+      timeout: 5000,
+      validateStatus: (status) => status === 200,
+      headers: {
+        // Copy this from browser request and add it to .env - check .env.template
+        Cookie: ProcessEnv.authCookie,
```
minotogna: Should we remove from env template and ProcessEnv since it's used only here ? Whoever will execute the script , should know this .
Or what do you think ?

sorja: Sounds good

sorja:
@minotogna I tried figuring out the best way to test this - but I think in the end, the most important part is that the files exist in S3 - everything else should be reduntant because we didn't change other logic.
The URLs for files are equal and untouched, so if the node server routes correctly to S3 using the file uuid and the file exists, then there is no problems

I couldn't cause other issues in the platform other than a timeout from S3 from spamming requests

minotogna (approved):
👍

PR #4211, opened 2025-01-14
4001 - Datatable use Datagrid - First col auto width & cells min-width 

yaguzmang:
Examples of new layouts:

https://github.com/user-attachments/assets/9e538f32-c391-403f-90a7-c7d10303de72

Input resizing: 

https://github.com/user-attachments/assets/a12a0214-870f-48f0-9b46-bebc4f0fa566


src/client/pages/Section/DataTable/Table/hooks/useGridTemplateColumns.ts
```diff
@@ -22,6 +22,6 @@ export const useGridTemplateColumns = (props: Props): Returned => {
     const gridTemplateColumnsFromTable = table.props.style?.[cycle.uuid]?.gridTemplateColumns
     if (!Objects.isEmpty(gridTemplateColumnsFromTable)) return gridTemplateColumnsFromTable
 
-    return `auto repeat(${headers.length}, minmax(90px, 1fr))`
+    return `minmax(100px, auto) repeat(${headers.length}, minmax(90px, 1fr))`
```
yaguzmang: Prevents this case where the `auto` col is shrunk completely in smaller resolutions:

https://github.com/user-attachments/assets/3428ac8b-fd51-4cbe-a2ac-b6e0d0e2d3a8

src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
```diff
@@ -4,6 +4,7 @@
   align-items: center;
   display: flex;
   justify-content: end;
+  overflow: hidden;
```
minotogna: This is risky . Any reason for this ?

minotogna (commented):
Good improvement. Just one doubt

src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
yaguzmang: I noticed that this could happen in the prev implementation so I added the overflow hidden to prevent it in the current one:

https://github.com/user-attachments/assets/92031150-a505-4a42-96a3-d1c032ec9cbf

Should we just let it overflow in case? I can remove the line

minotogna: The cell should expand , all content should fit , and not be hidden . Can you check ?

yaguzmang: Seems like we don't handle this yet when the texts are contiguos:
image

So the same happens with numbers in the data grid. I will check if we can make the cols expand horizontally as well, or if we might need to break the text.

minotogna: Nope, don't break the text . Numbers should remain in one row

yaguzmang: I moved the min width 90px to the `table-grid__data-cell` class, and added `minmax(min-content, 1fr))` to the gridTemplateColumns. This was the only way I could find to have the column absolute min-width be 90px, while expanding the width to the content and defaulting to 1 fraction if not necessary.

https://github.com/user-attachments/assets/bf65e3b2-bc44-4436-8900-e76130eda224

However, since the `input` in the datacell doesn't have a set width (unlike a div when read only is enabled), its width defaults to the min 90px.

I was looking into having the input resize based on its content using css, but unfortunately the property that does that isn't supported in Firefox and Safari: 
```
  .input-text {
    field-sizing: content;
  }
```
This is how it looks for reference:

https://github.com/user-attachments/assets/ce52f419-1c94-464c-ae99-dc92293f01f2

It could be accomplished with js though. What do you think?

(Marking the PR as draft since I need to edit and retest the gridTemplateCols in the migrations after we land on a solution 👍 )

minotogna: CSS would be preferred, there should be a way . 
What if you put width 100% to the input ? Is it the only problem left ?

yaguzmang: It has width 100%, but it doesn't affect (setting a pixel specific width does). I'm trying a couple other things now but it seems a dynamic width doesn't apply.
And yes, this is the only problem left with the gridTemplateCols

yaguzmang: Added resizing and updated the gridTemplateCols 👍 : 

https://github.com/user-attachments/assets/a12a0214-870f-48f0-9b46-bebc4f0fa566

src/client/components/Inputs/InputText/hooks/useResizeStyle.ts
```diff
@@ -0,0 +1,37 @@
...
+type Returned = Pick<CSSProperties, 'minWidth'> | undefined
+
+export const useResizeStyle = (props: Props): Returned => {
+  const { disabled, inputRef, resize, value } = props
+
+  const valueLength = (value ?? '').toString().length
+
+  const [paddingPx, setPaddingPx] = useState(0)
+
+  useEffect(() => {
```
minotogna: Can we useLayoutEffect here ?

src/client/components/Inputs/InputText/hooks/useResizeStyle.ts
```diff
@@ -0,0 +1,37 @@
...
+    const paddingLeft = parseFloat(computedStyle.paddingLeft) || 0
+    const paddingRight = parseFloat(computedStyle.paddingRight) || 0
+    setPaddingPx(paddingLeft + paddingRight)
+  }, [disabled, inputRef, resize])
+
+  const style = useMemo<Returned>(() => {
+    if (!resize) return undefined
+    if (valueLength === 0) return undefined
+
+    return { minWidth: `calc(${valueLength}ch + ${paddingPx}px)` }
```
minotogna, line 33: So it is working ! ☺️💪

src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
```diff
@@ -4,6 +4,7 @@
   align-items: center;
   display: flex;
   justify-content: end;
+  min-width: 90px;
```
minotogna, line 7: Can this go in the gridTemplateColumns or should it be here ?

minotogna (commented):
Nice 😊

src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss
yaguzmang: It needs to be here so that we can use `min-content` in the gridTemplate cols.
 It wasn't possible to do it all in the template cols with something like `minmax(90px, min(min-content, 1fr))` as it isn't valid this way.

minotogna (approved):

PR #4230, opened 2025-01-18
4001 - Datatable use Datagrid - Editable borders corrections

yaguzmang:
Fixing cell border rendering when adjacent cells are non-editable:

https://github.com/user-attachments/assets/fb5004aa-9b6d-4f88-8255-d5fb1d58416f

The first approach which is consistent with how the borders are currently done (top,left for each cell) doesn't quite work. Since some cells span multiple columns, setting the border top for those can show and editable border section where it shouldn't be:

https://github.com/user-attachments/assets/0ecb8a80-3ba0-4c76-9bf1-867c1d316ec2

Also, because some cells have a non-transparent background color, the outline of the focused editable cell is hidden. Fixing that with z-index wasn't possible because the sticky headers should always have a higher z-index. 

Due to that, the best approach was to have some double borders, which doesn't create those issues and helps distinguish more clearly editable vs readonly cells.

### DIFF
--- a/src/client/components/DataGrid/DataCell/DataCell.scss
+++ b/src/client/components/DataGrid/DataCell/DataCell.scss
@@ -79,7 +79,7 @@ $highlight-top-bottom: inset 0 1px $ui-accent, inset 0 -1px $ui-accent;
   border-left: $border-editable;
   border-top: $border-editable;
 
-  &.bottomEditable {
+  &.lastEditableRow {
     border-bottom: $border-editable;
   }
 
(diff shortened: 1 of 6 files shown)


src/client/components/DataGrid/DataCell/DataCell.scss
```diff
@@ -79,6 +79,10 @@ $highlight-top-bottom: inset 0 1px $ui-accent, inset 0 -1px $ui-accent;
   border-left: $border-editable;
   border-top: $border-editable;
 
+  &.bottomEditable {
```
minotogna: let's call those classes lastEditableRow, lastEditableCol (and add them next to lastCol ,...)

minotogna (approved):

PR #4238, opened 2025-01-23
4001 - Datatable use Datagrid - Bug fix: handle null values in sanitizer functions

yaguzmang:
Small bug fix after doing the final testing round for all tables. After this the main PR will be opened. 

This fixes the case where the node raw value was not defined, so the input value was null and the sanitizer function was run onBlur with that empty value, causing `newValue.trim()` to fail.

### DIFF
--- a/src/client/utils/sanitizer/decimal.ts
+++ b/src/client/utils/sanitizer/decimal.ts
@@ -1,7 +1,8 @@
 import { Numbers } from 'utils/numbers'
 import { Objects } from 'utils/objects'
 
-export const acceptableAsDecimal = (newValue: string) => {
+export const acceptableAsDecimal = (newValue: string | null) => {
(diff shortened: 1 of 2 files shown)


src/client/utils/sanitizer/decimal.ts
```diff
@@ -1,7 +1,8 @@
 import { Numbers } from 'utils/numbers'
 import { Objects } from 'utils/objects'
 
-export const acceptableAsDecimal = (newValue: string) => {
+export const acceptableAsDecimal = (newValue: string | null) => {
+  if (Objects.isNil(newValue)) return false
```
minotogna, line 5: When do we pass null ?

minotogna (approved):

minotogna (approved):

PR #4242, opened 2025-01-24
4204 - FRA Kiosk: Latest Activities map - Markers

yaguzmang:
https://github.com/user-attachments/assets/65b3f4ad-61ed-47d9-8694-66ade4d01b1c


src/server/static/kiosk/latestActivities.csv
```diff
@@ -0,0 +1,6 @@
...
+2024-10-15,3.4516,76.532,Colombia,COL,"The Food and Agriculture Organization of the United Nations (FAO) together with Griffith University, the International Union for Conservation of Nature (IUCN), the United Nations Forum on Forests (UNFF) Secretariat, and the Wild Heritage Foundation, launched a ground-breaking conservation initiative at the 2024 United Nations Biodiversity Conference (CBD COP16): the project “Strengthening Conservation of Primary Forests through Partnership Enhancement and Coordination of Support,” funded by the Global Environment Facility (GEF). The project’s contribution to improve monitoring and reporting on the extent of primary forests was also shared at the 2024 UN Climate Change Conference (UNFCCC COP 29)."
```
yaguzmang: As seen in the video, some coordinates are not correct (like Canada with coordinates somewhere in China, or Colombia with coordinates in Asia). We should ask for an updated version of the file.

minotogna: And they are not inverted ? Have you tried ?

src/client/pages/Kiosk/LatestActivities/Map/hooks/useFetchAndMarkActivities.ts
```diff
@@ -0,0 +1,76 @@
...
+    const fetchCSVData = async () => {
+      setIsLoading(true)
+      setError(null)
+
+      try {
+        const response = await axios.get(ApiEndPoint.Kiosk.latestActivities(), {
+          responseType: 'text',
+        })
+
+        Papa.parse(response.data, {
```
minotogna: Maybe the server can return the data in the format the client wants ?

src/server/static/kiosk/latestActivities.csv
yaguzmang: After looking into the event cities, I realized the problem was some missing minus signs. Fixed now :+1:

src/client/pages/Kiosk/LatestActivities/Map/hooks/useFetchAndMarkActivities.ts
yaguzmang: Added and removed papaparse as it is no longer needed 👍

minotogna (approved):

Issue #4221, opened 2025-01-16
Github test action: Init database


minotogna:
- [ ] Add metadata and data from fra/panEuropean 2020/2025 (from review)
- [ ] Create a flexible script to easily recreate the db once 2025 prod is published

PR #4246, opened 2025-01-24
4245 - E2E Testing framework and test runner in Github Actions

### LINKED ISSUE #4142: Separate commands for test:integration test:unit
Currently no matter what you run, everything is executed.

(Easily fixed by fixing e.g. the paths for the commands.
This issue is more of a note so I can keep track)

### LINKED ISSUE #4245: Use docker-compose for CI E2E

sorja:
Resolves #4142

todo
- [x] partial db copy


minotogna (approved):

.github/workflows/test.js.yml
```diff
@@ -1,86 +1,101 @@
...
+#   uses: mxschmitt/action-tmate@v3
+#   with:
+#     limit-access-to-actor: true
+#     timeout-minutes: 30
 
 on:
   push:
     branches-ignore:
       - 'master'
       - 'production'
```
minotogna: I might need to update the automatic rules in branch settings.

I would propose, the build action should run on all branches, expect development .
Test action only on development .
What do you think @sorja @yaguzmang ?

(ps requesting changes now only to avoid accidental merge)

minotogna (changes requested):

minotogna:
Resolves #4221

.github/workflows/test.js.yml
yaguzmang: > Test action only on development 

Yes, I think that's a good idea, <s>otherwise we might run into S3 transfer limits over a given month</s>. Maybe also we can have it run for opened PRs that point directly to development? then we see if there are any issues before merging. 

Something like:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore

sorja: This sounds good

Regarding S3 limit - it shouldn't happen?
I tried adding the DB to cache to avoid firing requests to S3 when not needed
I added manual refetching workflow trigger for updating the db

image

sorja: Let me know @minotogna what you think about

`pull_request_target: development`

minotogna: sounds good @sorja 
test only on dev. build all other branches . what do u think ?

sorja: sounds good

.github/workflows/test.yml
```diff
@@ -0,0 +1,72 @@
+name: Run Tests
+
```
minotogna: Good you made two actions . Should call the files test-unit and test-e2e ?

jest.config.integration.js
```diff
@@ -0,0 +1,7 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const baseConfig = require('./jest.config.base')
+
+module.exports = {
+  ...baseConfig,
+  testMatch: ['**/integration.test.ts'],
+}
```
minotogna, line 7: Why all this jest config files ? Can't we have only one ?

package.json
```diff
@@ -27,12 +27,13 @@
     "run:prod:server": "node dist/server/start",
     "start": "run-s migration-public:run run:prod:server",
     "//test": "mochapack --webpack-config webpack.config.babel.test.js",
-    "test": "run-s test:integration",
-    "test:integration": "yarn jest --no-cache --detectOpenHandles --setupFiles dotenv/config",
+    "test": "run-p test:unit test:integration",
+    "test:unit": "yarn jest --no-cache --detectOpenHandles --setupFiles dotenv/config --config jest.config.unit.js",
+    "test:integration": "yarn jest --no-cache --detectOpenHandles --setupFiles dotenv/config --config jest.config.integration.js src/test/integration/integration.test.ts",
```
minotogna, line 32: Why test unit and test integration ?

.github/workflows/test.yml
sorja: its running both integration and unit tests

jest.config.integration.js
sorja: we cant for our setup (unit tests scattered accross the project, integration single entry point)

package.json
sorja: two different tests - i dont want to run integration tests if i am running / testing unit tests and other way around

separation of concerns

.github/workflows/test.yml
```diff
@@ -0,0 +1,72 @@
...
+on:
+  pull_request:
+    branches-ignore:
+      - 'master'
+      - 'production'
```
minotogna: should we run this action on ALL branches, except development ?
What do you think?

src/test/e2e/jest-playwright.config.ts
```diff
@@ -28,18 +28,7 @@ const config: PlaywrightTestConfig = {
     },
   ],
```
minotogna, line 30: can we omit the jest prefix from playwright config files ?

minotogna (commented):
👍

.github/workflows/test.yml
sorja: Do you mean like so?

```
on:
  pull_request:
    branches-ignore:
      - 'development'
```

minotogna: Yes

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,96 @@
+name: Run E2E Tests
+
```
minotogna: Should we call this action and file test all ?

.github/workflows/test.yml
```diff
@@ -0,0 +1,71 @@
+name: Run Tests
```
minotogna: And this test partial ? Or test unit/integration ?
I mean some meaningful names that tells with more precision what they do. What do you think ?

sorja: Sure, I will revisit the names

sorja: image

minotogna (approved):

PR #4280, opened 2025-01-28
4249 - Data History: Descriptions text comparison (Empty dataset)

sorja:
https://github.com/user-attachments/assets/a60a6022-1579-44a1-872e-c9fa02a94242

- [x] Add endpoint (same than descriptions with added /history ?) that checks : if Cycle doesn't have previous cycle returns empty value 
- [x] add the value to the store 
- [x] show differences in UI (green)


src/client/components/DiffText/DiffText.tsx
```diff
@@ -0,0 +1,39 @@
...
+          </React.Fragment>
+        )
+      })}
+    </div>
+  )
+}
+
+DiffText.defaultProps = {
+  className: '',
+}
```
minotogna: I think we should start considering to not add defaultProps anymore. we should remove the eslint rule ? what do you think ?

src/client/pages/Section/hooks/useGetDescriptionValues.ts
```diff
@@ -16,10 +16,14 @@ export const useGetDescriptionValues = (props: Props): void => {
...
+  const historyLastApprovedIsActive = useHistoryLastApprovedIsActive()
 
   useEffect(() => {
     if (!print) {
       dispatch(DataActions.getDescription({ countryIso, assessmentName, cycleName, sectionName }))
     }
-  }, [assessmentName, countryIso, cycleName, dispatch, print, sectionName])
+    if (!print && historyLastApprovedIsActive) {
+      dispatch(DataActions.getDescriptionsHistory({ countryIso, assessmentName, cycleName, sectionName }))
+    }
```
minotogna: I would add a separate hook to not alter original fetch . what do u think ?
Otherwise every time user is switching history last approved, (remaining in the same page) this hook will re-trigger unnecessary data fetch (line 23), 
They can be two parallel fetches.

What do u think?

src/client/store/data/selectors/index.ts
```diff
@@ -23,6 +25,23 @@ const getHistoryItems = createSelector(getHistoryActivities, (history) => histor
 const getHistoryLastApproved = createSelector(getHistory, (history) => history.lastApproved ?? {})
 const isHistoryLastApprovedActive = createSelector(getHistoryLastApproved, (history) => Boolean(history?.active))
 
+const getLastApprovedDescriptions = createSelector(
+  [
+    getHistory,
```
minotogna: here you can use getHistoryLastApproved selector directly

src/client/components/DiffText/DiffText.tsx
sorja: yes, definitely, i will check

minotogna (approved):

PR #4286, opened 2025-01-29
4279: Cycle Source reference

### LINKED ISSUE #4279: Cycle Source Reference
Add new column to `assessment_cycle`
`cycle_uuid_source: uuid`

- [ ] Migrate cycles

minotogna:
After migration steps and cloning:
image

image

### DIFF
--- a/src/test/migrations/steps/20250129053400-step-add-cycle_source_uuid-column.ts
+++ b/src/test/migrations/steps/20250129053400-step-add-cycle_source_uuid-column.ts
@@ -1,0 +1,44 @@
+import { Objects } from 'utils/objects'
+
+import { CycleName } from 'meta/assessment'
+
+import { AssessmentController } from 'server/controller/assessment'
+import { BaseProtocol, DB } from 'server/db'
+
+const mapping: Record<CycleName, CycleName> = {
+  '2025': '2020',
+  latest: '2025',
+}
+
+export default async (client: BaseProtocol) => {
+  await DB.query(`
+      alter table public.assessment_cycle
+          add cycle_uuid_source uuid
+              constraint assessment_cycle_assessment_cycle_source_fk
+                  references public.assessment_cycle (uuid)
+                  on update cascade on delete set null;
+  `)
+
+  const assessments = await AssessmentController.getAll({}, client)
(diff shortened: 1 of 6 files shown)


src/test/migrations/steps/20250129053400-step-add-cycle_source_uuid-column.ts
```diff
@@ -0,0 +1,44 @@
...
+      alter table public.assessment_cycle
+          add cycle_uuid_source uuid
+              constraint assessment_cycle_assessment_cycle_source_fk
+                  references public.assessment_cycle (uuid)
+                  on update cascade on delete set null;
+  `)
+
+  const assessments = await AssessmentController.getAll({}, client)
+  await Promise.all(
+    assessments.map((assessment) =>
```
sorja, line 24: Not sure if using `client (tx)` and `DB` mixed can cause problems?
`client` is tx outside of `DB`

In some cases changes from with `DB` might not be visible inside `tx`

sorja (dismissed):

src/test/migrations/steps/20250129053400-step-add-cycle_source_uuid-column.ts
minotogna: not really. always the same thing @sorja 
DDL cannot run within transition . client is a transition. DB no

sorja: Exactly what I meant, should use `DB` for all or `DB.tx` after `DB.query`

minotogna: why? I'd rather execute DML in transactions

minotogna: I'm not updating data, I don't need a transaction starting from that. the DML is immediately visible

minotogna: @sorja please have another review. had to update the create schema

sorja (approved):

PR #4295, opened 2025-01-30
4291 - 4288 - 4289 - Fix Editing permission with status 'notStarted'

### LINKED ISSUE #4288: DataTable: Review indicator missing
Image

### LINKED ISSUE #4289: Review FRA latest - Table twist
@minotogna 
**Describe the bug**
Something wrong with the table 1a tier table,  table 2a tier table, table 2c tier table, table 3b, 5a, 5c, 6b, 7, 8

**Screenshots**
 table 1a tier table:
Image
 table 2a tier table:
Image
table 2c tier table
Image
table 3b
Image
table 5a
Image
table 5c
Image
table 6b
Image
table 7
Image
table 8
Image

### LINKED ISSUE #4291: Review FRA latest - editing mode unavailable
@minotogna 

**Describe the bug**
For admin seems ok to edit the content for country reports in the review FRA latest.

But are users(such as NCs) allowed to edit in the latest FRA review? Seems not even able to enter information in the profile interface, not to mention the tables.

**Screenshots**
Admin: 

Image

NC :
Image

Image

yaguzmang:
resolves #4291 
resolves #4289 
resolves #4288 

Table layouts:
Adding a piece of logic to update the gridTemplateColumns of the tables when adding a new column. This fixes the layout issues after cloning for tables 'disturbances' and 'affectedAreaByFire'. The rest of the tables mentioned in #4289 are fixed by cloning after merging the DataGrid PR.

image

image

Authorizer fixes: 
Since the cloned cycles don't have a 'status' prop, the check to show the review indicator was always returning false. This is fixed by using `status = Areas.getStatus(country)` to get the status, which is consistent with how we use to display the Status Indicator.

TODO: 
Fix the `canViewReview` tests, which are failing even though the platform is working as expected. 
For example this test fails but it works in the platform:
`✕ should return true for administrator in review status`

https://github.com/user-attachments/assets/f9b5356d-0e87-46ce-8bcc-ef31cd8c6108

Resolves #4288

### DIFF
--- a/src/meta/area/areas.ts
+++ b/src/meta/area/areas.ts
@@ -16,7 +16,7 @@ const isISOGlobal = (isoCode: string): boolean => isoCode === Global.WO
 const isRegion = (isoCode: string): boolean => Object.values(RegionCode).includes(isoCode as RegionCode)
 const isFRARegion = (isoCode: string): boolean => fraRegionCodes.includes(isoCode as RegionCode)
 const getStatus = (country: Country): AssessmentStatus => {
-  if (!country?.lastEdit) return AssessmentStatus.notStarted
+  if (!country?.props?.status && !country?.lastEdit) return AssessmentStatus.notStarted
   if (!country?.props?.status && country?.lastEdit) return AssessmentStatus.editing
   return country?.props?.status
 }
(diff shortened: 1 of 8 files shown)


src/meta/area/areas.ts
```diff
@@ -16,7 +16,7 @@ const isISOGlobal = (isoCode: string): boolean => isoCode === Global.WO
 const isRegion = (isoCode: string): boolean => Object.values(RegionCode).includes(isoCode as RegionCode)
 const isFRARegion = (isoCode: string): boolean => fraRegionCodes.includes(isoCode as RegionCode)
 const getStatus = (country: Country): AssessmentStatus => {
-  if (!country?.lastEdit) return AssessmentStatus.notStarted
+  if (!country?.props?.status && !country?.lastEdit) return AssessmentStatus.notStarted
```
yaguzmang: It can happen that the status is set to `review` but the `lastEdit` is not in sync and hasn't been refreshed. This would cause this function to return `notStarted` even though there is a set status. 
This change fixes that.

minotogna: Well that is the root problem . We should fix that . I'll take a look later as well

minotogna (approved):

PR #4314, opened 2025-01-31
4313 - Data History auth: if (user is admin or reviewer) and status !== notStarted

minotogna:


src/meta/user/authorizer/index.ts
```diff
@@ -217,25 +217,15 @@ const canViewHistory = (props: {
...
- * @param props
- * @param props.country
- * @param props.cycle
- * @param props.user
- * @returns boolean
+ * returns true if (user is admin or reviewer) and status !== notStarted
  */
 const canViewHistoryLastApproved = (props: { country: Country; cycle: Cycle; user: User }): boolean => {
   const { country, cycle, user } = props
   const { status } = country?.props ?? {}
```
yaguzmang: The status being `undefined` shouldn't be an issue anymore, but should we now use `const status = Areas.getStatus(country)`  here for consistency?

src/meta/user/authorizer/index.ts
```diff
@@ -217,25 +217,15 @@ const canViewHistory = (props: {
...
- *
- * Reviewer: if country status = review
- * Administrator: if status >=review
- *
- * @param props
- * @param props.country
- * @param props.cycle
- * @param props.user
- * @returns boolean
+ * returns true if (user is admin or reviewer) and status !== notStarted
```
sorja, line 220: should we use jsdoc? (bonus: we can generate doc later when releasing API?)

minotogna: sure. let's see later on.

yaguzmang (approved):

PR #4317, opened 2025-02-02
4253 - Data History: ODP comparison (empty dataset) - Grid table diffs

yaguzmang:
(UPDATED)

https://github.com/user-attachments/assets/ccc88859-169a-4e01-b1f7-5d7c3a9d3a19

Coming PR:
Diffs for the tables that are not using the new DataGrid:
image


src/client/pages/OriginalDataPoint/components/DiffView/DiffView.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+import { OriginalDataPoint } from 'meta/assessment'
+
+import { useFieldChanges } from 'client/pages/OriginalDataPoint/hooks/useFieldChanges'
+import CellDiff from 'client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/HistoryCompare/CellDiff'
```
yaguzmang: Should we make this component a common DataGrid component like DataCell? it is already generic

minotogna: I have the impression this component should not exist.

The Grid Cell should be rendered outside and use the common Diff already existing ?

src/client/pages/OriginalDataPoint/components/NationalClasses/components/NationalClass.tsx
```diff
@@ -45,24 +49,37 @@ const NationalClass: React.FC<Props> = (props) => {
...
-          onPaste={onPasteName}
-          placeholder={placeHolder && index === 0 ? t('nationalDataPoint.enterOrCopyPasteNationalClasses') : ''}
-          value={name ?? ''}
-        />
-      </DataCell>
+      {historyLastApprovedIsActive ? (
+        <DiffView lastRow={lastRow} originalDataPoint={originalDataPoint} path={['nationalClasses', index, 'name']} />
+      ) : (
+        <DataCell error={error} lastRow={lastRow}>
+          <InputText
```
minotogna, line 56: Like here . The conditional rendering can happen here , so that the grid cell is already rendered

src/client/store/ui/originalDataPoint/hooks/index.ts
```diff
@@ -22,6 +22,9 @@ export const useOriginalDataPoint = (): OriginalDataPoint => {
   return originalDataPoint ?? originalDataPointTemplate
 }
 
+export const useOriginalDataPointHistory = (): OriginalDataPoint | undefined =>
+  useAppSelector((state) => state.ui.originalDataPoint?.history)
```
minotogna: I would put this under history books , selector and data

src/client/pages/OriginalDataPoint/components/DiffView/DiffView.tsx
sorja: We should use DiffText (exists)
What I wonder is if the hooks `useChange`'s should be also near this component?
E.g.

`const changes = useChangesHTML(a,b)`
`const changes = useChangesNumeric(a,b)`
`const changes = useChangesTable(a,b)`
...
etc?

Or should we - like we do now - get the changes where we use the component?

minotogna: Wait guys . 
We have two components DiffText and CellDiff 

From what I have seen in the ODP we need to use DiffText , not CellDiff

You can have a common ODPDiffText that takes the path , get the odp changes and render DiffText .
This way you are able to use this in the other odp tables

yaguzmang: The reason why I used CellDiff and not DiffText here is because CellDiff can show differences per line. If we use DiffText and there are more than one Methods selected, it would look like this:

https://github.com/user-attachments/assets/5fe19b13-8e4a-4186-8f5c-e7777b167064


I took `HistoryCompare` data sources -> FRA Variable as an example:
image

 I can introduce `DiffLines` and render that or `DiffText` depending on the ODP field, what do you guys think?

src/client/store/ui/originalDataPoint/hooks/index.ts
yaguzmang: image
You mean together here in data -> history -> lastApproved -> originalDataPoint, right?

minotogna: Yes . Record of assessmentName-cycle name-countryIso-year->odp

src/client/pages/OriginalDataPoint/components/DiffView/DiffView.tsx
minotogna: Why don't we see diff lines in DiffText ?

minotogna: Can we make one diff text generic component? I thought we already had it

src/client/store/ui/originalDataPoint/hooks/index.ts
yaguzmang: image
Added and moved the action as well for consistency :+1:

src/client/pages/OriginalDataPoint/components/DiffView/DiffView.tsx
yaguzmang: Added a small fix to DiffText and it now can also shows the diff lines. So now CellDiff is removed, conditional render fixed, and renamed DiffView to ODPDiffText 👍

src/client/pages/OriginalDataPoint/hooks/useGetOriginalDataPointHistory.ts
```diff
@@ -0,0 +1,33 @@
...
+  const historyLastApprovedIsActive = useHistoryLastApprovedIsActive()
+
+  useEffect(() => {
+    if (year === '-1') return
+    if (!historyLastApprovedIsActive) return
+
+    dispatch(
+      DataActions.getOriginalDataPointHistory({
+        assessmentName,
+        countryIso: countryIso as CountryIso,
```
minotogna: Can u pass a type to the hook to avoid cast here ?

src/client/store/data/state.ts
```diff
@@ -63,6 +64,7 @@ export type HistoryActivitiesState = {
 export type HistoryLastApprovedState = {
   active?: boolean
   descriptions?: DescriptionsState
+  originalDataPoint?: RecordAssessmentOriginalDataPoint
```
minotogna: Can we use plurals ?

src/server/api/cycleData/originalDataPoint/getOriginalDataPointHistory.ts
```diff
@@ -0,0 +1,29 @@
...
+      // Example of the rest of the object:
+      // dataSourceAdditionalComments: '',
+      // dataSourceMethods: [],
+      // dataSourceReferences: '',
+      // description: '',
+      // nationalClasses: [],
+      // values: {},
+    }
+
+    Requests.send(res, odpHistory)
```
minotogna, line 22: Can this return an empty object ?

src/client/pages/OriginalDataPoint/hooks/useFieldChanges.ts
```diff
@@ -0,0 +1,54 @@
...
+  originalDataPoint: OriginalDataPoint
+  path: Array<string | number>
+}
+
+type Returned = Array<Change>
+
+const _getSourceMethodText = (values: Array<string> | undefined, t: TFunction): string =>
+  (values ?? [])?.map((value) => t(`nationalDataPoint.dataSourceMethodsOptions.${value}`)).join('\n\r')
+
+const _getHtmlTextContent = (html: string): string => {
```
minotogna: Do we have a utility DOMs?

minotogna (approved):

PR #4321, opened 2025-02-03
4304 - Data History: Data table (no ODP) comparison (Prev cycle)

### LINKED ISSUE #4304: Data History: Data table (no ODP) comparison (Prev cycle)
Note: When comparing table data with previous cycles
if table didn't exist in prev cycle, then empty dataset.

if some columndidn't exist in prev cycle, then do not show diffs. show diffs only for current cycle columns

sorja:
image

https://github.com/user-attachments/assets/d5eabe43-34af-4c7d-a163-2e073c08b5db

Draft?:
- [x] Handle select table component: displaying value through i18n
- [ ] Ignore values that don't exist in current table schema/metadata (e.g. removed / added cols x rows)
- [ ]

### DIFF
--- a/src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.scss
+++ b/src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.scss
@@ -1,0 +1,6 @@
+// Adjust calculated cell padding
+.table-grid__data-cell.readonly {
+  .input-text.disabled.table-grid__data-cell-input-number {
+    padding: unset;
+  }
+}
--- a/src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.scss
+++ b/src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.scss
@@ -1,6 +1,0 @@
-// Adjust calculated cell padding
-.table-grid__data-cell.readonly {
-  .input-text.disabled.table-grid__data-cell-input-number {
-    padding: unset;
-  }
-}
--- a/src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.tsx
+++ b/src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.tsx
@@ -1,11 +1,12 @@
 import './History.scss'
 import React from 'react'
 
+import classNames from 'classnames'
+
 import { CountryIso } from 'meta/area'
-import { Cols, Cycles, NodeValue } from 'meta/assessment'
+import { Cols, NodeValue } from 'meta/assessment'
 import { RecordAssessmentDatas } from 'meta/data'
 
-import { useAssessment, useCycle } from 'client/store/assessment'
 import { useLastApprovedHistoryTableData } from 'client/store/data/hooks/useLastApprovedHistoryTableData'
 import { useCountryRouteParams } from 'client/hooks/useRouteParams'
 import DiffText from 'client/components/DiffText'
@@ -15,19 +16,12 @@ import { useChanges } from './hooks/useChanges'
 
 const History: React.FC<PropsCell> = (props) => {
   const { nodeValue, table, col, row } = props
-  const { assessmentName, /* cycleName: cycleNameParam, */ countryIso } = useCountryRouteParams<CountryIso>()
-  const assessment = useAssessment()
-  const cycle = useCycle()
+  const { assessmentName, cycleName, countryIso } = useCountryRouteParams<CountryIso>()
   const tableName = table.props.name
   const { colName } = col.props
   const { variableName } = row.props
   const isNumeric = Cols.isNumeric(col)
   const data = useLastApprovedHistoryTableData()
-
-  // TODO: Depending on HistoryLastApprovedInfo, cycleName is either current cycleName or prevCycleName
-  // const info: HistoryLastApprovedInfo = useHistoryLastApprovedInfo()
-  const prevCycle = Cycles.getPreviousCycle({ assessment, cycle })
-  const cycleName = prevCycle?.name
 
   const nodeValueProps = { assessmentName, cycleName, countryIso, tableName, colName, variableName, data }
   const nodeValueA = RecordAssessmentDatas.getNodeValue(nodeValueProps) ?? ({} as NodeValue)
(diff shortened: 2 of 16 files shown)


src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.tsx
```diff
@@ -15,35 +19,55 @@ import { PropsCell } from '../props'
...
+  const { assessmentName, /* cycleName: cycleNameParam, */ countryIso } = useCountryRouteParams<CountryIso>()
+  const assessment = useAssessment()
+  const cycle = useCycle()
   const tableName = table.props.name
   const { colName } = col.props
   const { variableName } = row.props
+  const isNumeric = useIsNumeric(col)
   const data = useLastApprovedHistoryTableData()
+
+  // TODO: Depending on HistoryLastApprovedInfo, cycleName is either current cycleName or prevCycleName
```
sorja: Upcoming PRs when we use HistoryLastApprovedInfo for prevCycle vs lastApproved ts

minotogna: why would we need previous cycle here ?

sorja: is in format RecordAssessmentData

sorja: should we return previous cycle data under the current cycle key?

minotogna: yes. basically you need to compare that datable with the same keys in the history last approve slice.
The BE will returns the object. the client should just render the diffs. it shouldn't care the prev cycle or not.

src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.scss
```diff
@@ -0,0 +1,6 @@
+// Adjust calculated cell padding
+.table-grid__data-cell.readonly {
+  .input-text.disabled.table-grid__data-cell-input-number {
+    padding: unset;
+  }
+}
```
minotogna: this is a bit strange. the layout should remain as it is? maybe in the history cell some class is wrong?
do we use the same cell class name ?

src/meta/assessment/cycles.ts
```diff
@@ -1,9 +1,16 @@
+import { Assessment } from './assessment'
 import { Cycle, CycleStatus } from './cycle'
 
 const isPublished = (cycle: Cycle): boolean => {
   return cycle.props.status === CycleStatus.published
 }
 
+const getPreviousCycle = (props: { assessment: Assessment; cycle: Cycle }): Cycle | undefined => {
```
minotogna, line 8: can we use this in HistoryLastApproveController ? (where find prev cycle. it's this code)

src/server/api/cycleData/table/getTableDataHistory.ts
```diff
@@ -1,18 +1,29 @@
...
   try {
-    Requests.send(res, {})
+    const { countryIso, assessmentName, cycleName, tableNames } = req.query
+    const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName })
+    const info = await CycleDataController.History.LastApproved.getInfo({ assessment, cycle, countryIso })
+
+    let tableData = {}
+    if (!Objects.isNil(info)) {
+      const props = { assessment, cycle, info, countryISOs: [countryIso], tableNames }
+      tableData = await CycleDataController.TableData.getTableDataLastApproved(props) // getTableDataLastApproved -> import './getTableData' -> filter tableNames + replace cycle
```
minotogna: we can get rid of this comment I guess. even next pr.

src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.scss
sorja: image
image

Without this, it gives this kind of larger padding, probably the padding is set in the row or parent

yaguzmang: Having a similar situation in my PR. In input cells, the input itself has a padding. That padding is not set from the data cell because the input should take the entire space for the borders to work.

minotogna: In this case @sorja handle the calculated cell style as the number input style

src/meta/assessment/cycles.ts
sorja: added

src/client/pages/Section/DataTable/Table/RowData/Cell/History/History.scss
sorja: image

(includes calculated)
image


It is now like that - but the padding comes from outside

minotogna: That must be fixed . The padding outside should be consistent with numeric cells.
Why numeric cells don't have this extra padding in history mode ?

minotogna: We have useClassName in the cell if I recall correctly .

sorja: https://github.com/openforis/fra-platform/blob/358853e0a5308fda16a22f4c7da1d968f371171c/src/client/pages/Section/DataTable/Table/RowData/Cell/Cell.scss#L3-L22

It is coming from here

```
  &.readonly {
    padding: $spacing-xxs;
  }
```

If we remove this:

image

When history is on:
image


@yaguzmang do you have any suggestions?

minotogna: @sorja I had a look at the code. the issue is in your classes. you need to add padding only if it's decimal or integer cell. not when it's calculated.

minotogna: I can push the fix if you wish

minotogna (approved):

PR #4328, opened 2025-02-04
4253 - Data History: ODP partial

yaguzmang:
Tested cells editing/pasting and data export comparing against review platform, for different years and countries:

https://github.com/user-attachments/assets/667b1550-e0a8-47d1-a84f-8933fe99d815

review/local
image

image
---

image

image

---
image

image

---
image

image


src/meta/assessment/originalDataPoint/odps/nationalClassUtils.ts
```diff
@@ -1,13 +1,13 @@
...
 
-export const calculateNationalClassOtherLandPercent = (nc: ODPNationalClass): BigNumber | null => {
-  if (Objects.isEmpty(nc.forestPercent) || Objects.isEmpty(nc.otherWoodedLandPercent)) return null
-
-  return Numbers.sub(100, Numbers.add(nc.forestPercent, nc.otherWoodedLandPercent))
+export const calculateNationalClassOtherLandPercent = (nc: ODPNationalClass): string | null => {
+  if (!Objects.isEmpty(nc.forestPercent) || !Objects.isEmpty(nc.otherWoodedLandPercent)) {
+    return Numbers.format(Numbers.sub(100, Numbers.add(nc.forestPercent ?? 0, nc.otherWoodedLandPercent ?? 0)), 3)
+  }
+  return null
```
yaguzmang, line 10: Updating this function to accomplish the same behavior as before:
image

src/client/pages/OriginalDataPoint/components/ExtentOfForest/hooks/useCalculatedValues.ts
```diff
@@ -20,9 +20,11 @@ export const useCalculatedValues = (props: Props): Returned => {
 
   return useMemo<Returned>(() => {
     const totalArea = Numbers.format(ODPs.calcTotalArea({ originalDataPoint }))
-    const totalForestPercent = Numbers.format(originalDataPoint.values.totalForestArea)
+    const totalForestPercent = Numbers.format(ODPs.calcTotalFieldArea({ originalDataPoint, field: 'forestPercent' }))
```
yaguzmang, line 23: After further checking, actually these values might not be present in the odp data even though they can be calculated. e.g: forest total in local before this change:
Pasted Graphic 5
(empty "-" instead of calculate value)
when it should have been:
Pasted Graphic 4
So, better to leave everything working as before, and take note of this and look into it when we want to work on the odp itself.

minotogna (approved):

minotogna (changes requested):
U mean you found issues maybe better to start again ?

PR #4332, opened 2025-02-05
4305 - Data History: Data table (with ODP) comparison (Prev cycle)

minotogna:
https://github.com/user-attachments/assets/3e137d1b-20da-4cd6-9163-83f84accca59


src/client/pages/Section/DataTable/Table/GridHeadCell/hooks/useGridHeadCellProps.ts
```diff
@@ -7,10 +7,11 @@ import { Cols } from 'meta/assessment'
 
 import { useAssessmentCountry } from 'client/store/area'
 import { useCycle } from 'client/store/assessment'
-import { useOriginalDataPointYears } from 'client/store/data'
 import { useShowOriginalDatapoints } from 'client/store/ui/assessmentSection'
 import { getODPColSpan } from 'client/pages/Section/DataTable/Table/utils/getODPColSpan'
 
+import { useOriginalDataPointYearsWithHistory } from '../../hooks/useOriginalDataPointYearsWithHistory'
+import { ODPYear } from '../../types'
```
yaguzmang: should this import have the entire path as in other components? `'client/...`

src/client/pages/Section/DataTable/Table/GridHeadCell/OdpHeaderCell/OdpHeaderCell.tsx
```diff
@@ -54,20 +59,15 @@ const OdpHeaderCell: React.FC<Props> = (props) => {
     >
       <div>
         <Link
-          className="link"
+          className={classNames('table-grid__odp-link', { added: historyInfo?.added, removed: historyInfo?.removed })}
```
yaguzmang, line 64: Would it make sense to pass here class names `'diff-text added'` and `'diff-text removed'` so the colors are handled that way and we don't have the diff colors in different places?

src/client/pages/Section/DataTable/Table/types.ts
```diff
@@ -12,3 +12,13 @@ export type RowProps = {
   sectionName: string
   table: Table
 }
+
+export type ColHeader = {
+  columnName: string
+  odpYear?: ODPYear
+}
+
+export type ODPYear = {
```
sorja: Type name is misleading, ODPYearHeader?

minotogna: renamed to ODPColHeader .

sorja (approved):

PR #4337, opened 2025-02-06
4253 - Data History: ODP comparison (empty dataset) - Extent of forest: other land percent diffs

yaguzmang:
(UPDATED)

https://github.com/user-attachments/assets/1c666cd9-96cc-4950-b455-cc39c589a6b9

rlt:

https://github.com/user-attachments/assets/ccace318-051e-4c63-8754-a1d1290c5bc0

### DIFF
--- a/src/client/pages/OriginalDataPoint/OriginalDataPoint.scss
+++ b/src/client/pages/OriginalDataPoint/OriginalDataPoint.scss
@@ -194,3 +194,24 @@
     max-width: 150px !important;
   }
 }
+
+.odp-calculated-percent-diff,
+.odp-percent-diff {
+  align-items: center;
+  direction: ltr;
+  display: flex;
+  justify-content: end;
+}
+
+.odp-calculated-percent-diff {
+  > span {
+    margin-left: $spacing-xxs;
+  }
+}
+
+.odp-percent-diff,
+.odp-data-input-diff {
+  cursor: default;
+  gap: 10px;
+  padding: 10px 9px 10px 10px; // Matches PercentInput right spacing of 9px
+}
--- a/src/client/pages/OriginalDataPoint/components/ExtentOfForest/ExtentOfForestRow.tsx
+++ b/src/client/pages/OriginalDataPoint/components/ExtentOfForest/ExtentOfForestRow.tsx
@@ -12,6 +12,7 @@ import { TooltipId } from 'meta/tooltip'
 
 import { useCycle } from 'client/store/assessment'
 import { useHistoryLastApprovedIsActive } from 'client/store/data'
+import DiffText from 'client/components/DiffText'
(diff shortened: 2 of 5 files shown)


src/client/pages/OriginalDataPoint/components/ExtentOfForest/ExtentOfForestRow.tsx
```diff
@@ -176,8 +176,14 @@ const ExtentOfForestRow: React.FC<Props> = (props) => {
...
 
       <td className="fra-table__calculated-cell">
-        <span>{otherLand}</span>
-        <span style={{ marginLeft: '8px' }}>%</span>
+        {historyLastApprovedIsActive ? (
+          <DiffText changes={otherLandPercentChange} />
+        ) : (
+          <>
+            <span>{otherLandPercent}</span>
+            <span style={{ marginLeft: '8px' }}>%</span>
```
minotogna, line 190: Should we exclude the % sign from the diffs ?

yaguzmang: I think it makes sense to keep it since the col names don't have the percent unit symbol, so it's informative to have it. Especially when navigating the platform with the history active.
Also if users let's say for example, share a screenshot of the diffs with someone else that might not have the entire context. 
What do you think?

minotogna: What I meant is that the % is always visible in the ui . The diff should always be for the values .
The % remains where it is and it doesn't change color .
What do you think ?

yaguzmang: Ah I see, yes, sure 👍 will update here and the other instances. 
To note: this case is different than the PercentInput component because there the percent sign is handled internally. So, I will add an extra `span %` for those diffs

minotogna: The layout should not shift .
I'm sure you'll find the way 😉

src/client/pages/OriginalDataPoint/OriginalDataPoint.scss
```diff
@@ -194,3 +194,24 @@
...
+    margin-left: $spacing-xxs;
+  }
+}
+
+.odp-percent-diff,
+.odp-data-input-diff {
+  cursor: default;
+  gap: 10px;
+  padding: 10px 9px 10px 10px; // Matches PercentInput right spacing of 9px
+}
```
yaguzmang, line 217: Adding these classes at this level since they will be used for both odp -> extent of forest and odp -> forest characteristics

minotogna (approved):

PR #4366, opened 2025-02-12
4342 - Show diffs after table loaded

### LINKED ISSUE #4342: Data History: UI Diffs render logic
Total request time might be over 0.5ms - should not show history before data is loaded
Display loading indicator (maybe button disabled, button pulse, spinner, ...)

- [x] add hook useHistoryLastApprovedDataTableFetched
- [x] add hook useHistoryLastApprovedDescriptionFetched
- [x] add hook useHistoryLastApprovedODPFetched

UI ->  Do not render DiffText if data history for that information has not been loaded (probably if the relative data slice is undefined)

- [x] DataTable (cell)
- [x] Descriptions Text
- [x] Descriptions Data Source

sorja:
Resolves #4342

Example output:
```json
{"fra":{"2020":{"FIN":{"extentOfForest":{},"climaticDomain":{}}}}}
```


src/server/controller/cycleData/tableData/getTableDataLastApproved.ts
```diff
@@ -13,12 +13,32 @@ import { PropsGetTableData } from './props'
...
+  return data
+}
+
 export const getTableDataLastApproved = async (
   props: Props,
   client: BaseProtocol = DB
 ): Promise<RecordAssessmentData> => {
   const { assessment, countryISOs, cycle, info, tableNames } = props
-  const { prevCycle, lastAccepted } = info
+  const { prevCycle, lastAccepted } = info ?? {}
```
minotogna: Maybe better to fetch info internally here and then do the if not info return defaults .

What do you think ?

src/client/store/data/hooks/useLastApprovedHistoryTableData.ts
```diff
@@ -32,10 +33,13 @@ export const useLastApprovedHistoryNodeValue = (props: PropsNodeValue): NodeValu
   }, [assessmentName, colName, countryIso, cycleName, data, tableName, variableName])
 }
 
-export const useHistoryLastApprovedDataTableFetched = (): boolean => {
+export const useHistoryLastApprovedDataTableFetched = (tableName: TableName): boolean => {
```
minotogna, line 36: Can we also do similarly for description and add the props to the odp hook ?

sorja: >  add the props to the odp hook ?

what do you mean, year?

minotogna: The year for odp is the identifier . And to check that the odp history data is loaded , we need the year as parameter of the hook (relative to odp history data is fetched )

minotogna (approved):

PR #4369, opened 2025-02-12
4364 table 1a forest area   previously reported   validator   latest   2025 [RFC]

sorja:
- [ ] migration step
- [ ] add unit test for type `fra['2025'].extentOfForest.forestArea['2025']` format formula

### DIFF
--- a/src/server/controller/assessment/generateMetaCache/dependencyEvaluator/evalDependencies/context.ts
+++ b/src/server/controller/assessment/generateMetaCache/dependencyEvaluator/evalDependencies/context.ts
@@ -1,10 +1,12 @@
 import { ExpressionContext } from '@openforis/arena-core'
 
-import { Assessment, AssessmentName, CycleName, RowCache } from 'meta/assessment'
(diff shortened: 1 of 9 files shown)


src/server/controller/assessment/generateMetaCache/dependencyEvaluator/evalDependencies/context.ts
```diff
@@ -1,10 +1,12 @@
 import { ExpressionContext } from '@openforis/arena-core'
 
-import { Assessment, AssessmentName, CycleName, RowCache } from 'meta/assessment'
+import { Assessment, AssessmentName, Cycle, CycleName, RowCache } from 'meta/assessment'
 
 export interface Context extends ExpressionContext {
   assessments: Array<Assessment>
+  assessment: Assessment
   assessmentName: AssessmentName
+  cycle: Cycle
```
minotogna, line 9: can u get assessment from assessments.find[assessmentName]? and cycle as well from cycleName? The idea is to pass all assessments. it seems a bit of a duplication  .
but assessments can be passed as RecordAssessments . what do u think?

PR #4379, opened 2025-02-17
4372 - Assessment Switch Header: Admin - enable on Global/Regions

### LINKED ISSUE #4372: Assessment Switch Header: Admin - enable on Global/Regions
the header should always allow to switch assessment cycles.

do not alter current behaviour when accessing countries, only add the following case when on **regions for admin**:
- if same assessment navigate to region/global  
- if different assessment go to cycle home

sorja:
https://github.com/user-attachments/assets/b028826c-2c52-490b-bc62-9b98e61fa91a

### DIFF
--- a/src/client/components/PageLayout/Header/CycleSwitcher/hooks/useNavigateTo.tsx
+++ b/src/client/components/PageLayout/Header/CycleSwitcher/hooks/useNavigateTo.tsx
@@ -22,13 +22,12 @@ export const useNavigateTo = () => {
       const cycleName = cycle.name
       const hasRoleInCountry = Users.hasRoleInCountry({ user, cycle, countryIso })
 
-      let link = ''
+      let link = Routes.Cycle.generatePath({ assessmentName, cycleName })
       if (Areas.isISOCountry(countryIso) && (Cycles.isPublished(cycle) || hasRoleInCountry))
         link = Routes.CountryHome.generatePath({ countryIso, assessmentName, cycleName })
       if (!Areas.isISOCountry(countryIso) && (Cycles.isPublished(cycle) || hasRoleInCountry))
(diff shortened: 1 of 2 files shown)


src/client/components/PageLayout/Header/CycleSwitcher/hooks/useNavigateTo.tsx
```diff
@@ -22,8 +23,10 @@ export const useNavigateTo = () => {
       const hasRoleInCountry = Users.hasRoleInCountry({ user, cycle, countryIso })
 
       let link = ''
-      if (countryIso && (Cycles.isPublished(cycle) || hasRoleInCountry))
+      if (Areas.isISOCountry(countryIso) && (Cycles.isPublished(cycle) || hasRoleInCountry))
         link = Routes.CountryHome.generatePath({ countryIso, assessmentName, cycleName })
+      if (!Areas.isISOCountry(countryIso) && (Cycles.isPublished(cycle) || hasRoleInCountry))
+        link = Routes.CountryHome.generatePath({ countryIso: Global.WO, assessmentName, cycleName })
```
minotogna: if we are in a region, e.g. EU, do we navigate to EU ? does WO exists in pan-european ?

sorja: I must have misread this
> if same assessment navigate to region/global

I will update so that the region doesn't change if a region is selected (default , previous behavior)

sorja: @minotogna 
I have updated the logic and aded a video to description navigation inside the same assessment and to another assessment

It should be as expected now

👍

minotogna (approved):

PR #4381, opened 2025-02-18
4307 - Data History: Descriptions html diffs - Initial DiffDOM

yaguzmang:
Initial DiffDOM

(UPDATED)

https://github.com/user-attachments/assets/bcb32b96-d5d0-4a4e-b4e3-2008675c990a

Coming PRs:
- [ ] Find a way to add back non breaking spaces so the layout stays similar
- [ ] removeElement
- [ ] removeTextElement
- [ ] modifyTextElement

For links:
- [ ] addAttribute
- [ ] modifyAttribute
- [ ] removeAttribute

--
- [ ] Investigate why this diff is showing like this:

image

image
(most likely because diff text isn't implemented yet)


src/client/components/DiffDOM/hooks/useDOMChanges.ts
```diff
@@ -0,0 +1,66 @@
...
+export const useDOMChanges = (props: Props) => {
+  const { current, prev, ref } = props
+
+  useLayoutEffect(() => {
+    if (!ref.current) return
+    const diffDOM = new DiffDOM({
+      caseSensitive: true,
+      preDiffApply(info): boolean {
+        if (info.diff.action === 'replaceElement') {
+          const div = document.createElement('div')
```
minotogna: should we create two inner function replaceElement and addElememt (exported from their own file) to keep this file reasonable ?

src/client/pages/Section/Descriptions/CommentableDescription/DescriptionDiffView/DescriptionDiffView.tsx
```diff
@@ -18,9 +19,14 @@ const DescriptionDiffView: React.FC<Props> = (props) => {
...
   const descriptionB = useCommentableDescriptionValue({ name, sectionName })?.text || ''
 
-  const changes = useChanges({ descriptionA, descriptionB })
+  const changes = useChanges({ descriptionA, descriptionB, name })
 
-  return <DiffText changes={changes} />
+  return (
+    <>
+      {name !== CommentableDescriptionName.generalComments && <DiffText changes={changes} />}
+      {name === CommentableDescriptionName.generalComments && <DiffDOM current={descriptionB} prev={descriptionA} />}
```
minotogna: should we use DiffDom directly for all already now? in my poc I was only testing with generalComments

src/client/pages/Section/Descriptions/CommentableDescription/hooks/useChanges.ts
```diff
@@ -3,19 +3,24 @@ import { useMemo } from 'react'
...
+  name: CommentableDescriptionName
 }
 
 type Returned = Array<Change>
 
 // TODO: Rename to useChangesHTML
 // TODO: Add support for JSON/DataSource type
 const useChanges = (props: Props): Returned => {
-  const { descriptionA, descriptionB } = props
+  const { descriptionA, descriptionB, name } = props
```
minotogna: and remove this

sorja:
Looks good, can you provide a video with more complete / complex test set / demo ?

E.g. adding and removing whole tables, editing values inside of a table, adding or removing cols?

Adding / mixing whole elements which are supported through Jodit (e.g. are images supported?)

Just curious to see how it looks

minotogna:
> Looks good, can you provide a video with more complete / complex test set / demo ?
> 
> E.g. adding and removing whole tables, editing values inside of a table, adding or removing cols?
> 
> Adding / mixing whole elements which are supported through Jodit (e.g. are images supported?)
> 
> Just curious to see how it looks

we don 't support images @sorja !

minotogna:
> > Looks good, can you provide a video with more complete / complex test set / demo ?
> > E.g. adding and removing whole tables, editing values inside of a table, adding or removing cols?
> > Adding / mixing whole elements which are supported through Jodit (e.g. are images supported?)
> > Just curious to see how it looks
> 
> we don 't support images @sorja !

@sorja but 2020 has images. damn... let's just proceed with the development . there's still some work to do/
final video and testing has to be done by all of us. this diff is complex

src/client/components/DiffDOM/DiffDOM.tsx
```diff
@@ -0,0 +1,31 @@
...
+    prev,
+    ref,
+  })
+
+  return (
+    <div
+      ref={ref}
+      className={classNames('editorWYSIWYG jodit-wysiwyg', className)}
+      dangerouslySetInnerHTML={{
+        __html: cleanDOM(prev ?? ''),
```
minotogna: better to create a variable with useMemo

src/client/components/DiffDOM/DiffDOM.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useRef } from 'react'
+
+import classNames from 'classnames'
+
+import { useDOMChanges } from './hooks/useDOMChanges'
+import { cleanDOM } from './hooks/utils'
+import { DiffDOMProps } from './types'
+
+const DiffDOM: React.FC<DiffDOMProps> = (props) => {
+  const { className = '', current, prev } = props
```
minotogna: do we ever pass a className from outside now ? 
If not, I always suggest to not add things we don't use

src/client/components/DiffDOM/types.ts
```diff
@@ -2,3 +2,11 @@ export type DiffDOMProps = {
   current: string
   prev: string
 }
+
+export type DiffInfo = {
```
yaguzmang: The library doesn't export a type for this, and `@types/diff-dom` doesn't seem to be available. So, adding this generic here to avoid ts warnings.

src/client/components/DiffDOM/hooks/_replaceElement.ts
```diff
@@ -0,0 +1,13 @@
...
+export const replaceElement = (info: DiffInfo) => {
+  const div = document.createElement('div')
+  div.className = 'diff-text'
+
+  const divRemoved = document.createElement('div')
+  divRemoved.className = 'removed'
+  divRemoved.appendChild(info.node.cloneNode(true))
+
+  div.appendChild(divRemoved)
+  info.node.replaceWith(div)
```
minotogna: We need to add the added node as well as green .

Also , instead of removing nodes, can we use the same way u did for the other diff type ?

src/client/components/DiffDOM/types.ts
```diff
@@ -0,0 +1,12 @@
+export type DiffDOMProps = {
+  current: string
+  prev: string
+}
+
+export type DiffInfo = {
+  diff: {
+    action: 'replaceElement' | 'addElement'
```
minotogna: Can we add an enum ?

src/client/components/DiffDOM/types.ts
```diff
@@ -0,0 +1,12 @@
...
+    [key: string]: any
+  }
+  [key: string]: any
```
minotogna: Instead of any , can we use unknown or never ?

src/client/pages/Section/Descriptions/CommentableDescription/DescriptionDiffView/DescriptionDiffView.tsx
```diff
@@ -15,12 +14,10 @@ const DescriptionDiffView: React.FC<Props> = (props) => {
 
   const descriptions = useLastApprovedHistoryDescriptions()
   // TODO: Support data sources after 2020
```
minotogna: Do we need this TODO ?

src/client/components/DiffDOM/hooks/utils.ts
```diff
@@ -0,0 +1,37 @@
+import { DiffElement, DiffElementNode, DiffTextNodeName, DiffType } from 'client/components/DiffDOM/types'
+
+export const cleanDOM = (value: string): string => {
+  return value
+    .replace(/[\n\t\r]/g, '')
+    .replace(/\u00a0|&nbsp;/g, ' ')
```
minotogna: Removing space was giving me problems . Why was this enabled ?

src/client/components/DiffDOM/hooks/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  const elementNode = element as DiffElementNode
+  const newClass = elementNode.attributes?.class ? `${elementNode.attributes.class} ${type}` : type
+
+  return {
+    ...elementNode,
+    attributes: {
+      ...elementNode.attributes,
+      class: newClass,
+    },
+    childNodes: elementNode.childNodes?.map((child) => applyDiffClassToElement(child, type)) || [],
```
minotogna, line 40: Recursion should be avoided . Can u make an iterative version ?

src/client/components/DiffDOM/types.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export type DiffElement = DiffElementNode | DiffTextNode
+
+export enum DiffInfoAction {
+  addElement = 'addElement',
+  replaceElement = 'replaceElement',
+}
+
+export type Diff =
```
minotogna, line 37: We could have different types per action that extend base Diff .
What do u think ?

src/client/components/DiffDOM/hooks/utils.ts
minotogna: Wait @yaguzmang . Sorry this should be fine . Ignore this comment please

src/client/components/DiffDOM/hooks/utils.ts
yaguzmang: Maybe it wasn't an issue using DOM nodes, but when using the diff Element this happens:
image
Seems those non breaking spaces aren't handled correctly by the library.
I will check later if we can keep those using `postDiffApply` to parse them maybe

src/client/components/DiffDOM/types.ts
yaguzmang: Something like this?
```ts
interface BaseDiff {
  route: Array<number>;
}


export interface AddElementDiff extends BaseDiff {
  action: DiffInfoAction.addElement;
  element: DiffElement;
}

export interface ReplaceElementDiff extends BaseDiff {
  action: DiffInfoAction.replaceElement;
  newValue: DiffElement;
  oldValue: DiffElement;
}


export type Diff = AddElementDiff | ReplaceElementDiff;
```

minotogna (approved):

PR #4382, opened 2025-02-18
4373 - Global Countries Filter

### LINKED ISSUE #4373: Global Countries filter
Image

- [ ] Hide Filter countries when there is no dashboard data
- [ ] Update the UI (and behaviour of course) as per admin collaborators page and it must look exactly as the image below:

Image

(https://fra-data.review.fao.org/assessments/fra/2020/admin/collaborators)

sorja:
[WO] Country/Region home:

image

image

image

Admin still working as before:

image

image


src/client/components/CountryMultiSelect/CountryMultiSelect.scss
```diff
@@ -1,7 +1,11 @@
 @import 'src/client/style/partials';
 
-div.filter-country {
+.filter-multiselect__container {
```
minotogna: Should we remove the concept of filter from country multi select ?

src/client/components/CountryMultiSelect/hooks/useTooltipContent.tsx
```diff
@@ -4,35 +4,49 @@ import { useTranslation } from 'react-i18next'
...
-import { useCountriesByRegionOptions } from 'client/components/CountryMultiSelect/hooks/useCountriesByRegionOptions'
 import { OptionsGroup } from 'client/components/Inputs/Select'
 
-type Props = {
-  fieldName: string
-  path: string
+import { useCountriesByRegionOptions } from './useCountriesByRegionOptions'
+
+interface Props {
+  selection: Array<CountryIso>
```
minotogna: I would call this value . To be consistent with select

src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
```diff
@@ -1,27 +1,68 @@
...
-  const { multiLabelSummaryKey, placeholder } = props
+const CountryMultiSelect: React.FC<Props> = (props) => {
+  const { value, onChange, placeholder, error = false, minCountries, onMenuClose } = props
 
   const { t } = useTranslation()
-
   const optionGroups = useCountriesByRegionOptions()
 
+  const errorMessage = minCountries
+    ? t('statisticalFactsheets.validation.selectAtLeastNCountries', { count: minCountries })
```
minotogna: The error should be passed from outside .
The country multi select has nothing to do with dashboard (the i18n string still read statistical fact sheet )

src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
```diff
@@ -1,27 +1,68 @@
...
+    : undefined
+
+  const { hideTooltip, showTooltip, tooltipContent, dataTooltipId } = useTooltipContent({
+    selection: (value as Array<CountryIso>) ?? [],
+    error,
+    errorMessage,
```
minotogna: Can we pass only one optional prop 'error' type string ?
If error is passed then there is an error

src/client/components/Hr/Hr.scss
```diff
@@ -3,4 +3,9 @@
 .hr {
   background: radial-gradient(#c1c1c1, white);
   height: 1px;
+
+  &.vertical {
```
minotogna, line 7: Can we get the same style than admin vertical separator ? Maybe using there this Hr ?

src/client/components/TablePaginated/Filters/MultiSelect/MultiSelect.scss
```diff
@@ -55,4 +55,8 @@ div.filter-multiselect__container {
   &.active {
     border: 1px solid $ui-accent;
   }
+
+  &.error {
```
minotogna: do we use errors in table paginated filters ?

sorja: I need to investigate this - I realised that if we don't add this , then the error border doesn't show...
I must see why is that so

minotogna: sure

src/client/components/Hr/Hr.scss
sorja: updated as in pics:

kuva

kuva

src/client/components/TablePaginated/TablePaginated.scss
```diff
@@ -58,9 +58,3 @@
     }
   }
 }
-
-.table-paginated-actions-sep {
-  background: radial-gradient($ui-accent-dark, $ui-bg);
```
minotogna, line 63: Can we still have the gradient for the hr ?

src/client/pages/CountryHome/CountryHeader/CountrySelector/CountrySelector.scss
```diff
@@ -0,0 +1,18 @@
...
+.country-selector {
+  align-items: center;
+  display: flex;
+  gap: $spacing-xxs;
+  height: 30px;
+  width: 100%;
+
+  .hr.vertical {
+    margin-left: $spacing-xxs;
+    margin-right: $spacing-xxs;
```
minotogna: What about a flex container with gap (to avoid these margins ?)

src/client/components/Hr/Hr.tsx
```diff
@@ -5,15 +5,17 @@ import classNames from 'classnames'
...
 }
 
 const Hr: React.FC<Props> = (props: Props) => {
-  const { className } = props
-  return <div className={classNames('hr', className)} />
+  const { className, vertical } = props
+  return <div className={classNames('hr', { vertical }, className)} />
 }
 
 Hr.defaultProps = {
```
minotogna: Can we remove default props ?
I thinks when we face components with default props , we should remove them .
What do u think ?

src/client/components/TablePaginated/TablePaginated.scss
sorja: kuva

gradient is set in the Hr.scss, would you prefer the gradient we had before?

sorja: Added the previous gradient

src/client/pages/CountryHome/CountryHeader/CountrySelector/CountrySelector.scss
sorja: kuva

Updated - looks like above now

minotogna (approved):

PR #4386, opened 2025-02-20
4307 - Data History: DiffDOM - Modify text element

yaguzmang:
Adding diffs with `textDiff`, which  is called internally on the modifyTextElement case ([as seen here](https://github.com/fiduswriter/diffDOM/blob/main/src/diffDOM/dom/apply.ts#L83))

(UPDATED)

https://github.com/user-attachments/assets/ab014ccc-3070-4da2-9a62-1ab45cc67002

### DIFF
--- a/src/client/components/DiffDOM/hooks/_getTextDiffNode.ts
+++ b/src/client/components/DiffDOM/hooks/_getTextDiffNode.ts
@@ -1,11 +1,11 @@
 import * as Diff from 'diff'
 
 export const getTextDiffNode = (currentValue: string, newValue: string): HTMLDivElement => {
-  const changes = Diff.diffChars(currentValue.replaceAll('&nbsp;', '\u00A0'), newValue.replaceAll('&nbsp;', '\u00A0'))
+  const changes = Diff.diffLines(currentValue.replaceAll('&nbsp;', '\u00A0'), newValue.replaceAll('&nbsp;', '\u00A0'))
 
   const newNode = document.createElement('div')
 
-  changes.forEach((change) => {
+  changes.forEach((change, i) => {
     const { added, removed, value } = change
 
     value.split('\n\r').forEach((text) => {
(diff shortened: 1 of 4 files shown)


src/client/components/DiffDOM/hooks/_getTextDiffNode.ts
```diff
@@ -0,0 +1,21 @@
+import * as Diff from 'diff'
+
+export const getTextDiffNode = (currentValue: string, newValue: string): HTMLDivElement => {
+  const changes = Diff.diffChars(currentValue.replaceAll('&nbsp;', '\u00A0'), newValue.replaceAll('&nbsp;', '\u00A0'))
```
yaguzmang: To note: with diffChars, some diffs are hard to understand, especially in some cases with numbers:
image

minotogna: Let's use diff lines ? Have you tried ?

yaguzmang: Yes, this:
image

Becomes this:
image

Harder to see small changes in paragraphs but works better for more cases.
Will add it now 👍

src/client/components/DiffDOM/hooks/_getTextDiffNode.ts
```diff
@@ -0,0 +1,25 @@
+import * as Diff from 'diff'
+
+export const getTextDiffNode = (currentValue: string, newValue: string): HTMLDivElement => {
+  const changes = Diff.diffLines(currentValue.replaceAll('&nbsp;', '\u00A0'), newValue.replaceAll('&nbsp;', '\u00A0'))
```
minotogna, line 4: Should we have a utility Strings.nbspToAscii ?

minotogna: Even next pr

yaguzmang: Sure, will add in the next one 👍

minotogna (approved):

PR #4397, opened 2025-02-24
4396 - Application Crash when selecting country when not logged in

### LINKED ISSUE #4396: Application Crash when selecting country when not logged in
missing null check for user.roles

sorja:


minotogna (approved):
👍

PR #4409, opened 2025-02-27
4408 - PDF and Print: Show dash for empty comment sections

### LINKED ISSUE #4408: PDF and Print: Show dash for empty comment sections
Don't create virtual DOM

sorja:
Returns original functionality overlooked in previous PR.

- Update to checking empty HTML:
Don't create virtual DOM
Just use simple regex pattern to match for empty

- Remove prop drilling for showDashEmptyContent (only used in print)

image

### DIFF
--- a/src/client/pages/Section/Descriptions/CommentableDescription/CommentableDescription.tsx
+++ b/src/client/pages/Section/Descriptions/CommentableDescription/CommentableDescription.tsx
@@ -27,7 +27,7 @@ type Props = {
 }
 
 const CommentableDescription: React.FC<Props> = (props) => {
-  const { name, repository, template, title } = props
+  const { name, repository, template = { text: '' }, title } = props
   const { print } = useIsPrintRoute()
   const { sectionName } = useSectionContext()
   const value = useCommentableDescriptionValue({ name, sectionName, template })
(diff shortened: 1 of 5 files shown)


src/client/pages/Section/Descriptions/CommentableDescription/hooks/useDescriptionErrorState.ts
```diff
@@ -17,16 +26,15 @@ type Returned = {
   empty: boolean
 }
 
+/*
+    Note: This is used only in print view
```
minotogna: what is used only in print view ? this hook ? isn't it always used ?

src/client/pages/Section/Descriptions/CommentableDescription/CommentableDescription.tsx
```diff
@@ -72,7 +72,6 @@ const CommentableDescription: React.FC<Props> = (props) => {
 
 CommentableDescription.defaultProps = {
```
minotogna: can we remove default props here ?

src/client/pages/Section/Descriptions/CommentableDescription/hooks/useDescriptionErrorState.ts
sorja: The value of the hook is disregarded if it's not a print view:

It's possible to remove the hook altogether when introducing the print switch for EditorWYSIWYG
image

minotogna: from the comment I understand that this hook `useDescriptionErrorState` is used only in print view.
If it's not the case, can you update the comment please ?

minotogna (approved):

PR #4410, opened 2025-02-27
4266 - PDF - page space

### LINKED ISSUE #4266: PDF - page space
@minotogna 

**Describe the bug**
In the PDF, some pages have a lot of empty space that could contain all the content of the next page(In some cases, the entire page is blank).

**To Reproduce**
Examples:
Zambia PDF:
2d carbon stock; 7 Non wood forest products removals and value 2020

Yemen PDF:
Table 1a NDP

**Expected behavior**
Wherever possible, keep the necessary space between sub-components of a section, avoiding excessive empty space in the middle.

**Screenshots**
Zimbia:
Image

Image

Yemen:

Image

sorja:


src/client/pages/Section/Descriptions/Descriptions.tsx
```diff
@@ -16,22 +15,18 @@ type Props = {
...
   }
 
   return (
-    <>
-      <div className="descriptions">
-        {nationalData && <NationalDataDescriptions nationalData={nationalData} />}
-
-        {analysisAndProcessing && <AnalysisDescriptions analysisAndProcessing={analysisAndProcessing} />}
-      </div>
-      {print && !onlyTables && <div className="page-break" />}
```
minotogna, line 33: Each description should be rendered in its own page . Is it the case ?

sorja: The main problem as seen in the issue that sometimes when we have empty descriptions or different length descriptions we might get only couple of symbols (e.g. on empty we get a dash) on a single page
https://github.com/openforis/fra-platform/issues/4266


One thing I was thinking of we can try to group the subheader with the content if it's possible (I will investigate)

Current PR is proposed fix for the issue, if page-break for descriptions is preferred, let me know

minotogna: That was the best compromise we found in past assessment to let the content fit .
We should not alter this behavior, otherwise we'll create other problems

sorja: Sure, should I look into grouping the title with the content to avoid empty pages containig only `-` ?
Or we can close this PR also

minotogna: Sorry @sorja I think the issue is not about dash or descriptions in single page .
I saw some screenshots in the issue that are not about descriptions . Am I mistaken ?

sorja: Yes

sorja: Which screenshot is not about descriptions, sorry? It seems all of them are?

sorja: For example first screenshot:

image

-> because of the page break we have


Also the next one:

image

-> because of the page break we have



--> The last one is because of the chart page-break I will double check this


But I think we can close this PR and issue if you like, as the behavioer is expected?

minotogna: The last one is not about descriptions. Can we at least fix that ?

minotogna: Also , if there's an empty page after a description , can it be fixed as well ?

sorja: Sure, should we roll back the change regarding descriptions?
I will remove the page break after the chart

sorja: Sure

> Also , if there's an empty page after a description , can it be fixed as well ?

I will investigate the issue more in depth and let you know
I think grouping the subheader with the content will resolve the issue

minotogna: What do you mean grouping ?
The layout should not change 
See the 2020 report on the fra website

minotogna: Maybe you want to work on another issue ?

PR #4415, opened 2025-03-04
4283 - PDF hide empty graph

### LINKED ISSUE #4283: PDF - graph
@minotogna 

**Describe the bug**
The graph is entirely empty when no data is available. As Örjan suggests, the graph should be removed if it is entirely empty.

Could you please quickly filter out countries that do not have **any of** these data for the main classes (Table 1a: Forest; Other wooded land; Table 1b: Naturally regenerating forest; Plantation forest; Other planted forest) to ensure they are not missed?

**To Reproduce**
Example

Albania PDF - Table 1b graph

**Screenshots**

Image

sorja:


src/client/pages/Section/DataTable/DataTable.tsx
```diff
@@ -38,7 +38,26 @@ const DataTable: React.FC<Props> = (props) => {
...
+      )
+    }
+
+    if (tableName === TableNames.forestCharacteristics) {
+      return ['plantedForest', 'naturalForestArea'].every((variableName) =>
+        RecordAssessmentDatas.isVariableDataEmpty({ ...baseProps, variableName })
+      )
+    }
+    return false
+  }, [baseProps, tableName])
```
minotogna: Can we have a hook for this ? More clean maybe

src/client/pages/Section/DataTable/DataTable.tsx
```diff
@@ -47,9 +66,15 @@ const DataTable: React.FC<Props> = (props) => {
     return null
   }
 
+  // Show chart in print only if chart variables are not empty
+  const showChartPrint = showOdpChart && print && !areChartVariablesEmpty
+
+  // By default, show chart always
+  const showChartDefault = showOdpChart && !print
```
minotogna: Can we have only one show chart ?

src/meta/data/recordAssessmentDatas/isVariableDataEmpty.ts
```diff
@@ -0,0 +1,26 @@
+import { Objects } from 'utils/objects'
+
+import { CountryIso } from 'meta/area'
+import { AssessmentName, CycleName, TableName, VariableName } from 'meta/assessment'
+
+import { RecordAssessmentData } from '../RecordAssessmentData'
+import { getTableData } from './getTableData'
+
+export const isVariableDataEmpty = (props: {
```
minotogna: Mate a type Props ?

src/test/migrations/steps/20250228094002-step-clean-up-descriptions-empty.ts
```diff
@@ -1,47 +1,70 @@
...
 
-import { Assessment, AssessmentNames } from 'meta/assessment'
+import { Assessment, AssessmentNames, Cycle } from 'meta/assessment'
 
 import { BaseProtocol, Schemas } from 'server/db'
 import { AssessmentRepository } from 'server/repository/assessment/assessment'
 
+const brTagRegex = '<br\\s*/?>'
+
 export default async (client: BaseProtocol) => {
```
minotogna: This has already been executed . Another migration is needed

src/client/pages/Section/DataTable/hooks/useAreChartVariablesEmpty.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return useMemo(() => {
+    const { tableName } = props
+    if (tableName === TableNames.extentOfForest) {
+      return ['forestArea', 'otherWoodedLand'].every((variableName) =>
+        RecordAssessmentDatas.isVariableDataEmpty({ ...props, variableName })
+      )
+    }
+
+    if (tableName === TableNames.forestCharacteristics) {
+      return ['plantedForest', 'naturalForestArea'].every((variableName) =>
```
minotogna: In 1b we have 3 variables if I recall correctly. Can u please check ?

src/meta/data/recordAssessmentDatas/isVariableDataEmpty.ts
```diff
@@ -0,0 +1,28 @@
...
+export const isVariableDataEmpty = (props: Props): boolean => {
+  const { assessmentName, cycleName, data, tableName, countryIso, variableName } = props
+  const tableData = getTableData({ assessmentName, cycleName, data, tableName, countryIso })
+
+  if (Objects.isEmpty(tableData)) {
+    return true
+  }
+
+  const recordRowData = Object.values(tableData)
+  return Objects.isEmpty(recordRowData.filter((rowData) => rowData[variableName]))
```
minotogna: Should we check the raw value ?

Maybe an every as well here ?

minotogna (commented):
👍minor comments

src/client/pages/Section/DataTable/hooks/useAreChartVariablesEmpty.tsx
sorja: I will check the code also, but in UI we can only see 2 I think
https://fra-data.review.fao.org/assessments/fra/latest/FIN/sections/forestCharacteristics

minotogna: yes, I remember that from 2020.
However we have useTrends to get dynamically the variables displayed in the chart.
maybe here we can use the same dynamic approach instead of hardcoding the variables ?
image

src/meta/data/recordAssessmentDatas/isVariableDataEmpty.ts
sorja: **I added every - but please note - if we are missing a single value (e.g. forest 2000) the chart will be hidden**

Is this the behavior we want? Should be check _some_ value exists? What if one value only exists?

I think that normally, all values must exist because 1a and 1b are the main tables?

minotogna: Yes , every is empty 
some is not empty . 
Maybe you're right , let's use some is not empty . The loop will ends before

sorja: Ah, thank you for the clarification, I initally understood that if _every_ has value instead it should have been if _every_ is empty ;)

Added _some_ has not value

src/meta/data/recordAssessmentDatas/isVariableDataEmpty.ts
```diff
@@ -0,0 +1,28 @@
...
+export const isVariableDataEmpty = (props: Props): boolean => {
+  const { assessmentName, cycleName, data, tableName, countryIso, variableName } = props
+  const tableData = getTableData({ assessmentName, cycleName, data, tableName, countryIso })
+
+  if (Objects.isEmpty(tableData)) {
+    return true
+  }
+
+  const rowData = Object.values(tableData)
+  return rowData.some((rowData) => !Objects.isEmpty(rowData[variableName]))
```
minotogna: Does this check value.raw ? The node value can contain metadata but the value can be empty

minotogna (approved):

PR #4429, opened 2025-03-07
4428 - EditorWYSIWYGReadOnly adjust styling

### LINKED ISSUE #4428: PDF: Editor WYSIWYG remove padding
In the Print View (screen and pdf) there should be no padding surrounding the html content 

Image

sorja:


src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
```diff
@@ -49,6 +50,12 @@ div.editorWYSIWYG {
       border: unset;
     }
   }
+
+  &.read-only-html {
+    @extend .textarea;
```
minotogna: Is this needed ? Does it make a difference with the editor in non read only mode ?
Wouldn't be better to just remove padding in print view ?

sorja: image


If we remove the basic styling, it looks off: (without this)
image

It is emulating a text area element, so when we add the `@extend` it looks correct:
image


what do you suggest?

sorja: Note we are using this element in different places, not only for the comments (e.g. as seen in screenshots we might have it inline in tables)

minotogna: I suggest to have consistent style in read only and non read only mode .
Same css classes and remove padding in print view .
Is it doable ?

src/client/pages/Print/OriginalDataPointsPrint/OriginalDataPointsPrint.tsx
```diff
@@ -59,7 +59,7 @@ const OriginalDataPointsPrint: React.FC<Props> = (props) => {
       {hasDescriptions && (
         <div className="odp__section-print-mode print-break-inside-avoid">
           <h3 className="subhead">{i18n.t('dataSource.comments')}</h3>
-          <DataGrid className="odp__section" gridTemplateColumns="100px minmax(240px, 40%)">
+          <DataGrid className="odp__section" gridTemplateColumns="100px 1fr">
```
sorja, line 62: full width odp comments

src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
sorja: You were right, I needed to add the relevant elements to make it work

minotogna (approved):
💪👍

Issue #4273, opened 2025-01-27
PDF - large table


pengchenglai:
@minotogna 

**Describe the bug**
Some data gets lost when there is a large table.
**To Reproduce**
Example

1. Cyprus PDF:
Go to table 2a

2. Denmark / Syrian Arab Republic / North Macedonia /Haiti
Table 1a & 1b, NDP

**Screenshots**
Cyprus:
![Image](https://github.com/user-attachments/assets/1725a5ab-02c6-47b1-bf06-af7db11f6bb8)

![Image](https://github.com/user-attachments/assets/a8d14d47-0109-4a80-b2e5-64dde9626062)

Denmark:

![Image](https://github.com/user-attachments/assets/ee0823c5-22a7-4496-8382-3c31e2e915cb)

pengchenglai:
@minotogna 

Do you have any suggestions? One possible solution proposed is to keep only the FRA years(such as 1990, 2000 ......) for the NDP issue (Table 1a &/ Table 1b).

pengchenglai:
In cases such as in Cyprus, where a user has entered a large table in section 2a **Estimation and Forecasting**, one possible solution could be to _adjust the column width_ for the years (2010, 2011, 2015, 2020, 2021, 2022, 2023, 2024, 2025) and _align the numbers to the left or right_, or make the table font smaller.

![Image](https://github.com/user-attachments/assets/47d1ce00-76d7-4d85-a2a9-3fcd393b9120)

minotogna:
@pengchenglai we can control only our tables, not free text content as it is not structured.
this problems will always happen if we let users enter whatever they want and be included in the print.
This kind of information should be entered as attached files and be included as annex. But this is not happening atm .

We can analyze and try to fix the cases (you have mentioned in this issue) by altering the user entered content in order to adjust the data. But this is not a solution, but we can try, if you want

pengchenglai:
@minotogna 

Yes, we need to separate this issue into two circumstances:

1. Structured tables with too many columns. The Admin has kindly agreed to keep the FRA years for the simplicity of life, as it preserves the trend and the most important information.
2. Plz give it a shot. No, it is not the best solution, but we could set up some limitations for the upcoming assessments when users enter the tables (e.g., a maximum of 10 columns; otherwise, users will have to upload the table as an annex and attach the link).

minotogna:
> [@minotogna](https://github.com/minotogna)
> 
> Yes, we need to separate this issue into two circumstances:
[...]

PR #4434, opened 2025-03-10
4273 - Fix overflowing description tables

sorja:
Part of https://github.com/openforis/fra-platform/issues/4273

### DIFF
--- a/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
+++ b/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
@@ -54,5 +54,9 @@ div.editorWYSIWYG {
 @include print {
   div.editorWYSIWYG {
     height: unset;
+
+    table {
+      width: 100%;
+    }
   }
 }
--- a/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
+++ b/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
@@ -56,7 +56,7 @@ div.editorWYSIWYG {
     height: unset;
 
     table {
-      width: 100%;
+      max-width: 100%;
     }
   }
 }
--- a/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
+++ b/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
@@ -56,7 +56,7 @@ div.editorWYSIWYG {
     height: unset;
 
     table {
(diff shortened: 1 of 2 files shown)


src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
```diff
@@ -54,5 +54,9 @@ div.editorWYSIWYG {
 @include print {
   div.editorWYSIWYG {
     height: unset;
+
+    table {
+      width: 100%;
```
minotogna: This might change user table if the table is less than 100%?

sorja: Which user table do you refer to which is inside of the editorWYSIWYG?

minotogna: Does max-width fix the issue as well ?

sorja: Sure, some tables are not fitted then though (they are the original size if they are less than 100%)
Ill change

minotogna: > Sure, some tables are not fitted then though (they are the original size if they are less than 100%) Ill change

sure. wouldn't that be better ?

src/client/pages/Section/Descriptions/Descriptions.scss
```diff
@@ -4,6 +4,9 @@
   display: grid;
   grid-row-gap: $spacing-s;
   margin-bottom: $spacing-s; // TODO: refactor section style with grid and row gap - future task
+  max-width: 100%;
+  min-width: 0;
+  overflow-x: auto;
```
minotogna: is this `overflow-x: auto` needed ?

src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
sorja: Sure, CYP had some weird tables but I guess it's PEBKAC in that case ;)

sorja: I was mistaken actually, further testing shows that this actually needs to be `width: 100%` or `width: auto` to avoid overflowing

Committing as `width: auto` after further testing

minotogna (approved):

PR #4453, opened 2025-03-14
4449 - Pasting links: Add toaster message on modified paste content

### LINKED ISSUE #4449: Pasting links: Add toaster message on modified paste content
Error logic:

If visual elements are modified

Error message:

Your content has been updated, because it contains invalid HTML.
Only text and links can be pasted in this field.

sorja:
https://github.com/user-attachments/assets/cffb37b1-24c0-4119-bbe7-18e3afaaed06

### DIFF
--- a/src/client/components/EditorWYSIWYG/hooks/_processPaste.ts
+++ b/src/client/components/EditorWYSIWYG/hooks/_processPaste.ts
@@ -1,6 +1,9 @@
+import { TFunction } from 'i18next'
 import type { IJodit } from 'jodit/esm/types/jodit'
 
-export const _processPaste = (_: IJodit, html: string) => {
+import { ToasterHook } from 'client/hooks/useToaster'
+
+export const _processPaste = (toaster: ToasterHook['toaster'], t: TFunction) => (_: IJodit, html: string) => {
   try {
     const tempDiv = document.createElement('div')
     tempDiv.innerHTML = html || ''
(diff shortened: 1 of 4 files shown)


src/i18n/resources/en/dataSource.json
```diff
@@ -12,5 +12,6 @@
   "typeOfDataSourceText": "$t(dataSource.typeOfDataSource)",
   "variable": "Variable",
   "vegetationMapping": "Vegetation mapping",
-  "yearForDataSource": "Year for data source"
+  "yearForDataSource": "Year for data source",
+  "contentUpdatedOnlyLinks": "Your content has been updated, because it contains invalid HTML.\nOnly text and links can be pasted in this field."
```
minotogna: I would put the message to the generic validation messages , since this editor is generic and not necessarily related to data sources only .
What do you think ?

src/client/components/EditorWYSIWYG/hooks/_processPaste.ts
```diff
@@ -29,6 +32,28 @@ export const _processPaste = (_: IJodit, html: string) => {
...
+      'figure',
+      'hr',
+      'br',
+      'ul',
+      'ol',
+      'li',
+      'blockquote',
+    ]
+
+    if (visualElements.some((element) => html.toLowerCase().includes(`<${element}`))) {
```
minotogna, line 53: Should we do the opposite instead ? Only allowing certain tags ?
What do you think ?

sorja: Causes issue when pasting, better to do like it is (sent more info on slack)

minotogna: Sure @sorja 👍

minotogna (approved):

PR #4464, opened 2025-03-17
4450 - remove status from materialized view

sorja:


src/server/repository/assessmentCycle/country/getMany.ts
```diff
@@ -17,7 +17,8 @@ export const getMany = async (props: Props, client: BaseProtocol = DB): Promise<
 
   return client.map<Country>(
     `
-        select c.*,
+        select c.country_iso,
+               props || jsonb_build_object('status', c.status) as props,
```
minotogna, line 21: Why is status added to props ?
Should we update now the country type and (country summary if needed)?

src/server/repository/assessmentCycle/countrySummary/getOneOrNone.ts
```diff
@@ -18,8 +18,9 @@ export const getOneOrNone = async (props: Props, client: BaseProtocol = DB): Pro
 
   return client.oneOrNone<CountrySummary>(
     `
-        select *
+        select *, c.status
```
minotogna: Should we start exploiting the fields we select ?

src/server/repository/assessmentCycle/countrySummary/getMany.ts
```diff
@@ -22,8 +22,9 @@ export const getMany = async (props: Props, client: BaseProtocol = DB): Promise<
 
   return client.map(
     `
-        select *
-        from ${schemaCycle}.country_summary c
+        select cs.*, c.status
```
minotogna: Should we start exploiting the fields we select ?

sorja: sure, ill add this

src/server/repository/assessmentCycle/country/getMany.ts
sorja: i did this but rolled back as it was a bit larger refactor than i expected, this is only to keep everything working
I will separate status from props in upcoming PRs

Regarding this also:
Are we going to remove props altogether from e.g. cycleSchema.country?
I think the general idea of props or properties is not bad at all, it's a way of grouping

We can later create a jsonb object from the columns which is then used as it is now?

Either way is fine, so in the end, should it be:

country.status

or

country.props.status?


Here is current type def.


```typescript
export type CountryProps = {
  deskStudy: boolean
  domain: string // ex: tropical
  forestCharacteristics: { useOriginalDataPoint: boolean }
  hideContactsTable?: boolean
  status: AssessmentStatus
}

export type Country = {
  countryIso: CountryIso
  props?: CountryProps
  regionCodes?: Array<RegionCode>
} & Pick<CountrySummary, 'lastAccepted' | 'lastEdit' | 'lastForApproval' | 'lastInReview' | 'lastUpdate'>
```

minotogna: If we separate country props now , it will be a super long task . We can analyze all props and decide later on . What do you think ?

sorja: Sure, but should we keep status inside props or not?

1. Refactor only status now and other props in the future
2. Keep status inside props (typescript definition) and refactor all props later

minotogna: We can even leave country.props and put status in it as you did .
Let's talk in the coming days about it ?

sorja: Sure, i'll keet like it is now and resolve the other comments 👍 👍

minotogna (approved):

PR #4478, opened 2025-03-20
4450 - API middleware: Populate base params

sorja:
```
export interface RequestContext {
  assessment?: Assessment
  cycle?: Cycle
  country?: Country
}

declare module 'express' {
  interface Request {
    context: RequestContext
  }
}
```

Usage

```
request.context.assessment
request.context.cycle
request.context.country
```


src/server/serverInit.ts
```diff
@@ -32,6 +33,8 @@ export const serverInit = () => {
 
   app.use(compression({ threshold: 512 }))
 
+  app.use(RequestPopulateMiddleware.populateBaseParams)
```
minotogna: Does this work only on /api* ?

src/meta/types/express.d.ts
```diff
@@ -0,0 +1,16 @@
+import { Country } from 'meta/area'
+import { Assessment, Cycle } from 'meta/assessment'
+
```
minotogna: Should this be under server/types/express ?

src/server/serverInit.ts
```diff
@@ -7,6 +7,7 @@ import * as express from 'express'
 import * as morgan from 'morgan'
 
 import { Api } from 'server/api'
+import { RequestPopulateMiddleware } from 'server/middleware'
```
minotogna: I would export this from server/middleware/requestApiContext and call this RequestApiContextMiddleware ?

What do you think ?

src/server/middleware/populateBaseParams.ts
```diff
@@ -0,0 +1,38 @@
...
+import { AssessmentController } from 'server/controller/assessment'
+import { CountryRepository } from 'server/repository/assessmentCycle/country'
+
+type BaseType = {
+  assessmentName?: AssessmentName
+  cycleName?: CycleName
+  countryIso?: CountryIso
+}
+
+export const populateBaseParams = async (req: Request, _: Response, next: NextFunction): Promise<void> => {
```
minotogna: And call this method init or initContext ?

minotogna (commented):
Nice 😊 just some really minor detail

src/server/serverInit.ts
minotogna: > I would export this from server/middleware/requestApiContext and call this RequestApiContextMiddleware ?
> 
> What do you think ?

Or simply ApiContextMiddleware with one method init ?

src/meta/types/express.d.ts
sorja: Let me know - we can have it under types for now (also to avoid possible to avoid barrel files, although not related to this probably):
Let's say we add several more files, I think it's ok to have them separated as files vs folders and the folder structure is still well readable

Let me know which you prefer:

image

src/server/middleware/populateBaseParams.ts
sorja: Updated like so:

image

src/meta/types/express.d.ts
minotogna: I meant under server/types/express.d.ts ?

sorja: Ah sure, it was my initial implementation there, but was under the impression all (most) types should go under meta -> I will move it there!

sorja: [f1565c0](https://github.com/openforis/fra-platform/pull/4478/commits/f1565c0450a71b997aeedc32e9d4db4145ac5045)

src/server/middleware/apiContext.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const initContext = async (req: Request, _: Response, next: NextFunction): Promise<void> => {
+  try {
+    const params = { ...req.params, ...req.query, ...req.body } as BaseType
+    const { assessmentName, cycleName, countryIso } = params
+
+    Objects.setInPath({ obj: req, path: ['context'], value: {} })
+
+    if (assessmentName && cycleName) {
```
minotogna, line 22: Can we have only assessment and not cycle ?

src/server/serverInit.ts
```diff
@@ -43,6 +44,7 @@ export const serverInit = () => {
   /*
    * Initialize API
    */
+  app.use('/api*', ApiContextMiddleware.initContext)
```
minotogna: Should this go within API.init method ? What do yo think ?

sorja: moved - (note: for me it was more clear to have them outside to see at glance all middlewares)

src/server/middleware/apiContext.ts
sorja: @minotogna pr updated

minotogna: I don't see the change yet in GitHub

src/server/middleware/apiContext.ts
```diff
@@ -0,0 +1,41 @@
...
+      const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName })
+      Objects.setInPath({ obj: req, path: ['context', 'assessment'], value: assessment })
+
+      if (countryIso) {
+        const country = await CountryRepository.getOne({ assessment, cycle, countryIso })
```
minotogna, line 31: We should use controller here

sorja: Next PR will replace this, hope it's ok

src/server/middleware/apiContext.ts
sorja: image

minotogna: Opppss sorry Miró, what I meant is : 

Do we have api request where we pass only assessmentName and not cycle name ?
If yes and in the request we don't have cycle name , we should set in the context only assessment ,
If no then we should set both in the context, assessment and cycle 😅😅

src/server/middleware/apiContext.ts
minotogna: Sure

src/server/middleware/apiContext.ts
sorja: Ok, I will double check, I noticed we have single endpoint (getAreas?) that check for default assessment through settings table and default cycle

minotogna: @sorja I think it's better if we address this in a separate task. Would that work for you ?

sorja: Sure, would seem large to fit in this PR anyway
I will anyway take a look but not change anything related

minotogna: Adding assessment and cycle to the context is a bigger task , since all endpoints , auth middleware, jobs will be refactored .
Plus we need to add them to the cache in an optimized way as well .
It is a big task, in the pipeline and it must be detailed and planned yet .

Also, adding assessment and cycle to the context now has no positive impact, since we're not going to use them in the scope of this pr; au contrarie it's adding a database read operation not used.

And we're spending energy on this now that I would rather spend later.

minotogna (approved):

PR #4488, opened 2025-03-24
4450 - Update country status on data changes

sorja:
- [x] contacts
- [x] descriptions
- [x] table data
- [x] odp
- [x] check if other


src/server/controller/area/updateCountryStatus.ts
```diff
@@ -0,0 +1,40 @@
...
+  | {
+      country: Country
+    }
+  | {
+      countryIso: CountryIso
+    }
+)
+
+// TODO: Better name, progressCountryStatus, startCountry, initCountry,..
+export const updateCountryStatus = async (props: Props, client: BaseProtocol = DB) => {
```
sorja: How should we call this @minotogna

src/server/api/cycleData/contacts/create.ts
```diff
@@ -22,6 +23,8 @@ export const createContact = async (req: CycleDataRequest<never, Body>, res: Res
     const props = { assessment, cycle, countryIso, sectionName, user, contact }
     const createdContact = await CycleDataController.Contacts.create(props)
 
+    await AreaController.updateCountryStatus({ assessment, cycle, countryIso, user })
```
minotogna: Can we have is it possible to use a middleware instead ? Have you tried ?

sorja: Sure, we can do with middleware, like this?

image

I guess not global with pattern matching the urls, I think this should be ok

Let me know

sorja: Added

src/server/api/cycleData/index.ts
```diff
@@ -65,19 +66,35 @@ export const CycleDataApi = {
...
+      ApiEndPoint.CycleData.Table.nodes(),
+      AuthMiddleware.requireEditTableData,
+      CountryStatusMiddleware.updateCountryStatus,
+      persistNodeValues
+    )
     express.post(
       ApiEndPoint.CycleData.Table.estimate(),
       queue({ activeLimit: 1 }),
       AuthMiddleware.requireEditTableData,
+      CountryStatusMiddleware.updateCountryStatus,
```
minotogna: Does after the endpoint work ? If for some reason , the endpoint update fails

sorja: Good point, currently it would first run the country update, then the endpoint handler (e.g. createContact) which means if createContact fails , the status is already updated

sorja: Converting to draft until resolved

I think the previous implementation with the controlelr was the most straightforward - let me know if you had some other idea regarding the middleware

sorja: Testing changing orders, will get back to this soon...

sorja: So as I thought, we end the request with `Requests.send(res, createdContact)`, so we never get to the middleware that is defined after the route handler

I will revert the middleware approach, let me know

minotogna: yes, and that country update has to happen in the same transaction of the outside edit operation.
So the controller that performs the update (e.g. data) should then call the CountryRepository.updateCountryEditing => this method (you can find a better name) should update the status if and only if the country.status = not started

minotogna: what do you think @sorja ?

sorja: Sounds good

src/server/service/country/setCountryStatusEditing.ts
```diff
@@ -0,0 +1,28 @@
...
+  assessment: Assessment
+  cycle: Cycle
+  countryIso: CountryIso
+  user: User
+}
+
+export const setCountryStatusEditing = async (props: Props, client: BaseProtocol = DB) => {
+  const { assessment, cycle, countryIso, user } = props
+
+  const country: Country = await CountryRepository.getOne({ assessment, cycle, countryIso }, client)
```
sorja: should we use cached country? probably not to avoid stale data editing?

minotogna: we can use the cached country. I think it's safe?

src/server/repository/assessmentCycle/country/update.ts
```diff
@@ -1,27 +1,33 @@
...
-  props: { country: Country; countryIso: CountryIso; assessment: Assessment; cycle: Cycle },
-  client: BaseProtocol = DB
-): Promise<Country> => {
+import { getOne } from './getOne'
+
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+  country: Country
+  countryIso?: CountryIso
```
minotogna: why do we pass both ?

sorja: boh, from previous implementation, will remove

src/server/service/country/setCountryStatusEditing.ts
minotogna: actually, can we pass the country from the request ? ;)

sorja: we need to pass country for each controller then, this is ok?

minotogna: why not. the same way we pass the assessment and cycle. what do you think?

sorja: sounds good to me

src/server/api/cycleData/table/clearTable.ts
```diff
@@ -18,11 +20,12 @@ export const clearTable = async (req: CycleDataRequest, res: Response) => {
 
     const nodes = await CycleDataController.clearTableData({
       assessment,
+      country,
```
minotogna: Are there many cases where we before passed countryIso and now country ?
We should probably avoid to pass both ?
What do you think ?

sorja: Most if not all - I will replace (it was the initial implementation, then I added back countryIso..! :D )

minotogna (approved):

sorja:
@minotogna fixed test

minotogna:
> @minotogna fixed test

Sure @sorja

PR #4489, opened 2025-03-24
4481 - PDF: reports in different languages

### LINKED ISSUE #4481: PDF: reports in different languages

minotogna:


src/tools/report/ISO3_country_lan.csv
```diff
@@ -0,0 +1,237 @@
+ISO3,Country,Language,,,,,,
```
sorja, line 1: Should we have country.props.reportLanguage ? (For future)

sorja (approved):

src/tools/report/ISO3_country_lan.csv
minotogna: Maybe . The country language is an ongoing discussion , and so far no conclusion yet 😅
Let's see how this will go

PR #4495, opened 2025-03-27
4450 - Notify client in real time of country status change

sorja:
## Note:
This PR concludes main issue

image

Note that the change is _also_ visible in the country selector if another user is viewing the list, which is nice:
image


src/client/pages/Cycle/hooks/useCountryStatusListener.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { useAppDispatch } from 'client/store'
+import { AreaActions } from 'client/store/area'
+import { useAssessment, useCycle } from 'client/store/assessment'
+import { SocketClient } from 'client/service/socket'
+
+export const useCountryStatusListener = (): void => {
+  const dispatch = useAppDispatch()
+  const assessment = useAssessment()
+  const cycle = useCycle()
```
minotogna: Assessment name and cycle name can be retrieved from route params hook .

src/client/pages/Cycle/hooks/useCountryStatusListener.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    const eventName = Sockets.getCountryStatusUpdateEvent({
+      assessmentName: assessment.props.name,
+      cycleName: cycle.name,
+    })
+
+    const handleStatusUpdate = (args: [{ [key: string]: AssessmentStatus }]) => {
```
minotogna: If the event name contains assessment name , cycle name and country iso , then args can be just one parameter status .

src/server/service/socket/socketService/country/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  assessmentName: AssessmentName
+  cycleName: CycleName
+  countryIso: CountryIso
+  status: AssessmentStatus
+}
+
+const notifyStatusUpdate = (props: props) => {
+  const { countryIso, assessmentName, cycleName, status } = props
+  const eventName = Sockets.getCountryStatusUpdateEvent({ assessmentName, cycleName })
+  SocketServer.emit(eventName, { [countryIso]: status })
```
minotogna: Better to be country specific . Add country iso to event name . What do you think ?

src/server/controller/area/updateCountry.ts
```diff
@@ -32,6 +33,15 @@ export const updateCountry = async (props: Props, client: BaseProtocol = DB): Pr
     const message = ActivityLogMessage.assessmentStatusUpdate
     const activityLog = { target, section: 'assessment', message, countryIso, user }
     await ActivityLogRepository.insertActivityLog({ activityLog, assessment, cycle }, t)
+
+    // notify client
+    SocketService.Country.notifyStatusUpdate({
```
minotogna: Should we notify client only if status is updated ?

src/server/service/socket/socketService/country/index.ts
sorja: It was initial implementation, but this is more flexbile, we don't need the country information in the event name anyway
E.g. if we want to send many countries?

src/server/controller/area/updateCountry.ts
sorja: +

src/client/pages/Cycle/hooks/useCountryStatusListener.tsx
sorja: +

src/server/service/socket/socketService/country/index.ts
sorja: +

src/client/pages/Cycle/hooks/useCountryStatusListener.tsx
```diff
@@ -0,0 +1,39 @@
...
+      countryIso,
+    })
+
+    const handleStatusUpdate = (args: [{ status: AssessmentStatus }]) => {
+      const [{ status }] = args
+      dispatch(
+        AreaActions.updateCountryStatus({
+          assessmentName,
+          cycleName,
+          countryIso: countryIso as CountryIso,
```
minotogna: U don't need to cast here

minotogna (approved):

minotogna (approved):

PR #4536, opened 2025-04-02
4525 -  Add new fields to cycleSchema.country and update createSchemaDDL 

sorja:
Next PR: Remove fields from materialized view and fix queries using it (join from schemaCycle.country)


minotogna (approved):
👍

PR #4539, opened 2025-04-02
4484 - PDF - Tab 8 SDG number alignment fix

### LINKED ISSUE #4484: PDF - Tab 8 SDG number
@minotogna 

**Describe the bug**
The numbers in tab 8 are not in the same horizontal line.

**To Reproduce**
Go to any country report -  tab 8a

**Screenshots**

Image

#4452

yaguzmang:
Table Before/after

image

image

PDF before/after:
image


src/client/pages/Section/DataTable/Table/RowData/Cell/Number/Number.scss
```diff
@@ -2,6 +2,10 @@
 
 .table-grid__data-cell-input-number {
   text-align: right;
+
+  &.disabled {
+    height: unset;
+  }
```
yaguzmang, line 8: The misalignment happens because when the input is disabled, the displayed div with the number has height=100%; and the content is not centered internally.
image

Previously to avoid the misalignment  we had:
image

To avoid having to add the flex property, we can simply unset the height and the parent div will center the input disabled.

Or alternatively, we could center it directly in the input-text component but test thoroughly everywhere we use it

minotogna: @yaguzmang thanks for finding the problem. 
I removed those classes because it's a bit strange to have different style when a cell is disabled. the layout should not be affected when it's disabled and when it's not.
Plus, calculated cells should have the same layout than the numeric cells. 
The best solution would be to have same layout between numeric and calculated cells; and the layout should not change when the cells are disabled. is there any possible alternative keeping this in mind? If it takes too long, we'll think about this later and we can move with the proposed solution. 
please let me know . thank you

minotogna (approved):

PR #4549, opened 2025-04-04
4514 - User Accept Invitation: fix wrong redirect

### LINKED ISSUE #4514: User Accept Invitation: wrong redirect
When a user accepts an invitation , after accepting, he should be redirected to the county/cycle just accepted. now it goes to 2020

tested on latest Finland , invited a national correspondent

yaguzmang:
TODO: 
- [x] Fix redirect when accepting with google (via api)

Accept invitation local:

https://github.com/user-attachments/assets/5d565008-5ffb-478a-b81c-09ea6fb7851f

Accept invitation with google:

https://github.com/user-attachments/assets/8bac7477-df5e-4990-af0e-7e24cc135224

Logins unaffected:

https://github.com/user-attachments/assets/ad01ef46-560f-408b-abf2-2d5a5c1e45fa

https://github.com/user-attachments/assets/237afe2f-4fbc-42ec-bcf1-1f46a931e92f

### DIFF
--- a/src/client/pages/Login/components/AcceptInvitationButtons/hooks/useOnAcceptInvitationLocal.ts
+++ b/src/client/pages/Login/components/AcceptInvitationButtons/hooks/useOnAcceptInvitationLocal.ts
@@ -4,8 +4,9 @@ import { useNavigate } from 'react-router-dom'
 import { Routes } from 'meta/routes'
 
 import { useAppDispatch } from 'client/store'
-import { LoginActions } from 'client/store/login'
+import { LoginActions, useInvitation } from 'client/store/login'
 import { AcceptInvitationFormState } from 'client/store/login/stateType'
+import { useCycleRouteParams } from 'client/hooks/useRouteParams'
 import { isError, LoginValidator } from 'client/pages/Login/utils/LoginValidator'
 
 type Props = {
@@ -23,6 +24,11 @@ export const useOnAcceptInvitationLocal = (props: Props): Returned => {
   const navigate = useNavigate()
   const dispatch = useAppDispatch()
 
+  const { userInvitation } = useInvitation()
+  const { countryIso } = userInvitation
+  const { assessmentName, cycleName } = useCycleRouteParams()
+  const redirectUrl = Routes.Country.generatePath({ assessmentName, countryIso, cycleName })
+
   return useCallback<Returned>(() => {
     const fieldErrors = showPassword2
       ? LoginValidator.invitationValidate(email, password, password2)
@@ -38,8 +44,8 @@ export const useOnAcceptInvitationLocal = (props: Props): Returned => {
         })
       ).then(() => {
         dispatch(LoginActions.resetAcceptInvitationForm())
-        navigate(Routes.Root.path.absolute)
+        navigate(redirectUrl)
       })
     }
-  }, [dispatch, email, invitationUuid, navigate, password, password2, showPassword2])
+  }, [dispatch, email, invitationUuid, navigate, password, password2, redirectUrl, showPassword2])
 }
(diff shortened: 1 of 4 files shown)


src/client/pages/Login/components/AcceptInvitationButtons/hooks/useOnAcceptInvitationLocal.ts
```diff
@@ -23,6 +24,11 @@ export const useOnAcceptInvitationLocal = (props: Props): Returned => {
   const navigate = useNavigate()
   const dispatch = useAppDispatch()
 
+  const { userInvitation } = useInvitation()
+  const { countryIso } = userInvitation
+  const { assessmentName, cycleName } = useCycleRouteParams()
```
sorja: not sure if it's better to extract these from invitation if they exist in the type?
i think this might be ok too though

sorja:
P.s. also check file: `useUserRedirect`

@yaguzmang Just to update:

I _think_ if I remember right that the idea was here to always route the user to country home from which the user is then redirected to correct place
This is to have a single place for redirect logic

src/client/pages/Login/components/AcceptInvitationButtons/hooks/useOnAcceptInvitationLocal.ts
yaguzmang: From `useInvitation()`:
image
In `userInvitation` it has assessmentUuid and cycleUuid. I can cross reference with `assessment` and get the names from it, but thought using params adds less complexity. What do you think?

sorja: I think this should be fine as long as we know (which we should) that assessmentName and cycleName are of the invitation

src/client/pages/Login/components/AcceptInvitationButtons/hooks/useOnAcceptInvitationLocal.ts
```diff
@@ -23,6 +24,11 @@ export const useOnAcceptInvitationLocal = (props: Props): Returned => {
...
+  const redirectUrl = Routes.Country.generatePath({ assessmentName, countryIso, cycleName })
```
yaguzmang, line 31: ~If we want to avoid the redirect to user profile from the Country route, we can send the user directly to Country Home -> overview. Let me know if that's preferable and I can make the change.~

As Miro noted, we want to route the user to where `useUserRedirect` is used, which it is in the base `Country` route.

src/client/pages/Login/components/AcceptInvitationButtons/hooks/useOnAcceptInvitationLocal.ts
yaguzmang: I noticed in other places we get assessment and cycle name from the user invitation, so I changed this now for consistency 👍

sorja (approved):

src/server/api/auth/login.ts
```diff
@@ -51,7 +52,13 @@ export const getGoogleCallback = (req: Request, res: Response, next: NextFunctio
       req.login(user, (err: any) => {
         if (err) next(err)
         setAuthToken(res, user)
-        const redirectUrl = process.env.NODE_ENV === 'development' ? '/' : ProcessEnv.appUri
+
+        const countryPath = !Objects.isEmpty(countryIso)
+          ? Routes.Country.generatePath({ assessmentName, countryIso, cycleName })
+          : ''
+        const redirectUrl =
+          process.env.NODE_ENV === 'development' ? countryPath || '/' : `${ProcessEnv.appUri}${countryPath}`
```
sorja, line 60: p.s. here better to use `ProcessEnv`

PR #4559, opened 2025-04-07
4535 - Settings: use assessment.props.default

### LINKED ISSUE #4535: Settings: use assessment.props.default
- [ ] Remove settings from db 
- [ ] add assessment.props.default (boolean) (fra is the default to set to true )

sorja:


src/server/api/area/getAreas.ts
```diff
@@ -4,16 +4,26 @@ import { InitRequest } from 'meta/api/request'
...
 export const getAreas = async (req: InitRequest, res: Response) => {
   try {
     const { assessmentName, cycleName } = req.query
 
-    const settings = await SettingsController.read()
-    const props = assessmentName ? { assessmentName } : { id: settings.defaultAssessmentId }
-    const { assessment, cycle } = await AssessmentController.getOneWithCycle({ ...props, cycleName })
+    let assessmentProps: { assessmentName: string }
+
+    if (assessmentName) {
```
minotogna: can we assume we always pass assessmentName to this endpoint ?

src/client/pages/Landing/hooks/useRedirectAssessmentAndCycle.tsx
```diff
@@ -2,19 +2,17 @@ import { Assessments } from 'meta/assessment/assessments'
...
 
-import { useAssessments } from 'client/store/assessment'
-import { useSettings } from 'client/store/assessment/hooks'
+import { useAssessmentDefault, useAssessments } from 'client/store/assessment'
 import { useUser } from 'client/store/user'
 
 export const useRedirectAssessmentAndCycle = () => {
   const assessments = useAssessments()
-  const { defaultAssessmentId } = useSettings()
+  const assessmentDefault = useAssessmentDefault()
```
minotogna: I don't think we need a separate hook with selector included to get the default assessment .
I'd rather isolate the inner logic of getting the assessment. you can have an inner hook useInitialAssessment that internally uses only assessments . then it returns 
```
const findFn = userLastRole?.assessmentUuid ? (assessment) => assessment.uuid === userLastRole?.assessmentUuid) : (assessment) => assessment.props.default
return assessments.find(findFn)
```
what do you think ?

sorja: It is used also in useAssessment -hook

image

sorja: What do you suggest?
Remove selector and do find also here?

src/client/pages/Landing/hooks/_useInitialAssessment.tsx
```diff
@@ -0,0 +1,14 @@
+import { UserRoles } from 'meta/user/userRoles'
+
+import { useAssessmentDefault, useAssessments } from 'client/store/assessment'
+import { useUser } from 'client/store/user'
+
+export const _useInitialAssessment = () => {
+  const assessments = useAssessments()
+  const assessmentDefault = useAssessmentDefault()
```
minotogna: Let's remove this use assessment default + selector ?
What do you think @sorja ?

src/client/pages/Landing/hooks/useRedirectAssessmentAndCycle.tsx
minotogna: I don't know . Let's see later

src/client/pages/Landing/hooks/_useInitialAssessment.tsx
sorja: Sure - should I pass assessmentUuid optional to useAssessment?

```
export const useAssessment = (): Assessment => {
  const { assessmentName } = useAssessmentRouteParams()

  return useAppSelector((state) => {
    if (assessmentName) return AssessmentSelectors.getAssessment(state, assessmentName)
    return AssessmentSelectors.getDefaultAssessment(state)
  })
}

```

would become
```
export const useAssessment = (assessmentUuid?: string): Assessment => {
  const { assessmentName } = useAssessmentRouteParams()

  return useAppSelector((state) => {
    if (aUuid) return AssessmentSelectors.getAssessmentUuid(state, assessmentName)
    if (assessmentName) return AssessmentSelectors.getAssessment(state, assessmentName)
    return AssessmentSelectors.getDefaultAssessment(state)
  })
}
```

sorja: Selector is used here

src/client/pages/Landing/hooks/useRedirectAssessmentAndCycle.tsx
```diff
@@ -1,20 +1,13 @@
...
 export const useRedirectAssessmentAndCycle = () => {
-  const assessments = useAssessments()
-  const { defaultAssessmentId } = useSettings()
   const user = useUser()
 
-  const userLastRole = UserRoles.getLastRole({ user })
-  const assessment = assessments.find(
-    (assessment) => Number(assessment.id) === Number(userLastRole?.assessmentId ?? defaultAssessmentId)
-  )
+  const assessment = _useInitialAssessment()
```
minotogna: Can we use the convention local hooks do not start with _ and are imported with local path .
I don't really mind what convention we use about it , as long as we use one .
What do you think ? @sorja

sorja: sure

minotogna (approved):

PR #4579, opened 2025-04-11
4541 - CycleController.publish

sorja:
- publish non published countries for cycle
- set activity logs
- set cycle to published

### DIFF
--- a/src/server/repository/assessmentCycle/country/publishAll.ts
+++ b/src/server/repository/assessmentCycle/country/publishAll.ts
@@ -19,8 +19,9 @@ export const publishAll = async (props: Props, client: BaseProtocol = DB): Promi
   const countryIsos = await client.map(
     `
     update ${schemaName}.country
-    set last_in_published = now()
-    where status != 'published'
+    set last_in_published = now(),
+        status = 'published'
+    where status = 'accepted'
     returning country_iso
   `,
     [],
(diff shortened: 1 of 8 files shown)


src/server/repository/assessmentCycle/country/publishAll.ts
```diff
@@ -0,0 +1,31 @@
...
+import { BaseProtocol, DB, Schemas } from 'server/db'
+
+import { getMany } from './getMany'
+
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+}
+
+export const publishAll = async (props: Props, client: BaseProtocol = DB): Promise<Array<Country>> => {
```
minotogna: I think it's more consistent to publish all the 'accepted' countries only 

Let's say a country is still in editing and not accepted , I suspect admin would want to keep that country in editing .
What do you think ?

sorja: Sure, now it works nicely

minotogna: should we rename the method as well in publishAllAccepted ?

sorja: renamed

minotogna (approved):

PR #4580, opened 2025-04-11
4555 - PDF - Title & Content Separation fix

### LINKED ISSUE #4555: PDF - Title & Content Separation
@minotogna 

**Describe the bug**
The content (table) is separated from the title (spanning two pages with only the title and subtitle).

**To Reproduce**
[Indonesia](https://unfao-my.sharepoint.com/:b:/g/personal/fra-platform_fao_org/ERbtuZmZvelKtHk-atfBVlYBOyjAy2Fr9YejL9wfLyvVeQ?e=XnKwEX) page 61 - 67.
Similarly, [Indonesia](https://unfao-my.sharepoint.com/:b:/g/personal/fra-platform_fao_org/ERbtuZmZvelKtHk-atfBVlYBOyjAy2Fr9YejL9wfLyvVeQ?e=XnKwEX) page 69 - 71.

yaguzmang:
Before/after
image

image

To consider: Because the data cells now don't split between pages, some borders can look like this when the next row goes to the next page:
image

This is because of how draw the borders. Ideally the border bottom would be there, but targeting that last row in the page didn't seem possible with css.

### DIFF
--- a/src/client/components/DataGrid/DataCell/DataCell.scss
+++ b/src/client/components/DataGrid/DataCell/DataCell.scss
@@ -168,3 +168,9 @@ $highlight-top-bottom: inset 0 1px $ui-accent, inset 0 -1px $ui-accent;
     border: unset;
   }
 }
+
+@include print {
+  .data-cell:not(:has(.editorWYSIWYG)) {
+    @include print-break-inside-avoid-page;
+  }
+}
--- a/src/client/components/DataGrid/DataGrid/DataGrid.tsx
+++ b/src/client/components/DataGrid/DataGrid/DataGrid.tsx
@@ -17,7 +17,7 @@ const DataGrid = forwardRef<HTMLDivElement, Props>((props, outerRef) => {
   }, [gridColumn, gridTemplateColumns, withActions])
 
   return (
-    <div ref={outerRef} className={classNames('data-grid print-break-inside-avoid', className)} style={style}>
+    <div ref={outerRef} className={classNames('data-grid', className)} style={style}>
       {React.Children.toArray(children)}
     </div>
   )
--- a/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
+++ b/src/client/components/EditorWYSIWYG/EditorWYSIWYG.scss
@@ -72,8 +72,7 @@ div.editorWYSIWYG {
     }
 
     tr {
-      break-inside: avoid-page;
-      page-break-inside: avoid;
+      @include print-break-inside-avoid-page;
     }
   }
 }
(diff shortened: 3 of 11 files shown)


src/client/components/DataGrid/DataCell/DataCell.scss
```diff
@@ -168,3 +168,9 @@ $highlight-top-bottom: inset 0 1px $ui-accent, inset 0 -1px $ui-accent;
     border: unset;
   }
 }
+
+@include print {
+  .data-cell:not(:has(.editorWYSIWYG)) {
+    @include print-break-inside-avoid-page;
+  }
+}
```
yaguzmang, line 176: As per [w3.org](https://www.w3.org/TR/css-grid-1/#pagination)
> The [break-before](https://www.w3.org/TR/css-break-3/#propdef-break-before) and [break-after](https://www.w3.org/TR/css-break-3/#propdef-break-after) properties on [grid items](https://www.w3.org/TR/css-grid-1/#grid-item) are propagated to their grid row.

So it works similarly to what we do for tables
```
    tr {
      @include print-break-inside-avoid-page;
    }
 ```
And only applying this to non-editorWYSIWYG cells because when the cell has large user-entered input, the same issue happens where the title/content splits.

minotogna:
> Before/after
> 
> image
> 
> 
> 
> image
> 
> 
> 
> To consider: Because the data cells now don't split between pages, some borders can look like this when the next row goes to the next page:
> 
> image
> 
> 
> 
> This is because of how draw the borders. Ideally the border bottom would be there, but targeting that last row in the page didn't seem possible with css. 

@pengchenglai please consider the side effect of this change .
Is that ok ?

pengchenglai:
> > Before/after
> > image
> > image
> > To consider: Because the data cells now don't split between pages, some borders can look like this when the next row goes to the next page:
> > image
> > This is because of how draw the borders. Ideally the border bottom would be there, but targeting that last row in the page didn't seem possible with css.
> 
> @pengchenglai please consider the side effect of this change . Is that ok ?

@minotogna @yaguzmang 

Thank you for letting me know and for the efforts that have been made. 
This is a bit tricky, considering there are many tables that span across pages—without bottom borders, it could be noticeable in many reports. Let us hold off on this issue for now and preserve the previous settings. I will be back in the office next week and will also bring this issue up for discussion with the admin.

yaguzmang:
@pengchenglai 

> Thank you for letting me know and for the efforts that have been made. This is a bit tricky, considering there are many tables that span across pages—without bottom borders, it could be noticeable in many reports. Let us hold off on this issue for now and preserve the previous settings. I will be back in the office next week and will also bring this issue up for discussion with the admin.


The only other alternative we have is to switch the borders. We would miss the top border in the carried row, instead of missing the bottom border in the last row of the previous page. This is how it would look:
image

Would this be a better solution for you?

pengchenglai:
@minotogna @yaguzmang 

Thank you for the updates! It looks much better and plz let me check with OJ this afternoon and will provide his feedback.

minotogna:
> @minotogna @yaguzmang 
> 
> 
> 
> Thank you for the updates! It looks much better and plz let me check with OJ this afternoon and will provide his feedback.

Sure. Thank you @pengchenglai

minotogna (approved):

PR #4582, opened 2025-04-16
4060 - API Documentation - Add Node schemas

yaguzmang:
(UPDATED)

https://github.com/user-attachments/assets/6854cf87-e078-47e8-b5ac-7434bbd932a2

image

image

### DIFF
--- a/src/docs/api/schemas/NodeValue.yml
+++ b/src/docs/api/schemas/NodeValue.yml
@@ -1,32 +1,5 @@
 components:
   schemas:
-    NodeValueValidation:
-      type: object
-      properties:
-        valid:
-          type: boolean
-        messages:
-          type: array
-          items:
-            $ref: '#/components/schemas/NodeValueValidationMessage'
-
-    NodeValueValidationMessage:
-      type: object
-      properties:
-        key:
-          type: string
-        params:
-          type: object
-          additionalProperties:
-            oneOf:
-              - type: string
-              - type: number
-              - type: array
-                items:
-                  oneOf:
-                    - type: string
-                    - type: number
-
     NodeValue:
       type: object
       properties:
@@ -47,8 +20,3 @@ components:
         estimated:
           type: boolean
           deprecated: true
-        odp:
-          type: boolean
-          deprecated: true
-        validation:
-          $ref: '#/components/schemas/NodeValueValidation'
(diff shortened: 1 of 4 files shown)


src/docs/api/schemas/RecordAssessmentData.yml
```diff
@@ -0,0 +1,31 @@
+components:
+  schemas:
+    RecordRowData:
```
minotogna, line 3: from the user point of view these types are very unclear .
Is it possible to explicit the types? e.g. `Record<CountryIso,...>` ?

yaguzmang: The OpenAPI Specification doesn't support named types for the property keys, but I can add a title:
```
    RecordAssessmentData:
      title: Record<Assessment Name, Record Cycle Data>
      type: object
      additionalProperties:
        $ref: '#/components/schemas/RecordCycleData'
 ```
and then it displays like this:

image

I can add descriptions too to make it clearer:
image

Comment (yaguzmang) [reply src/docs/api/schemas/RecordAssessmentData.yml:3]: Would look like this:
image
image

Let me know what you think

Comment (minotogna) [reply src/docs/api/schemas/RecordAssessmentData.yml:3]: It's more readable , that's what I meant 😎

**Comment by minotogna** [comment]
@yaguzmang  the agreed endpoint contains node value validation ?

To the users , we should only give the necessary information of the agreed endpoint only

File: src/docs/api/schemas/NodeValue.yml [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+            oneOf:
+              - type: string
+              - type: number
+              - type: array
+                items:
+                  oneOf:
+                    - type: string
+                    - type: number
+
+    NodeValue:
```
yaguzmang: @minotogna I've double checked now and we are doing the validation client side, so I will remove these types from here :+1:
Also:
The property `odp` in the type NodeValue is deprecated, and it seems it's also not returned by the endpoint currently, so I will remove it too.

The property `estimated` is deprecated also, but I will keep it because we still send it:
image

minotogna: Thanks @yaguzmang !

minotogna (approved):

PR #4587, opened 2025-04-17
4060 - API Documentation - ODP Get many

yaguzmang:
image
image
image

### DIFF
--- a/src/server/api/cycleData/table/getTableData.ts
+++ b/src/server/api/cycleData/table/getTableData.ts
@@ -22,13 +22,8 @@ type GetTableDataRequest = CycleDataRequest<{
  *   get:
  *     summary: Get table data
  *     parameters:
- *       - in: query
- *         name: assessmentName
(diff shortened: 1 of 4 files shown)


src/server/api/cycleData/table/getTableData.ts
```diff
@@ -22,13 +22,8 @@ type GetTableDataRequest = CycleDataRequest<{
...
- *       - in: query
- *         name: assessmentName
- *         description: The name of the assessment. For example, "fra" or "panEuropean".
- *         required: true
- *         schema:
- *           type: string
- *           example: fra
+ *       - $ref: '#/components/parameters/assessmentName'
+ *       - $ref: '#/components/parameters/cycleName'
  *       - in: query
```
yaguzmang, line 27: Not adding the common countryIso param because the description here is different: `A single country ISO code like "FIN", or a region like "EU".`

minotogna (approved):

PR #4588, opened 2025-04-17
4060 - API Documentation

### LINKED ISSUE #4060: API Documentation
Swagger ?

yaguzmang:
image

docs TODO: 
- [x] Table data
- [x] Descriptions
- [x] ODP


src/docs/api/schemas/CycleParams.yml
```diff
@@ -0,0 +1,28 @@
...
+        example: 2020
+
+    countryIso:
+      name: countryIso
+      in: query
+      required: true
+      description: A country ISO code like "FIN".
+      schema:
+        type: string
+        example: FIN
```
sorja: possibly to dynamically generate? and include drop down for cycle/assessmentName (countryIso?)

yaguzmang: Will check if it's possible :+1:

yaguzmang: Changing a few things, it is possible to add a list of countryIsos on init, for a specific assessment/cycle:
image
```JS
export const swaggerInit = async (app: Express): Promise<void> => {
  app.get('/api-docs/swagger.json', async (_req, res) => {
    const spec = swaggerJSDoc(swaggerOptions) as OpenAPIV3.Document;

    const { assessment, cycle } = await AssessmentController.getOneWithCycle({
      assessmentName: 'fra',
      cycleName: '2020',
    });
    const countries = await AreaController.getCountries({ assessment, cycle });
    const isos = countries.map((c) => c.countryIso);

    const params = spec.components!.parameters! as Record<string, OpenAPIV3.ParameterObject>;
    params.countryIso.schema = { type: 'string', enum: isos, example: isos[0] };

    res.json(spec);
  });

  app.use(
    '/api-docs',
    swaggerUi.serve,
    swaggerUi.setup(undefined, {
      swaggerUrl: '/api-docs/swagger.json'
    })
  );
};
```
however if we wanted to make the list dynamic and depend on  assessment/cycle, we would need to instead inject code that handles that to Swagger UI. Looking online, that seems to be bug prone as it changes Swagger UIs default behaviour. 
Would it still be worthy to add it? what do you think.

minotogna: It looks nice . 
Let's leave all fra country isos only ?
What do you think ?

minotogna: They should be sorted alphabetically and without X* (Atlantis) countries

yaguzmang: Sounds good 👍 adding them here #4595

src/server/api/cycleData/originalDataPoint/getOriginalDataPoints.ts
```diff
@@ -6,6 +6,62 @@ import { AssessmentController } from 'server/controller/assessment'
...
+ * /cycle-data/original-data-points:
+ *   get:
+ *     summary: Get all Original Data Points for a given country and cycle
+ *     parameters:
+ *       - $ref: '#/components/parameters/assessmentName'
+ *       - $ref: '#/components/parameters/countryIso'
+ *       - $ref: '#/components/parameters/cycleName'
+ *     responses:
+ *       200:
+ *         description: Successful operation.
```
sorja: Can you check if it's possibe to move to another file?
e.g. <fileName>.apidoc

yaguzmang: We can use .yml files for this as well. I initially added the doc as comments in the endpoints to avoid adding more files/folders, but we can also do that if it's preferred.
We could add another folder in each api section called `docs` and then all docs for that section can go there.
image

Let me know what you think @minotogna @sorja

sorja: can we do server/docs/<same struct>?
or server/api/docs? not sure what is best, also next to the endpoint maybe ok...
maybe @minotogna has some idea

sorja:
@minotogna looks good to me

minotogna (approved):

PR #4594, opened 2025-04-22
4541 - Country Status: Publish UI

sorja:


src/meta/area/countryStatuses/index.ts
```diff
@@ -47,6 +50,11 @@ export const CountryStatuses = {
     const transitions = statusTransitions[status]
     const rolesAllowed = statusRolesAllowed[status]
 
+    // When a cycle is published, a country cannot be moved back to previous steps
+    if (Cycles.isPublished(cycle)) {
```
minotogna: Does this take into account country status ? It cannot be moved back only if country status is published .

sorja: Updated accordingly

minotogna (approved):

PR #4603, opened 2025-04-23
4060 - API Documentation - Remove panEuropan & Add Table names enum

yaguzmang:
https://github.com/user-attachments/assets/e98f19ef-c90f-4dec-8c33-7f79484a61ab

### DIFF
--- a/src/server/api/cycleData/table/getTableData.ts
+++ b/src/server/api/cycleData/table/getTableData.ts
@@ -23,6 +23,7 @@ type GetTableDataRequest = CycleDataRequest<{
  *     summary: Get table data
  *     parameters:
  *       - $ref: '#/components/parameters/assessmentName'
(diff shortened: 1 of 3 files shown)


src/server/api/cycleData/table/getTableData.ts
```diff
@@ -23,6 +23,7 @@ type GetTableDataRequest = CycleDataRequest<{
  *     summary: Get table data
  *     parameters:
  *       - $ref: '#/components/parameters/assessmentName'
+ *       - $ref: '#/components/parameters/countryIso'
```
yaguzmang, line 26: Note: countryIso is still needed because it is used in `AuthMiddleware.requireView`.  The `getTableData` method itself doesn't require it, since it uses the `countryIsos` param instead.

minotogna (approved):

PR #4614, opened 2025-04-28
4597 - Download user profile pictures

### LINKED ISSUE #4597: Download user profile pictures
Prepare a zip folder with:
- [x] all user profile picture files 
- [x] a CSV file with country iso, name, surname, role, file name

sorja:
Simple script to fetch users info and pf for given assessment/cycle

### DIFF
--- a/src/tools/csvUsersWithProfilePicture/index.ts
+++ b/src/tools/csvUsersWithProfilePicture/index.ts
@@ -22,6 +22,7 @@ type DBUser = {
   email: string
   roles: Array<DBRole>
   profile_picture_file_uuid: string
+  file_name: string
 }
 
 type CSVUser = {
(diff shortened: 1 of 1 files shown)


src/tools/csvUsersWithProfilePicture/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  fileName: string
+}
+
+const transformUser = (user: DBUser, role: DBRole): CSVUser => {
+  return {
+    countryIso: role.country_iso,
+    fullName: user.full_name,
+    email: user.email,
+    role: role.role,
+    fileName: user.profile_picture_file_uuid,
```
minotogna: Do the files have an extension ?
Or can we download them and save as their original file name ?

sorja: No extensions/file names in s3, but I can cross join public.file to get the original file name if preferred?
Note: we can risk same file name

minotogna: Yes, then concatenate original file name with _id or _uuid

minotogna (approved):

PR #4617, opened 2025-04-28
4541 - AreaSelector

sorja:
- [x] areas can be viewed by all

### DIFF
--- a/src/client/pages/Country/hooks/useInitMetaCache.tsx
+++ b/src/client/pages/Country/hooks/useInitMetaCache.tsx
@@ -4,9 +4,11 @@ import { AssessmentMetaCaches } from 'meta/assessment/metaCaches'
 
 import { useAppDispatch } from 'client/store'
 import { AssessmentActions, useAssessment, useCycle } from 'client/store/assessment'
+import { useCountryRouteParams } from 'client/hooks/useRouteParams'
 
 export const useInitMetaCache = (): void => {
   const dispatch = useAppDispatch()
+  const { countryIso } = useCountryRouteParams()
   const assessment = useAssessment()
   const cycle = useCycle()
 
@@ -16,7 +18,7 @@ export const useInitMetaCache = (): void => {
 
   useEffect(() => {
     if (!metaCache) {
-      dispatch(AssessmentActions.getMetaCache({ assessmentName, cycleName }))
+      dispatch(AssessmentActions.getMetaCache({ assessmentName, cycleName, countryIso }))
     }
-  }, [assessmentName, cycleName, dispatch, metaCache])
+  }, [assessmentName, countryIso, cycleName, dispatch, metaCache])
 }
--- a/src/client/pages/Country/hooks/useInitSections.tsx
+++ b/src/client/pages/Country/hooks/useInitSections.tsx
@@ -4,16 +4,16 @@ import { Objects } from 'utils/objects'
 
 import { useAppDispatch } from 'client/store'
 import { MetadataActions, useSections } from 'client/store/metadata'
-import { useCycleRouteParams } from 'client/hooks/useRouteParams'
+import { useCountryRouteParams } from 'client/hooks/useRouteParams'
 
 export const useInitSections = () => {
-  const { assessmentName, cycleName } = useCycleRouteParams()
+  const { assessmentName, cycleName, countryIso } = useCountryRouteParams()
   const dispatch = useAppDispatch()
   const sections = useSections()
 
   useEffect(() => {
     if (Objects.isEmpty(sections)) {
-      dispatch(MetadataActions.getSections({ assessmentName, cycleName }))
+      dispatch(MetadataActions.getSections({ assessmentName, cycleName, countryIso }))
     }
-  }, [assessmentName, cycleName, dispatch, sections])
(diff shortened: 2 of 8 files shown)


src/client/store/assessment/actions/getMetaCache.ts
```diff
@@ -14,20 +14,21 @@ import { ThunkApiConfig } from 'client/store/types'
...
   async (props, { getState }) => {
-    const { assessmentName, cycleName } = props
+    const { assessmentName, cycleName, countryIso } = props
 
     const assessment = AssessmentSelectors.getAssessment(getState(), assessmentName)
     const cycle = assessment.cycles.find((c) => c.name === cycleName)
 
     if (!AssessmentMetaCaches.getMetaCache({ assessment, cycle })) {
-      const params = { assessmentName, cycleName, countryIso: Global.WO }
+      const params = { assessmentName, cycleName, countryIso }
```
sorja, line 31: Not sure if there was some special requirement or something why we were fetching metadata and metacache with country iso, possibly it used to be on higher (assessment, cycle) level?

Updated for consistency and avoid problems (canView for unpublished cycles)

src/server/controller/area/updateCountry.ts
```diff
@@ -34,8 +34,12 @@ export const updateCountry = async (props: Props, client: BaseProtocol = DB): Pr
...
     )
 
-    // update cache
-    await AreaRedisRepository.getOneCountry({ assessment, cycle, countryIso, force: true }, t)
+    // update cache for all cycles to keep lastPublished up to date
+    await Promise.all(
+      assessment.cycles.map((c) =>
+        AreaRedisRepository.getOneCountry({ assessment, cycle: c, countryIso, force: true }, t)
+      )
+    )
```
sorja, line 42: When we update country (status) we need to update all cycle's caches to be consistent with lastPublished information in cache

src/client/pages/Country/hooks/useInitMetaCache.tsx
```diff
@@ -16,7 +18,7 @@ export const useInitMetaCache = (): void => {
 
   useEffect(() => {
     if (!metaCache) {
-      dispatch(AssessmentActions.getMetaCache({ assessmentName, cycleName }))
+      dispatch(AssessmentActions.getMetaCache({ assessmentName, cycleName, countryIso }))
```
minotogna, line 21: Isn't the meta cache per cycle ? Why passing country iso ?

src/client/pages/Country/hooks/useInitSections.tsx
```diff
@@ -4,16 +4,16 @@ import { Objects } from 'utils/objects'
...
 export const useInitSections = () => {
-  const { assessmentName, cycleName } = useCycleRouteParams()
+  const { assessmentName, cycleName, countryIso } = useCountryRouteParams()
   const dispatch = useAppDispatch()
   const sections = useSections()
 
   useEffect(() => {
     if (Objects.isEmpty(sections)) {
-      dispatch(MetadataActions.getSections({ assessmentName, cycleName }))
+      dispatch(MetadataActions.getSections({ assessmentName, cycleName, countryIso }))
```
minotogna, line 16: Also here

src/client/store/assessment/actions/getMetaCache.ts
minotogna: Was it used by authorizer?

sorja: Yes, but the problem is now that the authorizer fails if the cycle is not published and we try fetching like this
But if we fetch with country iso that is published, all is ok as it should

src/client/pages/Country/hooks/useInitMetaCache.tsx
sorja: It is accessed when loading the Country.tsx
If accessing directly through URL to non-published cycle, it will fail

It is using canView

```
    express.get(ApiEndPoint.MetaData.metaCache(), AuthMiddleware.requireView, getMetaCache)
```

src/client/pages/Country/hooks/useInitSections.tsx
sorja: See above explanation

src/client/store/assessment/actions/getMetaCache.ts
minotogna: right 👍 
does it makes sense to authorise the metadata view also based on single country status? maybe yes?
Let's think for later on if needed any change,

sorja: This is something we should discuss, how much we want to "keep secret" the un-published metadata, e.g. new variables etc? If not at all, do we really need to authorizer it? Same for metacache

You cant use the application anyway without the metadata, right? So if the user has access to application, he should see the metadata

minotogna (approved):

src/client/store/assessment/actions/getMetaCache.ts
minotogna: yes exactly that's what I was wondering as well. 
maybe metadata / metacache should be always public. let's talk later about this later on for a future task in case

PR #4648, opened 2025-05-06
4541 - Minor fixes

sorja:
- [x] Missing param
- [x] no dashboard is available for regions (redirect)
- [x]  in public view, home link should always go to last published cycle
- [x] no countries are shown in home
- [x] remove AT

### DIFF
--- a/src/client/components/AreaSelector/CountryList/CountryList.tsx
+++ b/src/client/components/AreaSelector/CountryList/CountryList.tsx
@@ -8,10 +8,10 @@ import { i18n } from 'i18next'
 import { Areas, CountryIso, Global, Region, RegionCode, RegionGroup } from 'meta/area'
 import { UserRoles } from 'meta/user/userRoles'
 
-import { useCountries, useRegionGroups } from 'client/store/area'
+import { useRegionGroups } from 'client/store/area'
 import { useCycle } from 'client/store/assessment'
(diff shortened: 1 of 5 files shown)


src/client/components/AreaSelector/CountryList/CountryList.tsx
```diff
@@ -35,13 +37,15 @@ const filterRegions = (props: {
...
+  user: User
 }): Array<Region> => {
-  const { i18n, includeRegions, query, regionGroup } = props
+  const { i18n, includeRegions, query, regionGroup, user } = props
 
   // If includeRegions is empty array, include all regions
   if (includeRegions && includeRegions.length > 0 && !includeRegions.includes(regionGroup.name)) return []
 
   return regionGroup.regions.filter(({ regionCode }) => {
+    if (!user && regionCode === RegionCode.AT) return false
```
minotogna: `another issue: we should never show AT region. ever. public and non public view`
 
is it the case ?

sorja: Updated in [e210043](https://github.com/openforis/fra-platform/pull/4648/commits/e210043dee1c01518a10a5f1f4a40639230842cc)

minotogna (approved):

PR #4650, opened 2025-05-06
4601 - Admin csv download - Use current UI language

### LINKED ISSUE #4601: Admin csv download in different language
**Describe the bug**
AP:

When downloading the list of collaborators, the headings and country names were in French regardless of the fact that I was using the English language version. When I changed the language to English in the Admin page, all returned back to the normal behaviour.
 
The issue seems to be that I had another browser window open where I was using the French version of the platform. The language setting was inherited to the download functionality although not indicated in the UI.

**Expected behavior**
The headers should be the same as the browser language .

Proposed solution :
Let's add a header in the http request accepted-language that is the browser language

yaguzmang:


src/client/components/TablePaginated/ExportButton/hooks/useExportUrl.tsx
```diff
@@ -18,6 +19,7 @@ export const useExportUrl = (props: Props): string => {
   const { assessmentName, countryIso, cycleName, sectionName } = useSectionRouteParams()
   const orderBy = useTablePaginatedOrderBy(path)
   const filters = useTablePaginatedFilters(path)
+  const lang = useLanguage()
```
yaguzmang, line 22: From the issue:
>When downloading the list of collaborators, the headings and country names were in French regardless of the fact that I was using the English language version. When I changed the language to English in the Admin page, all returned back to the normal behaviour.
The issue seems to be that I had another browser window open where I was using the French version of the platform. The language setting was inherited to the download functionality although not indicated in the UI.

> Expected behavior
The headers should be the same as the browser language .

> Proposed solution :
Let's add a header in the http request accepted-language that is the browser language

Proposed solution would have the side effect that for downloading with a specific language, the browser settings need to change (e.g UI is in French but browser settings are in English. The download would be in english). 
I take it what was meant was to download using the current UI language.

minotogna: @yaguzmang my mistake in detailing the issue . I meant ui language , not
Browser language .

But, can we pass a header param ?

yaguzmang: Okay 👍 
To note:
In table paginated we just open an external link for the download
```typescript
    <div className="table-paginated-export-button">
      <Link className={className} target="_blank" to={exportUrl}>
        <Icon className="icon-sub icon-white" name="hit-down" />
        CSV
      </Link>
    </div>
```
To be able to add the header, we need to handle the request and download in the same page.

minotogna: Right . Maybe keep it as param then , it's easier . Isn't it ?

yaguzmang: Yes, and we keep consistency with how we download other things, like repository/links (in a new page).
Let me know if the current version with query params needs any changes :+1:

minotogna: Thanks @yaguzmang 
I'm taking a look

minotogna (approved):

PR #4651, opened 2025-05-07
4640 - Kiosk - Remove Youtube suggestions 

### LINKED ISSUE #4640: Remove youtube suggestions from Kiosk

yaguzmang:
https://github.com/user-attachments/assets/1753a53b-46e9-459d-bad0-97591a40c029


src/client/pages/Kiosk/FraProcess/FraProcess.tsx
```diff
@@ -8,7 +8,7 @@ const FraProcess: React.FC = () => {
     <object
       aria-label="FRA process video"
       className="kiosk-content__embedded-object"
-      data={`https://www.youtube.com/embed/${FRA_PROCESS_VIDEO_ID}`}
+      data={`https://www.youtube.com/embed/${FRA_PROCESS_VIDEO_ID}?rel=0`}
```
yaguzmang: Youtube doesn't allow to remove suggestions entirely anymore. However, you can limit the suggestions to videos from the same channel only with `?rel=0`
image

I'm looking into [Youtube IFrame Player API](https://developers.google.com/youtube/iframe_api_reference) to see if I can add an overlay when the video ends. It uses Iframe and requires you load an external script though.

sorja: What about Vimeo?

minotogna: @sorja @yaguzmang or even hosting it on our own s3 !

minotogna: oh wait. this is not hosted and managed by us. we should use the provided YouTube link, we can't do much about it

minotogna (approved):

PR #4653, opened 2025-05-07
4541 - Fix user countries

sorja:


src/client/components/AreaSelector/CountryList/hooks/useCountriesWithoutRole.ts
```diff
@@ -1,23 +1,31 @@
...
+    const isCyclePublished = Cycles.isPublished(cycle)
+    const allCountryISOs = allCountries.map(({ countryIso }) => countryIso)
+
+    if (userCountries && user && !isCyclePublished) {
       return _userCountries
     }
-    if (userCountries && !user) {
-      return allCountries.map(({ countryIso }) => countryIso).filter((countryIso) => !Areas.isAtlantis(countryIso))
+
+    if (userCountries) {
```
minotogna: It's not clear why the check with user countries is done twice (line 21 and 25).
Is it possible to improve this logic removing redundancies ?

src/client/components/AreaSelector/CountryList/hooks/useCountriesWithoutRole.ts
```diff
@@ -1,23 +1,33 @@
...
-      return allCountries.map(({ countryIso }) => countryIso).filter((countryIso) => !Areas.isAtlantis(countryIso))
+
+    // For unpublished cycles, return user's countries if user is logged in
+    if (user && !isCyclePublished) {
+      return _userCountries
     }
 
-    return allCountries.map(({ countryIso }) => countryIso)
-  }, [_userCountries, allCountries, user, userCountries])
+    // For published cycles or no user, return all countries except Atlantis
```
minotogna, line 26: Line 21: if not user countries is true when user is not logged in ? Making this a little bug ?

minotogna: now I see. `useUserCountries` (which is the wrong name) returns an empty array if there's no user or user has no role in the cycle. which makes the check at line 21 useless. can I push the fix ?

sorja: Please, do so!

minotogna (approved):

PR #4655, opened 2025-05-07
Fix/jrc title naming

andyarnell:
Updated typo in title year i.e. from JRC Forest 2000 to 2020.  Also fixed Spanish to reflect the original title (i.e without translation as it's the name of a dataset), and also added the title in the geojson for the other languages. (as they were previously missing)


andyarnell:
Made changes from out dated repo; closing and opening new PR instead

PR #4661, opened 2025-05-08
4605 - FRA Platform – Resetting Report Status(FRA Latest)

### LINKED ISSUE #4605: FRA Platform – Resetting Report Status(FRA Latest)
@minotogna 
**Describe the bug**
All country reports can be freely accessed and edited by NCs, ANCs, and collaborators.
Image

**Solution**
The ideal approach is to set all countries to the 'In review' status. Only for countries that request voluntary updates should the reviewers of that country change the status from 'In review' to 'Not started.' An email (to be prepared by Bastian) should be sent to inform the user(s) that the platform for that country is open for editing.

sorja:
image

### DIFF
--- a/src/test/migrations/steps/20250508102309-step-fra-latest-countries-in-review.ts
+++ b/src/test/migrations/steps/20250508102309-step-fra-latest-countries-in-review.ts
@@ -1,0 +1,26 @@
+import { CountryStatus } from 'meta/area'
+import { AssessmentNames } from 'meta/assessment/assessment'
+
+import { AssessmentController } from 'server/controller/assessment'
+import { CacheController } from 'server/controller/cache'
+import { BaseProtocol, Schemas } from 'server/db'
+
+export default async (client: BaseProtocol) => {
+  const { assessment, cycle } = await AssessmentController.getOneWithCycle({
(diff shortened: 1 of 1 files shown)


src/test/migrations/steps/20250508102309-step-fra-latest-countries-in-review.ts
```diff
@@ -0,0 +1,26 @@
...
+    assessmentName: AssessmentNames.fra,
+    cycleName: 'latest',
+  })
+  const schemaName = Schemas.getNameCycle(assessment, cycle)
+
+  await client.query(
+    `
+  update ${schemaName}.country
+  set status = $1,
+      last_in_${CountryStatus.review} = now()
```
minotogna: Maybe I would set the last in review at the same time the cycle was created . As well as last in edit . They basically all starts from time '0' (which is cycle creation ).
What do you think ?

minotogna: Or thinking better .

Last in review and last in edit should be left to null .

What do you think ?

minotogna (approved):

PR #4666, opened 2025-05-10
4584 - FRA platform - Editor keep table widths

### LINKED ISSUE #4662: Platform - editor - table width
@minotogna @sorja 

**Issue**

Cannot control the table width of the free text box.

See the recording.

yaguzmang:
https://github.com/user-attachments/assets/1e5ba5f0-7dd3-4b6a-8bba-3d1dbf91ca9a

### DIFF
--- a/src/client/components/EditorWYSIWYG/hooks/_processCustomPaste.ts
+++ b/src/client/components/EditorWYSIWYG/hooks/_processCustomPaste.ts
@@ -1,0 +1,45 @@
+import serialize from 'dom-serializer'
+import { DomUtils, parseDocument } from 'htmlparser2'
+import type { IJodit } from 'jodit/esm/types/jodit'
+
+import { tableTags } from './_sanitizer'
+
+export const _processCustomPaste = (_: IJodit, html: string): string => {
+  const dom = parseDocument(html, { lowerCaseAttributeNames: true, lowerCaseTags: true })
+
+  const styledTableNodes = DomUtils.findAll((node) => {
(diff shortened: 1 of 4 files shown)


src/client/components/EditorWYSIWYG/hooks/_processCustomPaste.ts
```diff
@@ -0,0 +1,45 @@
...
+import { tableTags } from './_sanitizer'
+
+export const _processCustomPaste = (_: IJodit, html: string): string => {
+  const dom = parseDocument(html, { lowerCaseAttributeNames: true, lowerCaseTags: true })
+
+  const styledTableNodes = DomUtils.findAll((node) => {
+    return node.type === 'tag' && tableTags.includes(node.name) && typeof node.attribs.style === 'string'
+  }, dom.children)
+
+  const lengthRE = /^\d+(?:\.\d+)?(?:px|em|rem|%)$/
```
sorja: Seems a bit strict, better any number of numbers (with dot), followed by any number of letters?

sorja:
Should we only sanitize on paste? Or should we also sanitize on blur?
What if user pastes something that gets removed onBlur?

src/client/components/EditorWYSIWYG/hooks/_processCustomPaste.ts
minotogna: and let's use the onBlur sanitizer on onPaste as well

minotogna (approved):

PR #4669, opened 2025-05-12
1992 - enableIf logic

sorja:
next pr migration step and video


src/client/pages/Section/DataTable/Table/RowData/Cell/hooks/useEnableIf.ts
```diff
@@ -0,0 +1,60 @@
...
+  row: Row
+  sectionName: string
+  table: Table
+}
+
+export const useEnableIf = (props: Props): boolean => {
+  const { data, col, row, sectionName, table } = props
+  const assessment = useAssessment()
+  const cycle = useCycle()
+  const countryIso = useCountryIso()
```
minotogna: better to use route param

src/tools/generateCache/index.ts
```diff
@@ -26,6 +24,8 @@ const exec = async () => {
       )
     })
   )
+
+  await AssessmentController.generateMetaCache()
```
minotogna: this should be done before generating metadata cache

src/meta/expressionEvaluator/functions/enableIf/valueEquals.ts
```diff
@@ -0,0 +1,18 @@
...
+export const valueEquals: ExpressionFunction<Context> = {
+  name: 'valueEquals',
+  minArity: 2,
+  executor: () => {
+    return (value: string, expectedValue: string): boolean => {
+      if (Objects.isNil(expectedValue)) {
+        return false
+      }
+
+      return value === expectedValue
```
minotogna: is there a real need to add this simple comparison check as function ? 
And, what if we want to enable a cell if the expected value is null ?

src/tools/generateCache/index.ts
sorja: generateMetaCache is using redis repository without force, then it would fetch the old row and/or cols to generate metacache, right? so doesn't it make sense to do it after?

src/meta/expressionEvaluator/functions/enableIf/valueEquals.ts
sorja: Do you mean we should just do in line the function and remove this file?

sorja: or should we introduce `valueEqualsOrNull`?
If value is null, then return true

src/tools/generateCache/index.ts
minotogna: Let me have a look

src/meta/expressionEvaluator/functions/enableIf/valueEquals.ts
minotogna: > Do you mean we should just do in line the function and remove this file?

Yes . It's better .

sorja: Ok, in case we want to show notification or error message later, maybe I will reintroduce
_I will update migration step to check for undef || x === 'yes'_

sorja: @minotogna PR updated

src/tools/generateCache/index.ts
minotogna: I couldn't see how the generateMetaCache uses Redis. that's what I remembered . generating metacache and metadata cache should not use any redis.
Where exactly did u see generate metacache using Redis ?

src/client/pages/Section/DataTable/Table/RowData/Cell/hooks/useEnableIf.ts
```diff
@@ -0,0 +1,61 @@
...
+    const rowCache: RowCache = {
+      ...row,
+      tableName: table.props.name,
+      sectionName,
+    }
+
+    try {
+      return ExpressionEvaluator.evalFormula<boolean>({
+        assessment,
+        assessments: { [assessment.props.name]: assessment },
```
minotogna, line 41: can we pass only assessments and assessmentName ? (omitting assessment ) ?

src/tools/generateCache/index.ts
sorja: I was mistaken, it's not using Redis at all!
(this change was rolled back)

minotogna: Ah ok

src/client/pages/Section/DataTable/Table/RowData/Cell/hooks/useEnableIf.ts
sorja: can we refactor in other PR?

minotogna (approved):

PR #4692, opened 2025-05-15
4205 - Data Explorer UI - Redux setup

yaguzmang:


src/client/store/explorer/state.ts
```diff
@@ -0,0 +1,18 @@
+import { AssessmentName } from 'meta/assessment/assessment'
+import { CycleName } from 'meta/assessment/cycle'
+import { SectionName } from 'meta/assessment/section'
+import { RecordAssessmentData } from 'meta/data'
+import { ExplorerFilter } from 'meta/explorer/filter'
+import { ExplorerMetadata } from 'meta/explorer/metadata'
+
+export interface ExplorerDataState {
```
minotogna: they all need to go to their own state file and without key. the key is given by the combine reducer 

e.g. 
`export type ExplorerDataState = RecordAssessmentData`

and from `src/client/store/explorer/data/state.ts`

src/client/store/explorer/metadata/slice.ts
```diff
@@ -0,0 +1,7 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+export const ExplorerMetadataSlice = createSlice({
+  name: 'explorerMetadata',
+  initialState: {},
+  reducers: {},
```
yaguzmang, line 6: @minotogna do we want to add anything else at this stage?

minotogna: Nope , just the skeleton is fine

src/client/store/explorer/data/slice.ts
```diff
@@ -0,0 +1,7 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+export const ExplorerDataSlice = createSlice({
+  name: 'explorerData',
```
minotogna: data is enough .

If you look in redux debugger the full path is prefixed with explorer already (from the combined reducer )

src/client/store/explorer/filter/slice.ts
```diff
@@ -0,0 +1,7 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+export const ExplorerFilterSlice = createSlice({
+  name: 'explorerFilter',
```
minotogna: Here filter

src/client/store/explorer/metadata/slice.ts
```diff
@@ -0,0 +1,7 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+export const ExplorerMetadataSlice = createSlice({
+  name: 'explorerMetadata',
```
minotogna: And here metadata 😉

src/client/store/explorer/filter/state.ts
```diff
@@ -0,0 +1,5 @@
+import { AssessmentName } from 'meta/assessment/assessment'
+import { CycleName } from 'meta/assessment/cycle'
+import { ExplorerFilter } from 'meta/explorer/filter'
+
+export type ExplorerFilterState = Record<AssessmentName, Record<CycleName, ExplorerFilter>>
```
minotogna: we need the section name as well. filters are per section

minotogna (approved):

PR #4699, opened 2025-05-17
4205 - Data Explorer UI - Countries Selection / Fra variable see definitions link

yaguzmang:
TODO: 
- [x] Filter country options based on Dashboard -> selected filtered countries
- [x] Add definitions link

(UPDATED)

https://github.com/user-attachments/assets/19d481d9-b7cb-461f-bee5-14be82d08615

### DIFF
--- a/src/client/pages/Explorer/Filters/Country/Country.tsx
+++ b/src/client/pages/Explorer/Filters/Country/Country.tsx
@@ -1,13 +1,24 @@
 import React from 'react'
 
+import { CountryIso } from 'meta/area'
+
+import { useAppDispatch } from 'client/store'
+import { ExplorerFilterActions } from 'client/store/explorer/filter/actions'
+import { useExplorerCountries } from 'client/store/explorer/filter/hooks/useExplorerCountries'
+import { useCycleRouteParams } from 'client/hooks/useRouteParams'
 import CountryMultiSelect from 'client/components/CountryMultiSelect'
 
 const Country: React.FC = () => {
-  const handleChange = (value: Array<string>) => {
-    return value
+  const { assessmentName, cycleName } = useCycleRouteParams()
+  const dispatch = useAppDispatch()
+
+  const explorerCountries = useExplorerCountries(assessmentName, cycleName)
+
+  const handleChange = (value: Array<CountryIso>) => {
+    dispatch(ExplorerFilterActions.setCountries({ assessmentName, countries: value, cycleName }))
   }
 
-  return <CountryMultiSelect onChange={handleChange} />
+  return <CountryMultiSelect onChange={handleChange} value={explorerCountries} />
 }
 
 export default Country
--- a/src/client/pages/Explorer/Filters/Country/Country.tsx
+++ b/src/client/pages/Explorer/Filters/Country/Country.tsx
@@ -1,24 +1,31 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 
 import { CountryIso } from 'meta/area'
 
 import { useAppDispatch } from 'client/store'
 import { ExplorerFilterActions } from 'client/store/explorer/filter/actions'
 import { useExplorerCountries } from 'client/store/explorer/filter/hooks/useExplorerCountries'
+import { useHomeCountriesFilter } from 'client/store/ui/home'
 import { useCycleRouteParams } from 'client/hooks/useRouteParams'
 import CountryMultiSelect from 'client/components/CountryMultiSelect'
 
 const Country: React.FC = () => {
   const { assessmentName, cycleName } = useCycleRouteParams()
   const dispatch = useAppDispatch()
+  const homeCountriesFilter = useHomeCountriesFilter()
+  const allowedCountries = homeCountriesFilter?.length > 0 ? homeCountriesFilter : undefined
 
   const explorerCountries = useExplorerCountries(assessmentName, cycleName)
+
+  useEffect(() => {
+    dispatch(ExplorerFilterActions.setCountries({ assessmentName, countries: [], cycleName }))
+  }, [assessmentName, cycleName, dispatch, homeCountriesFilter])
 
   const handleChange = (value: Array<CountryIso>) => {
     dispatch(ExplorerFilterActions.setCountries({ assessmentName, countries: value, cycleName }))
   }
 
-  return <CountryMultiSelect onChange={handleChange} value={explorerCountries} />
+  return <CountryMultiSelect allowedCountries={allowedCountries} onChange={handleChange} value={explorerCountries} />
 }
 
 export default Country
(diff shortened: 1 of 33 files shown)


src/client/store/explorer/filter/state.ts
```diff
@@ -1,6 +1,5 @@
 import { AssessmentName } from 'meta/assessment/assessment'
 import { CycleName } from 'meta/assessment/cycle'
-import { SectionName } from 'meta/assessment/section'
 import { ExplorerFilter } from 'meta/explorer/filter'
 
-export type ExplorerFilterState = Record<AssessmentName, Record<CycleName, Record<SectionName, ExplorerFilter>>>
+export type ExplorerFilterState = Record<AssessmentName, Record<CycleName, ExplorerFilter>>
```
yaguzmang, line 5: ExplorerFilter already has `Record<SectionName, ...`, and the countries are shared across sections:
```typescript
export type ExplorerFilter = {
  countries: Array<CountryIso>
  dimensions: Record<SectionName, Array<DimensionName>>
  measures: Record<SectionName, Array<MeasureName>>
}
```

minotogna: Yes , right .

src/client/store/explorer/filter/slice.ts
```diff
@@ -1,7 +1,31 @@
...
+    ) => {
+      const { assessmentName, countries, cycleName } = action.payload
+
+      Objects.setInPath({
+        obj: state,
+        path: [assessmentName, cycleName, 'countries'],
+        value: countries,
+      })
+    },
+  },
```
sorja: Not sure, just wondering if this should be in a separate file?

yaguzmang: Yes, thanks :+1:

src/client/store/explorer/filter/hooks/useExplorerCountries.ts
```diff
@@ -0,0 +1,13 @@
+import { AssessmentName } from 'meta/assessment/assessment'
+import { CycleName } from 'meta/assessment/cycle'
+import { ExplorerFilter } from 'meta/explorer/filter'
+
+import { ExplorerFilterSelectors } from 'client/store/explorer/filter/selectors'
+import { useAppSelector } from 'client/store/store'
+
+export const useExplorerCountries = (
+  assessmentName: AssessmentName,
```
minotogna: This hook can even take 0 arguments and use internally useCycleRouteParams

src/client/pages/Explorer/Filters/Measure/Measure.tsx
```diff
@@ -1,16 +1,43 @@
...
     return value
   }
 
-  return <MultiSelect onChange={handleChange} options={[]} placeholder={t('common.variable')} />
+  return (
+    <div className="measure-filter-container">
+      <MultiSelect onChange={handleChange} options={[]} placeholder={t('common.variable')} />
+      <button
+        className="btn-definitions-info"
+        onClick={() =>
```
minotogna: It's often a good practice to create the onClick with useCallback

src/client/pages/Explorer/Filters/Measure/Measure.tsx
```diff
@@ -1,16 +1,43 @@
...
+import { ApiEndPoint } from 'meta/api/endpoint'
+
+import { useCycleRouteParams } from 'client/hooks/useRouteParams'
+import Icon from 'client/components/Icon'
 import MultiSelect from 'client/components/Inputs/MultiSelect/MultiSelect'
 
 const Measure: React.FC = () => {
-  const { t } = useTranslation()
+  const {
+    i18n: { language },
```
minotogna: We have a hook useLang , it's preferred to use

src/client/pages/Explorer/Filters/Country/Country.tsx
```diff
@@ -1,13 +1,31 @@
...
+import { useCycleRouteParams } from 'client/hooks/useRouteParams'
 import CountryMultiSelect from 'client/components/CountryMultiSelect'
 
 const Country: React.FC = () => {
-  const handleChange = (value: Array<string>) => {
-    return value
+  const { assessmentName, cycleName } = useCycleRouteParams()
+  const dispatch = useAppDispatch()
+  const homeCountriesFilter = useHomeCountriesFilter()
+  const allowedCountries = homeCountriesFilter?.length > 0 ? homeCountriesFilter : undefined
```
minotogna, line 16: Can allowedCountries be already the result of useHomeCountries ?

minotogna (commented):
👍

src/client/store/explorer/filter/slice.ts
minotogna: Yes @sorja @yaguzmang 
Let's keep in mind in the new store conventions. It will help us to keep the code clean .
The big mega refactor will come afterwards 😅😅🙏

src/client/pages/Explorer/Filters/Measure/Measure.tsx
minotogna: Looking at DefinitionLink component , this is the same function and it might be good to export from client/components/DefinitionLink a hook called useOpenDefinition .

Code repetition should be avoided when possible 😉

src/client/pages/Explorer/Filters/Country/Country.tsx
yaguzmang: homeCountriesFilter is an empty array by default, that's why in that case I make allowedCountries = undefined.

Should I change the logic so that an empty array means no filtering? that way allowedCountries can be the result of useHomeCountries

minotogna: Or change useHomeCountries to handle undefined ? What do you think ?
(Test it with fra 2020 though )

yaguzmang: Updated it for consistency :+1:  but still added an empty array check because clearing the select sets [] in the store.

minotogna: There should be only one way to handle 'empty home country selection'
We shouldn't use undefined and empty array.
I think that home countries should go back to empty array only .

My comment was mainly about how to handle allowedCountries in country select .
I was thinking, if allowedCountries is an empty array , then no countries are allowed to select.
If allowedCountries is undefined, it means there's no restriction , I.e. all countries can be selected.

The prop allowedCountries shouldn't be considered as the selection of home countries that is the result of a select (which makes sense to have an empty array if nothing is selected)

Sorry @yaguzmang , I think it should be back to the way you have initially implemented it 🙏
I didn't have all the elements that you have mentioned now, but I should have known, perdóname señor 😅

yaguzmang: No problem, I can undo latest commit 👍 I didn't realize about the empty array selection until after also

minotogna (approved):

Issue #4670, opened 2025-05-12
Use Crypto.randomUUID over uuidv4


minotogna:
https://medium.com/@matynelawani/uuid-vs-crypto-randomuuid-vs-nanoid-313e18144d8c

sorja:
Some finding:
- Node and Browsers implement `crypto` and it works well and fast
- Problems with `jest` it does not recognize `crypto` and would need to be polyfilled

Developers note:
It's not worth it as long as we keep using jest or jest doesn't implement `crypto` in it's environment. Note: changing environment to `node` doesn't resolve the issue.

Resolution:
Just move uuids from utils to meta and rename `v4` to `getUuid`.
Usage example: `const uuid = UUIDs.getUuid()`

PR #4706, opened 2025-05-20
4670 - Move UUIDs to meta

sorja:
Note:
Keep package uuids

More info in issue comments #4670


src/client/hooks/useToaster.ts
```diff
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 
-import { UUIDs } from 'utils/uuids'
+import { UUIDs } from 'meta/uuids'
```
minotogna: Can we rename the folder uuid ? 
That's how it's called in explorer branch

minotogna (approved):

PR #4711, opened 2025-05-21
4205 - Data Explorer UI - Explorer Metadata cache

yaguzmang:


src/server/repository/measurement/systemOfMeasurement/getOne.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { SystemOfMeasurement, SystemOfMeasurementName } from 'meta/measurement/systemOfMeasurement'
+
+import { BaseProtocol, DB } from 'server/db'
+
+type Props = {
+  systemOfMeasurementName: SystemOfMeasurementName
+}
+
+export const getOne = (props: Props, client: BaseProtocol = DB): Promise<SystemOfMeasurement> => {
```
yaguzmang, line 11: This I had added based on the old comment in the previous issue. It is not used in this PR, however, it's already there for when we need it. Let me know if I should remove it though.

minotogna: Not sure if we will need it , but we can keep it for now.

minotogna (approved):

PR #4714, opened 2025-05-22
4214 - Move static files to S3

### LINKED ISSUE #4214: Move static files to S3
Note:
Should we move static files to S3? Is there an issue for it, if not, should we make?

_Originally posted by @sorja in https://github.com/openforis/fra-platform/pull/4213#pullrequestreview-2550198476_

sorja:
todo/
- assessment/cycle level files
- cycle duplicate-del-rename

done/
- Remove old files
- Fetch files from S3
- separate tool for uploading
- Single end point for static files
- No Migration file required
- No database rows added

File struct 1:1 to what is in tools/uploadStaticFiles/staticFiles

image


.eslintrc.js
```diff
@@ -60,6 +60,7 @@ module.exports = {
     'no-param-reassign': ['error', { props: true, ignorePropertyModificationsFor: ['state', 'acc'] }],
     '@typescript-eslint/no-use-before-define': ['error'],
     '@typescript-eslint/ban-ts-comment': 'off',
+    '@typescript-eslint/array-type': ['error', { default: 'generic' }],
```
sorja, line 63: add array type enforcement (X[] vs Array<X>)

src/meta/file/static.ts
```diff
@@ -0,0 +1,107 @@
+import { ApiEndPoint } from 'meta/api/endpoint'
```
sorja, line 1: Main change: Static file names/types/getters

src/tools/uploadStaticFiles/index.ts
```diff
@@ -0,0 +1,58 @@
+import '../scriptInit'
```
sorja, line 1: Main change : simple uploader of files in given folder (staticFiles)

src/server/api/file/getMultipleS3Files.ts
```diff
@@ -0,0 +1,44 @@
+import * as path from 'path'
```
sorja, line 1: Added support for fetching multiple files, should we omit this change instead? we are not using it currently @minotogna 

(I thought bulk download would need it so I added but on closer look it generated the needed files on the fly 👍 )

minotogna: Leave it @sorja

sorja:
@minotogna Would you like me to delete `public/static` files from S3 dev instance so you can try?

minotogna:
> @minotogna Would you like me to delete `public/static` files from S3 dev instance so you can try?

Yes please @sorja

minotogna (approved):

PR #4715, opened 2025-05-22
4205 - Data Explorer UI - Fetch metadata per section

yaguzmang:
https://github.com/user-attachments/assets/f9d79375-3bc1-4ea4-9fe7-62a5aaf5480c

TODO in a following PR: 
- [ ] Fix multi select disable layout


src/client/store/explorer/metadata/actions/getMetadata.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ApiEndPoint } from 'meta/api/endpoint'
+import { CycleParams } from 'meta/api/request'
+import { ExplorerMetadata } from 'meta/explorer/metadata'
+
+type Params = CycleParams & { sectionNames: Array<string> }
+
+type Returned = Record<string, ExplorerMetadata>
+
+export const getMetadata = createAsyncThunk<Returned, Params>('explorer/metadata/get', async (params) => {
+  const { data } = <{ data: Returned }>await axios.get(ApiEndPoint.Explorer.sectionsMetadata(), {
```
minotogna, line 13: In this case better to add generics to axios.get instead of outside  cast

minotogna (approved):

src/client/store/explorer/metadata/actions/getMetadata.ts
minotogna: Even next pr, this is not important, just to keep in mind

yaguzmang: Okay, will change it to `const { data } = await axios.get<Returned>` next PR 👍

PR #4731, opened 2025-05-26
4205 - Data Explorer UI - Fetch data

yaguzmang:


src/client/store/explorer/data/actions/index.ts
```diff
@@ -0,0 +1,8 @@
+import { ExplorerDataSlice } from 'client/store/explorer/data/slice'
+
+import { getData } from './getData'
+
+export const ExplorerDataActions = {
+  ...ExplorerDataSlice.actions,
```
minotogna, line 6: I think we should start removing the spread of Slice actions. we should always declare the actions separately if needed.
this is something I was thinking today, if we want to make actions independent from slice. we'll talk Friday anyway

you can remove this in a next pr.

what do you think?

src/client/store/explorer/data/hooks/data.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const useGetExplorerSectionData = () => {
+  const dispatch = useAppDispatch()
+
+  const { tableName } = useExplorerSectionMetadata() ?? {}
+  const countryISOs = useExplorerCountries()
+  const dimensions = useExplorerDimensions()
+  const measures = useExplorerMeasures()
+  const { assessmentName, countryIso: _countryIso, cycleName, sectionName } = useSectionRouteParams()
+  const countryIso = _countryIso as CountryIso
```
minotogna, line 32: you can use the type in `SectionRouteParams<CountryIso>` to avoid casting

minotogna (approved):

src/client/store/explorer/data/actions/index.ts
yaguzmang: Sure, I can remove it in a next PR.

So we would maybe do for example
`dispatch(ExplorerFilterSlice.actions.setCountries({...`
instead of 
`dispatch(ExplorerFilterActions.setCountries({....`
?

minotogna: Nope, that's not what I meant .

I meant to just remove the spread of the slice actions . All actions do not need to reference the slice .
They will be created with either create async think or create action

PR #4747, opened 2025-05-28
4205 - Data Explorer UI - forestAreaChange/growingStock

yaguzmang:
image

image


src/meta/measurement/measures/index.ts
```diff
@@ -1,11 +1,16 @@
-import { variableToMeasures } from 'meta/measurement/measure/variablesToMeasures'
+import { TableName } from 'meta/assessment/table'
+import { VariableName } from 'meta/assessment/variable'
+import { MeasureName } from 'meta/measurement/measure/measure'
+import { measureToVariables, variableToMeasures } from 'meta/measurement/measure/variablesToMeasures'
```
minotogna: Maybe we should move these two files under measures folder , since they are internal to Measures and accessed only through Measures ?

src/server/repository/redis/explorer/getManyMetadata.ts
```diff
@@ -1,6 +1,9 @@
+import { systemsOfMeasurement } from 'tools/migrations/public/steps/data/systemsOfMeasurement'
```
minotogna: Repository should never access tools .
Tools should never accessed by the application code.

Move this in a shared folder , maybe under meta ?

yaguzmang: Moved to meta/systemOfMeasurement, thanks 👍

minotogna (approved):

PR #4757, opened 2025-05-28
4205 - Data Explorer UI - Migrate growingStockComposition

yaguzmang:
image

### DIFF
--- a/src/tools/createMeasuresDimensions/index.ts
+++ b/src/tools/createMeasuresDimensions/index.ts
@@ -20,7 +20,7 @@ const client: BaseProtocol = DB
 
 type Props = {
   assessment: Assessment
-  systemOfMeasurementName: SystemOfMeasurementName
+  systemOfMeasurementName: SystemOfMeasurementName | null // null -> no system of measurement
   tableNames: Array<TableName>
 }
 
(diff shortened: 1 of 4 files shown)


src/tools/createMeasuresDimensions/index.ts
```diff
@@ -79,6 +78,21 @@ const _createMeasuresAndDimensionsForTables = async (props: Props) => {
       jsonb_array_elements_text(cn.names) as dim(name)
     where t.props ->> 'name' in ($1:csv)
       and t.props -> 'columnNames' is not null
+    
+    union -- Include columnNames from cellsExportAlways
```
yaguzmang, line 82: @minotogna 
In the previous implementation we had a special case for this in the UI: cellsExportAlways -> most recent year:
image
It wasn't listed in variables or columns.

Will we add something similar to the current implementation? (if so, in a new issue?)
image

mostRecentYear is both a dimension and a measure, so it is listed in both selectors.

src/tools/createMeasuresDimensions/index.ts
```diff
@@ -103,6 +117,12 @@ const createAllMeasuresDimensions = async () => {
     })
   )
 
+  await _createMeasuresAndDimensionsForTables({
+    assessment,
+    systemOfMeasurementName: null,
+    tableNames: ['growingStockComposition2025'],
```
yaguzmang, line 123: Should we add a mapper for dimension names as well? and insert the dimensions in camelCase
"scientific_name" -> "scientificName"
"common_name" -> "commonName"
image

minotogna: Yes @yaguzmang 
I'll create an issue

minotogna (approved):

PR #4768, opened 2025-05-30
4752 - Country last_edit: populate with result of activity_log.time insert

### LINKED ISSUE #4752: Country last_edit: populate with result of activity_log.time insert

sorja:


minotogna (approved):
👍

PR #4861, opened 2025-06-19
4857 - Permissions selector with default values reset

sorja:
- removed useInitialState -> moved defaults to each field
  - SelectField and PermissionField -> use Controller
  - removed useFieldReset hook (infite loop)
  - permissions reset properly when switching roles (shouldUnregister)


src/client/components/Form/hooks/useFieldReset.ts
```diff
@@ -0,0 +1,22 @@
...
+  setValue: UseFormSetValue<any>
+  defaultValues: Record<string, unknown>
+}
+
+export const useFieldReset = ({ defaultValues, fields, setValue, watchValues }: UseFieldResetProps) => {
+  useEffect(() => {
+    fields.forEach((fieldDefinition) => {
+      const { name, shouldShow } = fieldDefinition
+      if (shouldShow && !shouldShow(watchValues)) {
+        setValue(name, defaultValues[name])
```
minotogna: Should the values be reset to undefined ?
Default values gets applied when field is visible ?
In case we'll talk Monday about it

minotogna (approved):

PR #4874, opened 2025-06-25
4857 - Clean Auth

sorja:
- [x] unit test
- [x] check other usages for Collaborator and CollaboratorPermissions
- [x] remove old type
- [x] double check auth
- [x] triple check auth


src/meta/user/authorizer/canEditCycleData.ts
```diff
@@ -0,0 +1,84 @@
+import { Areas, Country, CountryStatus } from 'meta/area'
```
sorja: This file is now "single source of truth" for:

canEditData and canViewReview

Removed all duplicated, and passed optional `allowedStatuses?: Array<CountryStatus>` (this was the missing piece before).

It's possible to rename this better? We can introduce this as common SSOT and have `canEditCycleData` call this function instead


e.g.

```
export const canEditCycleData = (props: Props) => someGenericName(props)
```

src/server/api/metadata/index.ts
```diff
@@ -10,7 +10,7 @@ import { getSectionsMetadata } from './getSectionsMetadata'
 
 export const MetadataApi = {
   init: (express: Express) => {
-    express.get(ApiEndPoint.MetaData.metaCache(), AuthMiddleware.requireView, getMetaCache)
+    express.get(ApiEndPoint.MetaData.metaCache(), getMetaCache)
```
sorja, line 13: This auth is removed:
When accessing table 1a, we call previous cycles meta cache
This fails, if the user has no role in previous cycle

Should be ok

minotogna: Sure , thanks for finding this out

PR #4894, opened 2025-07-08
4872: Refactor User Edit/Invitation forms with factories

minotogna:


src/server/controller/user/update.ts
```diff
@@ -10,26 +9,31 @@ import { UserRepository } from 'server/repository/public/user'
...
+const toUserEditProps = (props: Props): UserEditProps => {
+  const { user, userEditForm } = props
 
-  const userToUpdate: Partial<Omit<User, 'props'> & { props: Partial<UserProps> }> = {
-    id: userEditForm.userId,
-    email: userEditForm.email,
-    props: userProps,
+  // only admin can update the email
+  if (!Users.isAdministrator(user)) {
+    delete userEditForm.user.email
```
sorja, line 26:
```
  // In user repository update:
  return getOne({ email: user.email, allowDisabled: true }, client)
```

src/meta/inputs/select.ts
```diff
@@ -0,0 +1,11 @@
...
+  label: string
+  value: string
+}
+
+export type OptionsGroup = {
+  label?: string
+  options: Array<Option>
+}
+
+export type OptionsOrGroups = ReadonlyArray<Option | OptionsGroup>
```
sorja: Not sure if UI types should be under meta? What do you think?

minotogna: They were imported by form fields factories and the server build fails .

minotogna: We either move back the factories under components or we move all these meta

src/server/controller/user/update.ts
minotogna: We need a way to prevent non admins to edit the email . I'll take a look later, but maybe we can use the id

src/meta/inputs/select.ts
minotogna: restored the hooks. pr is ready

src/server/controller/user/update.ts
minotogna: fixed in commit. pr is ready

sorja (approved):

PR #4900, opened 2025-07-10
4872 - User Role props

sorja:


src/server/controller/user/update.ts
```diff
@@ -29,30 +35,54 @@ const toUserEditProps = (props: Props): UserEditProps => {
...
+
+  await ActivityLogRepository.insertActivityLog({ activityLog }, client)
+}
+
+export const update = async (props: Props, client: BaseProtocol = DB): Promise<User> => {
+  const { cycle } = props
+
+  return client.tx(async (t) => {
+    const updatedUser = await _updateUser(props, t)
+    await _updateRole(props, t)
```
sorja: Separate functions for updateuser, updateRole

src/client/pages/User/hooks/useEditUserRules.ts
```diff
@@ -41,10 +43,10 @@ export const useEditUserRules = (props: Props): EditUserRules => {
     // still loading targetUser
     if (Objects.isNil(targetUser)) return rules
 
-    if (administrator) rules.emailDisabled = false
+    if (isAdministrator) rules.emailDisabled = false
 
-    if ((administrator || editingSelf) && countryPage) rules.rolePropsAvailable = true
+    if ((isAdministrator || isSelf) && isCountryPage /* && !isReviewer */) rules.rolePropsAvailable = true
```
minotogna: I think role props available should be explicit by role.
not if is not reviewer, but if it's nc,anc,c

src/client/pages/User/hooks/useFormDefinition.ts
```diff
@@ -27,6 +32,12 @@ export const useFormDefinition = (props: Props): FormDefinition => {
         label: '',
         defaultValue: targetUser?.id,
       },
+      {
```
minotogna: should we just use the uuid instead of the id ?

src/client/pages/User/hooks/useFormDefinition.ts
```diff
@@ -66,10 +77,99 @@ export const useFormDefinition = (props: Props): FormDefinition => {
       },
     ]
 
-    // if (rolePropsAvailable) {
-    //   fields.push(...roleFields)
-    // }
+    if (rolePropsAvailable) {
+      const role = Users.getRole(targetUser, countryIso, cycle) as UserRoleExtended<RoleName>
+
+      const roleFields: Array<FieldDefinition> = [
```
minotogna: better a separate hook ? and called useRolePropsFields ? ( we will also have a role name field)

src/server/controller/user/update.ts
```diff
@@ -29,30 +35,54 @@ const toUserEditProps = (props: Props): UserEditProps => {
...
-    const message = ActivityLogMessage.userUpdate
-    const params = { activityLog: { target, section: 'users', message, user } }
-    await ActivityLogRepository.insertActivityLog(params, t)
+  const target = { roles: [updatedRole], userUuid: userEditForm.user.uuid }
+  const message = ActivityLogMessage.userRolesUpdate
+  const activityLog = { target, section: 'users', message, user }
+
+  await ActivityLogRepository.insertActivityLog({ activityLog }, client)
+}
+
```
minotogna: this is a bit unreadable. maybe inner separate functions ?

src/server/repository/public/userRole/getOne.ts
```diff
@@ -0,0 +1,20 @@
+import { RoleName, UserRole } from 'meta/user'
+
+import { BaseProtocol, DB } from 'server/db'
+import { UserRoleAdapter } from 'server/repository/adapter/userRole'
+
+type Props = {
+  uuid: string
+}
+
+export const getOne = async (props: Props, client: BaseProtocol = DB): Promise<UserRole<RoleName> | null> => {
```
minotogna: why do we need this if get user already has the roles ?

src/client/pages/User/hooks/useEditUserRules.ts
```diff
@@ -41,10 +47,20 @@ export const useEditUserRules = (props: Props): EditUserRules => {
     // still loading targetUser
     if (Objects.isNil(targetUser)) return rules
 
-    if (administrator) rules.emailDisabled = false
+    if (isAdministrator) rules.emailDisabled = false
 
-    if ((administrator || editingSelf) && countryPage) rules.rolePropsAvailable = true
+    const rolePropsRoles =
+      isAdministrator && isAlternateNationalCorrespondent && isCollaborator && isNationalCorrespondent
```
minotogna: Maybe an or ? A user can have only one role

minotogna: Please also review the rule .

The props are available if target user is nc/anc/c and user who is editing is admin or is himself

src/server/controller/user/update/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { updateUser } from 'server/controller/user/update/updateUser'
+import { BaseProtocol, DB } from 'server/db'
+import { UserRepository } from 'server/repository/public/user'
+
+export const update = async (props: Props, client: BaseProtocol = DB): Promise<User> => {
+  const { cycle } = props
+
+  return client.tx(async (t) => {
+    const updatedUser = await updateUser(props, t)
+    await updateRole({ ...props, targetUser: updatedUser }, t)
```
minotogna, line 14: Should this be called conditionally ?

sorja: I must have removed it...

image


Is it preferred here to keep the main function clean or in the main function?

minotogna: The if  in the caller is better  .
Call the method if the role is available and
Let's check if we have a method Authorizer.canEditUserRoleProps ?

sorja: We have only `Authorizer.canEditUserRole` => for role type editing

src/client/pages/User/hooks/useEditUserRules.ts
sorja:
```

    const targetUserHasRoleProps =
      isTargetAlternateNationalCorrespondent || isTargetCollaborator || isTargetNationalCorrespondent

    if ((isAdministrator || isSelf) && targetUserHasRoleProps && isCountryPage) rules.rolePropsAvailable = true

```

src/server/controller/user/update/index.ts
minotogna: Can u add it ?

minotogna: And use it also in the client edit user rules ?

src/client/pages/User/hooks/useEditUserRules.ts
```diff
@@ -41,10 +42,11 @@ export const useEditUserRules = (props: Props): EditUserRules => {
     // still loading targetUser
     if (Objects.isNil(targetUser)) return rules
 
-    if (administrator) rules.emailDisabled = false
+    if (isAdministrator) rules.emailDisabled = false
 
-    if ((administrator || editingSelf) && countryPage) rules.rolePropsAvailable = true
+    if (Authorizer.canEditUserRoleProps({ user, countryIso, cycle, target: targetUser }))
```
minotogna, line 47: the countryPage check should still exists. this is shared with edit user admin page.
```
if(countryPage && Authorizer.canEditUserRoleProps)...
```

sorja: image

so i move it from here? now the check is in `canEditUserRoleProps`

minotogna: that's fine then .

minotogna (approved):

PR #4905, opened 2025-07-11
4895 - Data Explorer UI - Side panel

yaguzmang:
https://github.com/user-attachments/assets/a8419bc3-49b6-47c9-9149-9ef750b11a9d

image

Measures units selector are to be added on top in the next PR:

image

should we add titles for each blck? (e.g Select variable units, Select axes )


src/client/pages/Explorer/Filters/Options/Options.tsx
```diff
@@ -0,0 +1,54 @@
...
+const Options: React.FC = () => {
+  const { t } = useTranslation()
+  const [opened, setOpened] = useState(false)
+
+  const { applyAxisSelection, axisSelection, resetAxisSelection, toggleAxis } = useToggleAxis()
+
+  const openPanel = () => setOpened(true)
+  const closePanel = () => setOpened(false)
+
+  const handleCancel = () => {
```
minotogna: When passing a function to an inner component , better to use useCallback

src/client/pages/Explorer/Filters/Options/Options.tsx
```diff
@@ -0,0 +1,54 @@
...
+    resetAxisSelection()
+    closePanel()
+  }
+
+  const handleApply = () => {
```
minotogna: Same here

src/i18n/resources/en/common.json
```diff
@@ -51,6 +51,7 @@
   "login": "Reserved area",
   "name": "Name",
   "noItemsFound": "No items found",
+  "options": "Options",
```
minotogna: Should we use a gear icon and no label ? 
If you agree , 
If we have the icon , replace it
If we don't have it , use another icon and I'll find it later

sorja:
Should we add headers?

`Select Axis` or `Axis`

and

`Select unit` or `Unit`

Looks a bit naked in the first video

But it looks really good otherwise, nice job 👍

minotogna (approved):

PR #4915, opened 2025-07-15
4872: User Edit Form: Admin roles UI

minotogna:
**User Edit Form: Added Admin roles UI only**

**Some changes**

- [x] Handled isDisabled, resetIf, triggers, disabled options through watches in FormField
  - FormField component now accepts a renderInput function to render input fields (instead of children) with results from watches, isolating the values updates. 
- [x] Added FormField checkbox 
- [x] Added Admin roles validations / reset values / disabled values 
- [x] Handled custom error field to show only one error when validating roles 
- [x] Fixes to CountryMultiSelect 

https://github.com/user-attachments/assets/fea7ca53-972b-45ab-b3e0-23d668108e4a


sorja:
should we omit `Countries` text placeholder?

minotogna:
> should we omit `Countries` text placeholder?

Not sure , it actually makes it clear that user can select countries next to the role . But if you think it should be removed , we can remove it .

sorja (approved):

PR #4922, opened 2025-07-22
4872: User Edit Form: Persist user roles (admin)

minotogna:


src/server/controller/user/update/updateRoles.ts
```diff
@@ -0,0 +1,28 @@
...
+
+type UpdateRoleProps = Props & {
+  targetUser: User
+}
+
+export const updateRoles = async (props: UpdateRoleProps, client: BaseProtocol): Promise<void> => {
+  const { assessment, cycle, targetUser, user, userEditForm } = props
+  const { roles } = userEditForm
+
+  if (!roles || !Users.isAdministrator(user) || String(user.id) === String(targetUser.id)) return
```
sorja, line 18: Why checking for Admin here? Can't NC also update user roles?

src/server/repository/public/userRole/updateRoles.ts
```diff
@@ -0,0 +1,97 @@
...
+    return client.map<UserRole>(
+      `insert into users_role (user_uuid, role)
+       values ($1, $2)
+       returning *`,
+      [userUuid, RoleName.ADMINISTRATOR],
+      UserRoleAdapter
+    )
+  }
+
+  delete roles.ADMINISTRATOR
```
sorja, line 41: Is this delete required? Above we check & return

src/server/controller/user/update/updateRoles.ts
minotogna: because only admin can update the roles

sorja (approved):

PR #4934, opened 2025-08-05
4816 - Store Refactor: Geo->Map 

yaguzmang:


src/client/store/geo/map/actions/index.ts
```diff
@@ -0,0 +1,5 @@
+import { GeoMapSlice } from 'client/store/geo/map/slice'
+
+export const GeoMapActions = {
+  ...GeoMapSlice.actions,
```
sorja: This notation is not used anymore (`...[Slice].actions`)

src/client/store/store.ts
```diff
@@ -14,6 +16,9 @@ export interface LazyLoadedSlices {
   admin: {
     links: LinksState
   }
+  [GeoSliceName]: {
```
sorja, line 22: would keep same as others

src/client/store/geo/map/hooks/map.ts
```diff
@@ -0,0 +1,7 @@
+import { GeoMapSelectors } from 'client/store/geo/map/selectors'
+import { GeoMapOptions } from 'client/store/geo/map/state'
+import { useAppSelector } from 'client/store/hooks'
+
+export const useGeoMapOptions = (): Partial<GeoMapOptions> => {
+  return useAppSelector((state) => GeoMapSelectors.getOptions(state))
```
minotogna: for simplicity, this can even become ```useAppSelector(GeoMapSelectors.getOptions)```
But it's irrelevant. just to mention.

src/client/store/store.ts
minotogna: actually I think this is good. we should have a const for all slice names. 
and, I think they should not be exported from the slice root, but from slice/name.ts 
What do you think @sorja @yaguzmang . we can talk about this later in the meeting

src/client/store/geo/map/actions/index.ts
yaguzmang: So now we create an action for each reducer? 
in this case for setOptions
then we will have 
-setOptions (action)
and
-setOptionsReducer
?

src/client/store/store.ts
```diff
@@ -22,8 +22,8 @@ export interface LazyLoadedSlices {
   [GeoSliceName]: {
     map: GeoMapState
   }
-  [LoginSlice.name]: LoginState
-  [TablePaginatedSlice.name]: TablePaginatedState
+  [LoginSliceName]: LoginState
+  [TablePaginatedSliceName]: TablePaginatedState
```
yaguzmang, line 26: Error 1 of 1

it wasn't possible to just use < Slice >.name due to the issue above, so I added the slice names in name.ts.
Those new name.ts files will be moved in the next task where I will add all slice folders.

minotogna: Slice should be a folder . And then name.ts should include the name 😉

yaguzmang: I added the folder for geo. For login and table paginated, should I do it in the next task we talked about?

src/client/store/geo/map/actions/index.ts
minotogna: yes . to an action corresponds a reducer.

src/client/store/store.ts
minotogna: yes, there is the task we created together yesterday

minotogna (approved):

PR #4942, opened 2025-08-08
4816 - Store Refactor: Geo->Mosaic 

yaguzmang:


src/client/store/geo/mosaic/slice/extraReducers/applyOptionsReducer.ts
```diff
@@ -0,0 +1,13 @@
+import { ActionReducerMapBuilder } from '@reduxjs/toolkit'
+
+import { applyOptions } from 'client/store/geo/mosaic/actions/applyOptions'
+import { MosaicState } from 'client/store/geo/mosaic/state'
+import { LayerFetchStatus } from 'client/store/ui/geo/stateType'
+
+export const applyOptionsReducer = (builder: ActionReducerMapBuilder<MosaicState>) => {
+  builder.addCase(applyOptions, (state) => {
+    state.options = { ...state.ui }
```
sorja: why `state.options = state.ui`?

src/client/store/geo/mosaic/slice/extraReducers/setUiOptionsReducer.ts
```diff
@@ -0,0 +1,19 @@
+import { ActionReducerMapBuilder } from '@reduxjs/toolkit'
+
+import { setUiOptions } from 'client/store/geo/mosaic/actions/setUiOptions'
+import { MosaicState } from 'client/store/geo/mosaic/state'
+
+export const setUiOptionsReducer = (builder: ActionReducerMapBuilder<MosaicState>) => {
+  builder.addCase(setUiOptions, (state, action) => {
+    const { datum } = action.payload
+    const newUiOptions = {
```
sorja: should we use objects.setInPath?

src/client/store/geo/mosaic/slice/extraReducers/applyOptionsReducer.ts
yaguzmang: https://github.com/user-attachments/assets/dc8607de-f913-4a73-ae6b-bebf6cecf600

In the Mosaic section we edit the options and then apply them so that we don't make unnecessary calls to the backend while setting up. In this reducer we apply what's on the ui to the options used to fetch the layer.
Maybe I should call it `state.uiOptions`?

sorja: Is it possible to avoid duplication?

E.g. Throttle or only send when clicking Apply?

Not sure if relevant, let me know

yaguzmang: ~Yes it's possible to avoid duplication~. It will take some time to adapt but I think it's worth it for consistency and maintainability long term 👍  Will update soon

yaguzmang: Turns out we do need to keep track of the options used to fetch the layer so that the 'apply' button can be disabled when the options have not changed. Also, to know whether or not the layer needs to be fetched when toggling on. 

However, I'm now saving it at `state.urlTemplateData.requestOptions`, this way it makes more sense. It represents the options that were used to request the url that is currently in the state. Let me know what you think.

sorja: Sounds better, thanks

src/client/components/Navigation/NavGeo/SatelliteMosaic/MosaicControl/MosaicControl.tsx
```diff
@@ -11,58 +12,66 @@ import SelectPrimary from 'client/components/Inputs/SelectPrimary'
...
+        value={mosaicOptions.maxCloudCoverage}
       />
 
       <ButtonCheckbox
-        checked={uiMosaicOptions.snowMasking}
+        checked={mosaicOptions.snowMasking}
         className="geo-options-grid__one-col"
         label={t('geo.snowMasking')}
-        onClick={() => dispatch(GeoActions.setMosaicSnowMasking(!uiMosaicOptions.snowMasking))}
+        onClick={() => dispatch(MosaicActions.setOption({ key: 'snowMasking', value: !mosaicOptions.snowMasking }))}
```
sorja, line 67: I noticed now we use string here, should we introduce ENUM? Would it make sense in these cases? You can do it in next PR also

sorja (approved):

src/client/components/Navigation/NavGeo/SatelliteMosaic/MosaicControl/MosaicControl.tsx
yaguzmang: currently it is:
```
export interface MosaicOptions {
  maxCloudCoverage: number
  snowMasking: boolean
  sources: Partial<Record<MosaicSource, boolean>>
  year: number
}
```
Then in the action:
```
type Params<K extends keyof MosaicOptions = keyof MosaicOptions> = {
  key: K
  value: MosaicOptions[K]
}

export const setOption = createAction<Params>('geo/mosaic/options/set')
```
 So we get autocomplete and errors if not using a valid key
 
image

image

Let me know if you still think it's necessary and will add it in next PR 👍

PR #4951, opened 2025-08-14
3701 - Re-order the NWFP table by Value (1000 local currency)

### LINKED ISSUE #3701: Re-order the NWFP table by Value (1000 local currency)
@minotogna 

For Table 7, it would be great if a new tab(function) could be added here that allows the user to reorder the table by **Value (1000 local currency)** from big to small value.

**Screenshots**
image

sorja:
# Table sorting

https://github.com/user-attachments/assets/331cd997-ca7b-4896-a1eb-ce48915b8ded

Todo:
- [x] UI
- [x] Migration step
- [x] Find better icons

This pr allows sorting _some rows_ in the table.

E.g. we don't want to sort by all columns
E.g. we don't want to sort all rows when sorting (e.g. calculated totals, subtitles, placeholders etc should stay static)

Adding this metadata allows us to have full control
Example metadata:

```
// col

{
...
  "labels": {
    "66817a08-dc93-4151-b5ed-176d8f04e9b7": {
      "key": "nonWoodForestProductsRemovals.quantity"
    },
    "66da2217-da42-492f-9ff4-c99a59e6675c": {
      "key": "nonWoodForestProductsRemovals.quantity"
    },
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501": {
      "key": "nonWoodForestProductsRemovals.quantity"
    }
  },
  "colType": "header",
  "sortable": {
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501": "quantity"
  }
}


// row

{
  "type": "data",
  "index": 6,
  "cycles": [ ... ],
  "sortableBy": {
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501": [
      "quantity",
      "value"
    ]
  },
  "validateFns": {. ..  },
  "variableName": "product_7"
}

```


minotogna:
@sorja should we handle only col sorting prop ?

src/client/pages/Section/DataTable/Table/GridHeadCell/ButtonSort/ButtonSort.scss
```diff
@@ -0,0 +1,6 @@
+.grid-header-cell__button-sort {
+  position: absolute;
+  right: 0px;
```
minotogna, line 6: Should we handle rtl as well ?

src/client/pages/Section/DataTable/Table/GridHeadCell/ButtonSort/ButtonSort.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const sortColName = col.props.sortable?.[cycle.uuid]
+
+  const isCurrentlySorted = sortState?.colName === sortColName
+  const sortOrder = isCurrentlySorted ? sortState.order : SortOrder.NONE
+
+  if (!sortColName) return null
+
+  let label = '↕'
+  if (sortOrder === SortOrder.ASC) label = '↑'
+  if (sortOrder === SortOrder.DESC) label = '↓'
```
minotogna: Can u check icomoon to find a nice icon ? (Or I'll check next week)

src/client/pages/Section/DataTable/Table/hooks/useTableSorting/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  DESC = 'desc',
+  NONE = 'none',
+}
+
+export interface SortState {
+  colName: ColName | null
+  order: SortOrder
+}
+
+export interface Props {
```
minotogna: Should we call this PropsSort ?

src/meta/assessment/col/index.ts
```diff
@@ -62,6 +62,7 @@ export interface ColProps {
   linkedNodes?: Record<CycleUuid, ColLinkedNode>
   readonly?: boolean
   select?: Record<CycleUuid, ColSelectProps>
+  sortable?: Record<CycleUuid, ColName>
```
minotogna: Should this be record cycleuuid -> boolean ? 
colName is already a prop .

sorja: We cant - headers don't have props:

This caused some problems in the inital implementation

I think we should add colNames to headers, I don't know if it makes sense though:


For example, a header col meta for table Non wood forest products removals and value:

```
{
  "index": 5,
  "style": {
    "66817a08-dc93-4151-b5ed-176d8f04e9b7": {
      "colSpan": 1,
      "rowSpan": 1
    },
    "66da2217-da42-492f-9ff4-c99a59e6675c": {
      "colSpan": 1,
      "rowSpan": 1
    },
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501": {
      "colSpan": 1,
      "rowSpan": 1
    }
  },
  "cycles": [
    "66817a08-dc93-4151-b5ed-176d8f04e9b7",
    "66da2217-da42-492f-9ff4-c99a59e6675c",
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501"
  ],
  "labels": {
    "66817a08-dc93-4151-b5ed-176d8f04e9b7": {
      "key": "nonWoodForestProductsRemovals.value"
    },
    "66da2217-da42-492f-9ff4-c99a59e6675c": {
      "key": "nonWoodForestProductsRemovals.value"
    },
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501": {
      "key": "nonWoodForestProductsRemovals.value"
    }
  },
  "colType": "header",
  "sortable": {
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501": "value"
  }
}
```

As you can see, the only way to distinquish this column is by the added colName.

Let me know what you suggest

___Note:__ (Initial implementation was with boolean and rows having col.uuid as relation instead of array of col names)_

sorja:
> @sorja should we handle only col sorting prop ?

Do you mean removing the sortableBy prop from `row`?

This would result in rather "funny" looking tables, when we can't control which rows we want to sort.

We want to keep these in place, right?

kuva

Here for example, without the row.props.sortableBy

kuva

This is an example of a simple table

Here is an example of more complex table (maybe for the future..), which would benefit from the `sortableBy`

kuva

minotogna:
> > @sorja should we handle only col sorting prop ?
> 
> 
> 
> Do you mean removing the sortableBy prop from `row`?
> 
> 
> 
> This would result in rather "funny" looking tables, when we can't control which rows we want to sort.
> 
> 
> 
> We want to keep these in place, right?
> 
> 
> 
> kuva
> 
> 
> 
> Here for example, without the row.props.sortableBy
> 
> 
> 
> kuva
> 
> 
> 
> This is an example of a simple table
> 
> 
> 
> Here is an example of more complex table (maybe for the future..), which would benefit from the `sortableBy`
> 
> 
> 
> kuva
> 
> 

@sorja i see now .
My comment was because it is better to centralize in one place the sorting props . Then maybe let's add the sort by prop to the table directly (where we could specify the cols and rows affected )? 
What do you think ?

sorja:
> > > @sorja should we handle only col sorting prop ?
> > 
> > 
> > Do you mean removing the sortableBy prop from `row`?
> > This would result in rather "funny" looking tables, when we can't control which rows we want to sort.
> > We want to keep these in place, right?
> > kuva
> > Here for example, without the row.props.sortableBy
> > kuva
> > This is an example of a simple table
> > Here is an example of more complex table (maybe for the future..), which would benefit from the `sortableBy`
> > kuva
> 
> @sorja i see now . My comment was because it is better to centralize in one place the sorting props . Then maybe let's add the sort by prop to the table directly (where we could specify the cols and rows affected )? What do you think ?

What would the object look in this case?
```
table.props.sort = {

cycleUuid: {


    [colUuid_1]: { rows(Affected): Array<rowUuid>, colName: colName } ,
    [colUuid_2]: { rows(rowsAffected): Array<rowUuid>, colName: colName } 

  }

}

```

Or similar?

So idea:

When rendering the table head cell, we check if table.props.sort[cycleuuid][col_uuid] contains the current col.uuid?
to get rows affected, then we need to traverse with _colUuid_ (which we dont have, unless persisted in sorting state..)


I agree that single place would be better, but I think the benefit is not that huge over row x col

I need to think about it, let me know if you had something on your mind

minotogna:
> > > > @sorja should we handle only col sorting prop ?
> > > 
> > > 
> > > Do you mean removing the sortableBy prop from `row`?
> > > This would result in rather "funny" looking tables, when we can't control which rows we want to sort.
> > > We want to keep these in place, right?
> > > kuva
> > > Here for example, without the row.props.sortableBy
> > > kuva
> > > This is an example of a simple table
> > > Here is an example of more complex table (maybe for the future..), which would benefit from the `sortableBy`
> > > kuva
> > 
> > 
> > @sorja i see now . My comment was because it is better to centralize in one place the sorting props . Then maybe let's add the sort by prop to the table directly (where we could specify the cols and rows affected )? What do you think ?
> 
> What would the object look in this case?
> 
> ```
> table.props.sort = {
> 
> cycleUuid: {
> 
> 
>     [colUuid_1]: { rows(Affected): Array<rowUuid>, colName: colName } ,
>     [colUuid_2]: { rows(rowsAffected): Array<rowUuid>, colName: colName } 
> 
>   }
> 
> }
> ```
> 
> Or similar?
> 
> So idea:
> 
> When rendering the table head cell, we check if table.props.sort[cycleuuid][col_uuid] contains the current col.uuid? to get rows affected, then we need to traverse with _colUuid_ (which we dont have, unless persisted in sorting state..)
> 
> I agree that single place would be better, but I think the benefit is not that huge over row x col
> 
> I need to think about it, let me know if you had something on your mind

@sorja is it possible that one column can sort certain rows and another column other rows?
Probably not. what do you think?
In this case I would simply structure the prop as :

```typescript
sort: {
  columnNames:['colName1','colName2'],
  rowNames:['rowName1','rowName2']
}
```
I used names, because I think it's easier. what do you think?

src/client/pages/Section/DataTable/Table/GridHeadCell/ButtonSort/ButtonSort.tsx
minotogna: @sorja we already have sort icons, used also in table paginated.
Should we use the same paradigm ?
image
image

sorja: I will use them, if needed we can find better icons later

sorja: https://github.com/user-attachments/assets/76ae5657-9893-425c-8397-5504857af95e

Looks pretty good now :)

sorja:
> > > > > @sorja should we handle only col sorting prop ?
> > > > 
> > > > 
> > > > Do you mean removing the sortableBy prop from `row`?
> > > > This would result in rather "funny" looking tables, when we can't control which rows we want to sort.
> > > > We want to keep these in place, right?
> > > > kuva
> > > > Here for example, without the row.props.sortableBy
> > > > kuva
> > > > This is an example of a simple table
> > > > Here is an example of more complex table (maybe for the future..), which would benefit from the `sortableBy`
> > > > kuva
> > > 
> > > 
> > > @sorja i see now . My comment was because it is better to centralize in one place the sorting props . Then maybe let's add the sort by prop to the table directly (where we could specify the cols and rows affected )? What do you think ?
> > 
> > 
> > What would the object look in this case?
> > ```
> > table.props.sort = {
> > 
> > cycleUuid: {
> > 
> > 
> >     [colUuid_1]: { rows(Affected): Array<rowUuid>, colName: colName } ,
> >     [colUuid_2]: { rows(rowsAffected): Array<rowUuid>, colName: colName } 
> > 
> >   }
> > 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or similar?
> > So idea:
> > When rendering the table head cell, we check if table.props.sort[cycleuuid][col_uuid] contains the current col.uuid? to get rows affected, then we need to traverse with _colUuid_ (which we dont have, unless persisted in sorting state..)
> > I agree that single place would be better, but I think the benefit is not that huge over row x col
> > I need to think about it, let me know if you had something on your mind
> 
> @sorja is it possible that one column can sort certain rows and another column other rows? Probably not. what do you think? In this case I would simply structure the prop as :
> 
> ```ts
> sort: {
>   columnNames:['colName1','colName2'],
>   rowNames:['rowName1','rowName2']
> }
> ```
> 
> I used names, because I think it's easier. what do you think?

I am implemeting this but it's adding unnecessary complexity (which maybe not so obvious)

-> Sorting rows required col name
-> Col name is not available in header cols
-> Circumvent this using colUuid
-> fetching table data to sort rows requires col name

Still investigating, just a heads up that it's not probably as straightforward as we thought and it might be better (for now?) to keep col x props.

Let me know

sorja:
@yaguzmang If you have the chance, can you also give some ideas?

Problem is moving the sorting metadata to table.props

We need colName to know which column in the rows are sorted and for sorting (e.g. getRecordAssessmentData requires colName)

ColHeader have no colName, migrating all would seem quite hard, not every header even have a colName?

Let me know if you have some ideas

I would also prefer to have them in table.


--

Another solution is to construct a more complex object in the table, including the colName

In my local test, I constructed the table.props.sort object using what we have - instead of colName I used ColUUID -> but again, with this we cannot sort.

So, maybe a solution would construct complex object e.g.

```

table.props.sort = {
rowNames: Array<RowName> // this was ok initially
cols: Array<{ 
    colUuid: UUID // this we have already in DB - used for checking whether we render the sorting button for this cell
    colName: string // this we need to hard code in migration to in our case 'value' - this is used for sorting
}>
```

Thanks

yaguzmang:
> @yaguzmang If you have the chance, can you also give some ideas?

@sorja 
Just an idea about the sortable rows: since most rows are sortable and the ones that aren't are kind of special cases, should we assume all rows are sortable and only indicate in the DB the ones that aren't? then a prop like `"fixed" : boolean` in the Row object would make sense. Or, we do it in the table data, we add only the row names that shouldn't be sorted.

About the ColNames in Table for sorting:

```typescript
export interface TableProps {
  cellsExportAlways?: TableCellNames
  columnNames: TableColumnNames
  columnsExport?: TableColumnNames
  columnsExportAlways?: TableColumnNames
  dataExport: boolean
  ...
```
Since we already have props like columnsExport, columnsExportAlways, and so on.. which are col names, I agree we should add the sort cols prop to Table.

I was checking and some header cells already have the name 
image

So taking that into account, I would say we should add the name to the col headers of tables that we want to add sorting to, incrementally, little by little when needed (probably it's not possible to do programmatically?). Since that simplifies everything maybe it's worth doing it that way over time?

sorja:
@yaguzmang Thank you for the insight

On closer inspection we use _either_ `colName` or `labels` when rendering headers (there are no cols with both).
```
select * from assessment_fra.col where col.props ->> 'colType' = 'header' and col.props ->> 'colName' is not null and col.props ->> 'labels' is not null
```



We use them here, but if there is no other usages it should be easy to change this to prioritise labels:

```
const getLabel = (props: { cycle: Cycle; col: Col; t: TFunction }): string => {
  const { col, cycle, t } = props
  return col.props.colName ?? Labels.getCycleLabel({ cycle, labels: col.props.labels, t })
}
```

I will this issue rest for a bit and think

I suspect now that we need to migrate table:nonWoodForestProductsRemovals colNames to header cells

src/tools/migrations/steps/steps/20250818075114-step-sortable-nonWoodForestProductsRemovals-value.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const _updateTable = async (props: BaseProps, client: BaseProtocol): Promise<void> => {
+  const { assessment, cycle } = props
+  const table = await TableRedisRepository.getOne({ assessment, cycle, tableName })
+  const tableProps = { sort: { [cycle.uuid]: sortObject } }
+  await TableRepository.update({ assessment, tableId: table.id, tableProps }, client)
+}
+
+export default async (client: BaseProtocol): Promise<void> => {
+  const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName }, client)
```
minotogna: @sorja this should be applied to 2025 and latest fra cycles.

src/tools/migrations/steps/steps/20250818075114-step-sortable-nonWoodForestProductsRemovals-value.ts
```diff
@@ -0,0 +1,75 @@
...
+  assessment: Assessment
+  cycles: Array<Cycle>
+}
+
+const _updateCol = async (props: BaseProps, client: BaseProtocol): Promise<void> => {
+  const { assessment, cycles } = props
+
+  const schema = Schemas.getName(assessment)
+  const { id: colId } = await client.one(
+    `select id from ${schema}.col where props -> 'labels' -> '${cycles.at(0).uuid}' ->> 'key' = '${label}' `
```
minotogna, line 36: Getting by label is not the best approach. It's safer to use colType on union with index . Anyway now I approve since that label is unique .
Please resolve the comment when you read this

minotogna (approved):

PR #4958, opened 2025-08-20
4932 - Form improvements: Phone input field with prefix picker

### LINKED ISSUE #4932: Form improvements: Phone input field with prefix picker
- [ ] Phone input field with prefix picker

sorja:
https://github.com/user-attachments/assets/39818ed2-b0ed-4c79-99c4-fff1e49ae66d


sorja:
@minotogna If you're curious:

Here is sneak peek on the telephone field I am working on
I tried different ways, but I think this looks pretty neat and sits well in the form

It's only using 2 small packages, some helper to convert ISO3 -> ISO2 and to get the calling code (only supports ISO2)
Later on if we decide to add ISO2 to country prop, we can remove the other package :)

Note: The telephone number is only used for viewing, so it shouldn't matter in what format we save it in database (free text)

minotogna:
> @minotogna If you're curious:
> 
> 
> 
> Here is sneak peek on the telephone field I am working on
> 
> I tried different ways, but I think this looks pretty neat and sits well in the form
> 
> 
> 
> It's only using 2 small packages, some helper to convert ISO3 -> ISO2 and to get the calling code (only supports ISO2)
> 
> Later on if we decide to add ISO2 to country prop, we can remove the other package :)
> 
> 
> 
> Note: The telephone number is only used for viewing, so it shouldn't matter in what format we save it in database (free text)

That's nice @sorja 👍
Regarding iso 2, do u suggest to save it in the db ?

sorja:
> > @minotogna If you're curious:
> > Here is sneak peek on the telephone field I am working on
> > I tried different ways, but I think this looks pretty neat and sits well in the form
> > It's only using 2 small packages, some helper to convert ISO3 -> ISO2 and to get the calling code (only supports ISO2)
> > Later on if we decide to add ISO2 to country prop, we can remove the other package :)
> > Note: The telephone number is only used for viewing, so it shouldn't matter in what format we save it in database (free text)
> 
> That's nice @sorja 👍 Regarding iso 2, do u suggest to save it in the db ?

We can do that too, but maybe not now if we only use it here?
I  might be mistaken - but it seems that ISO2 is more commonly used?

minotogna:
> > > @minotogna If you're curious:
> 
> > > Here is sneak peek on the telephone field I am working on
> 
> > > I tried different ways, but I think this looks pretty neat and sits well in the form
> 
> > > It's only using 2 small packages, some helper to convert ISO3 -> ISO2 and to get the calling code (only supports ISO2)
> 
> > > Later on if we decide to add ISO2 to country prop, we can remove the other package :)
> 
> > > Note: The telephone number is only used for viewing, so it shouldn't matter in what format we save it in database (free text)
> 
> > 
> 
> > That's nice @sorja 👍 Regarding iso 2, do u suggest to save it in the db ?
> 
> 
> 
> We can do that too, but maybe not now if we only use it here?
> 
> I  might be mistaken - but it seems that ISO2 is more commonly used?

@sorja i honestly don't know what is more used . However, if it doesn't cost too much, we can add the iso 2 code in the db now . That would be a better solution . What do you think ?

sorja:
> > > > @minotogna If you're curious:
> > 
> > 
> > > > Here is sneak peek on the telephone field I am working on
> > 
> > 
> > > > I tried different ways, but I think this looks pretty neat and sits well in the form
> > 
> > 
> > > > It's only using 2 small packages, some helper to convert ISO3 -> ISO2 and to get the calling code (only supports ISO2)
> > 
> > 
> > > > Later on if we decide to add ISO2 to country prop, we can remove the other package :)
> > 
> > 
> > > > Note: The telephone number is only used for viewing, so it shouldn't matter in what format we save it in database (free text)
> > 
> > 
> > > 
> > 
> > 
> > > That's nice @sorja 👍 Regarding iso 2, do u suggest to save it in the db ?
> > 
> > 
> > We can do that too, but maybe not now if we only use it here?
> > I  might be mistaken - but it seems that ISO2 is more commonly used?
> 
> @sorja i honestly don't know what is more used . However, if it doesn't cost too much, we can add the iso 2 code in the db now . That would be a better solution . What do you think ?

I started working on this and migrated the iso2 to DB. Everything worked well. I added a csv in migrations/data of format: [iso3][iso2]

After migration I realized, that we don't need the iso2 but we need the _calling code_.

So I left the csv with iso2, but added calling code: [iso3][iso2][callingcode] - so if we want the iso2 in the future, it's easy to migrate.

Now that I have calling code, I think I can get rid of both NPM packages.

Confusing message, but in short: instead of iso2, I added calling code. I left iso2 for future in case

Let me know

minotogna:
@sorja in short, let's add iso 2 and calling code in db , ok ?

sorja:
@minotogna Video updated - let me know

Developers note:
image

This was slightly more complicated to get to work nicely
Tried different combinations, but I think the current video looks the best - just need to refine a bit how we show the two fields next to each other (e.g. separator `-`, spacing, width etc)

I also simplified the telephone number vlaidation requirements: I noticed some internal users (e.g. @ FAO ) use internal phone numbers (e.g. 061234) - I don't think there is a real reason to strictly validate the numbers, they are only for reading anyway.

I divided the component in meaningful parts:

The main component `TelephoneField` is divided in two components:
 - CallingCode (more complex logic)
 - PhoneNumberInput (simple text input)

The parent handles state for calling code and phone number


The complex component `CallingCode`:
- Contains a selector and text input
- Has some logic to glue everything together
- Contains the subcomponents:
  - Option (used to display the flag, country name and calling code in drop down)
  - FlagIcon (used by Option and CallingCode)

sorja:
@minotogna PR Description updated with latest video

Note after last video/changes: 
Added vertical divider after calling code
Aligned text vertically to center

src/client/components/Inputs/Select/types.ts
```diff
@@ -6,19 +7,25 @@ export type Option = {
   value: string
 }
 
+export type OptionElement = {
```
minotogna: This seems a duplicate of Option . Can u use React.Node for label Option ?

src/client/pages/User/hooks/useRoleValidationSchema.ts
```diff
@@ -42,12 +42,13 @@ export const useRoleValidationSchema = () => {
             .or(z.literal('')),
           primaryPhoneNumber: z
             .string()
-            .min(1, { error: t('form.errors.required', { field: t('editUser.primaryPhoneNumber') }) })
-            .refine((val) => z.e164().safeParse(val).success, {
-              error: t('form.errors.invalid', { field: t('editUser.primaryPhoneNumber') }),
-            }),
+            .min(1, { error: t('form.errors.required', { field: t('editUser.primaryPhoneNumber') }) }),
+          // .refine((val) => z.e164().safeParse(val).success, {
```
minotogna: Should we remove the comments ?

src/tools/migrations/steps/steps/data/iso3-iso2-calling_code.csv
```diff
@@ -0,0 +1,257 @@
+iso3,iso2,calling_code
```
minotogna: Should this be part of the migration step directly so that after reset it gets removed ?

src/client/pages/User/hooks/useRoleValidationSchema.ts
sorja: Sure, left in case we want to enable in future, but I can remove

sorja:
In addition, removed Atlantis countries from calling code selector

src/client/components/Inputs/Select/types.ts
```diff
@@ -2,7 +2,7 @@ import { ActionMeta, Props as ReactSelectProps } from 'react-select'
 import { CreatableProps } from 'react-select/creatable'
 
 export type Option = {
-  label: string
+  label: string | React.ReactNode
```
minotogna: string can be removed

minotogna (approved):

PR #4964, opened 2025-08-25
4933 - Data Sources: Align different col styles

### LINKED ISSUE #4933: NationalData->DataSources convert Year for DataSource as MultiSelect + custom option
migrate current options, if not a year, as creatable option

sorja:
Resolves #4933 

Reading:

image

Writing:

image

image

### DIFF
--- a/src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/DataSources.scss
+++ b/src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/DataSources.scss
@@ -17,3 +17,17 @@
     text-transform: uppercase;
   }
 }
+
+.data-source div.multiselect__container div.select__control {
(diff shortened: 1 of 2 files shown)


src/client/pages/Section/Descriptions/NationalDataDescriptions/DataSources/DataSources.scss
```diff
@@ -17,3 +17,19 @@
     text-transform: uppercase;
   }
 }
+
+.data-source div.multiselect__container div.select__control,
+.data-source div.select__control {
```
sorja, line 22: Didn't find better selector, seem like .multiselect__container is needed to make it work for the multiselect. Omitting this doesn't work

Should be ok for this case

minotogna (approved):

PR #4968, opened 2025-08-29
4534 - Cache assessments

sorja:
Note: Incremental PRs - target branch feature/issue branch

Example log
```
10:17:06 info: fra - with 3: 2020, 2025, latest
10:17:06 info: panEuropean - with 2: 2020, 2025
```

`assessment` map visible in Redis:
kuva

with following data:

kuva

...where notable change is new props `indexesByName` and `indexesByUuid`

```
  "indexesByName": {
    "2020": 0,
    "2025": 1,
    "latest": 2
  },
  "indexesByUuid": {
    "66817a08-dc93-4151-b5ed-176d8f04e9b7": 0,
    "66da2217-da42-492f-9ff4-c99a59e6675c": 1,
    "9fa1fd86-b4e3-46e4-bf48-125f082ea501": 2
  }
```

Next up: Meta cache map

### DIFF
--- a/src/meta/assessment/assessment/index.ts
+++ b/src/meta/assessment/assessment/index.ts
@@ -1,4 +1,4 @@
-import { Cycle, CycleUuid } from 'meta/assessment/cycle'
+import { Cycle, CycleName, CycleUuid } from 'meta/assessment/cycle'
 import { AssessmentMetaCache } from 'meta/assessment/metaCache'
 
 export type AssessmentName = string
@@ -15,11 +15,13 @@ export type AssessmentProps = {
 }
 
 export interface Assessment {
-  id: number
-  uuid: string
+  cycleIndexesByName: Record<CycleName, number>
+  cycleIndexesByUuid: Record<CycleUuid, number>
   cycles: Array<Cycle>
-  props: AssessmentProps
+  id: number
   metaCache?: Record<CycleUuid, AssessmentMetaCache>
+  props: AssessmentProps
+  uuid: string
 }
 
 export type RecordAssessments = Record<AssessmentName, Assessment>
--- a/src/server/repository/redis/assessment/index.ts
+++ b/src/server/repository/redis/assessment/index.ts
@@ -1,0 +1,5 @@
+import { getAssessmentMap } from 'server/repository/redis/assessment/getAssessmentMap'
+
+export const AssessmentRedisRepository = {
+  getAssessmentMap,
+}
--- a/src/server/repository/redis/keys.ts
+++ b/src/server/repository/redis/keys.ts
@@ -56,11 +56,6 @@ type PropsCountry = PropsCycle & {
   countryIso: CountryIso
 }
 
-// Identity function
-export const getKeyId = ({ key }: { key: Key }): Key => {
-  return key
-}
-
 export const getKeyAssessment = (props: PropsAssessment): string => {
   const { assessment, key } = props
   return `${key}:${assessment.props.name}`
--- a/src/server/repository/redis/keys.ts
+++ b/src/server/repository/redis/keys.ts
@@ -1,15 +1,30 @@
 import { CountryIso } from 'meta/area'
-import { Assessment } from 'meta/assessment/assessment'
+import { Assessment as AssessmentType } from 'meta/assessment/assessment'
 import { Cycle } from 'meta/assessment/cycle'
 
 // ===== Keys
 
+enum Area {
+  country = 'area:country',
+  regionGroups = 'area:regionGroups',
+}
+
+enum Assessment {
+  assessment = 'assessment',
+}
+
 enum Data {
   data = 'data',
 }
+
+enum Explorer {
+  metadata = 'explorer:metadata',
+}
+
 enum Row {
   row = 'row',
 }
+
 enum Section {
   sections = 'sections',
   sectionsIndex = 'sectionsIndex',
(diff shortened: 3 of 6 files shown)


src/server/repository/redis/keys.ts
```diff
@@ -47,6 +56,11 @@ type PropsCountry = PropsCycle & {
   countryIso: CountryIso
 }
 
+// Identity function
+export const getKeyId = ({ key }: { key: Key }): Key => {
```
sorja: Added ID function for consistency, would it be better to call to remove it?

```
const key = getKeyId({ key: Keys.Assessment.assessment })
```

src/meta/assessment/assessment/index.ts
```diff
@@ -15,11 +15,13 @@ export type AssessmentProps = {
 }
 
 export interface Assessment {
```
minotogna, line 17: Should we create an extension on Assessment ?
For example 
AssessmentBase (db) 
Assessment ( cached once)
What do you think or do you have any other suggestions ?

sorja: For DB we can have - we often have them in Adapters, right?

Maybe we can introduce `meta/db/Assessment` etc?

minotogna: Well there are 3 objects : 
1 from db (with lower case props name)
1 result from db repository 
1 from cache 

Any suggestion?

src/server/repository/redis/keys.ts
minotogna: Where is it used ?

sorja: in cacheAssessments in Redis repositories:

```
// file: src/server/repository/redis/assessment/getAssessmentMap.ts

const _cacheAssessments = async (client: BaseProtocol): Promise<AssessmentMap> => {
  const assessments = await AssessmentController.getAll({}, client)
 
  ...

  const key = getKeyId({ key: Keys.Assessment.assessment }) // line 42
  await _setCache({ key, assessmentMap })
  // TODO: Next: Add meta cache generation to './generateMetacCache'

  return assessmentMap
}
```

src/meta/assessment/assessment/index.ts
sorja: | object | path | type name |
|---|---|--|
| 1 from db (with lower case props name) | `meta/{db,postgres}/assessment` | `DBAssessment or AssessmentDB` |
| 1 result from db repository | `meta/assessment/assessment` | `AssessmentBase` |
| 1 from cache | `meta/assessment/assessment` | `Assessment` |


Let me know

minotogna: Ok , so you agree with my first proposal 😉

minotogna: The db objects can be stored in the src/server/ folder? But let's not worry about those now .
What do you think ?

sorja: Sure, next pr ok with types? If I change now I suspect a lot of changes

minotogna: Sure 👍

src/server/repository/redis/keys.ts
minotogna: Wasn't there a utility already ?

sorja: Yes - I think I wasn't clear though:

```
  const key = getKeyId({ key: Keys.Assessment.assessment }) // line 42
```

The getKeyId is identity function which returns the passed key params without altering it
I added it to keep the code consistent (-> fetching redis keys always from same file called `keys.ts`)

We can omit the function and deviate from consistency instead:

```
const key = Keys.Assessment.assessment
```

Let me know

src/server/repository/redis/assessment/index.ts
```diff
@@ -0,0 +1,5 @@
+import { getAssessmentMap } from 'server/repository/redis/assessment/getAssessmentMap'
+
+export const AssessmentRedisRepository = {
+  getAssessmentMap,
```
minotogna, line 4: For next pr , make it plural .
getAssessmentsMap

src/server/repository/redis/keys.ts
minotogna: Not sure I follow .

minotogna: Add the specific key method . I don't see how this method is generic

sorja: Ok, then it should be ok like this and I resolve this comment.

This PR can be approved if nothing else

minotogna: Sorry @sorja 
What key of does it return ?

sorja: `assessment`

minotogna: Can u call it for what it returns ? Like the other methods ? What do you think ?

minotogna: I see . And I wouldn't pass the key . Just return the key assessments .

sorja: kuva

minotogna: Is it used only for the 'assessments' redis entry ? Then it can be called getKeyAssessments with no arguments since it's used to return that key. Isn't it ?

minotogna (approved):

PR #4970, opened 2025-09-01
4534 - Types: AssessmentBase

sorja:
Next PR: RedisRepository.getAssessment and relevant types

### DIFF
--- a/src/server/controller/assessment/create.ts
+++ b/src/server/controller/assessment/create.ts
@@ -1,33 +1,23 @@
 import { ActivityLogMessage } from 'meta/assessment/activityLog'
-import { Assessment } from 'meta/assessment/assessment'
+import { Assessment, AssessmentBase } from 'meta/assessment/assessment'
 import { User } from 'meta/user'
 
 import { BaseProtocol, DB } from 'server/db'
 import { AssessmentRepository } from 'server/repository/assessment/assessment'
 import { ActivityLogRepository } from 'server/repository/public/activityLog'
 
-export const create = async (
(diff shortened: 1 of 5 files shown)


src/server/controller/assessment/create.ts
```diff
@@ -1,33 +1,23 @@
...
 import { AssessmentRepository } from 'server/repository/assessment/assessment'
 import { ActivityLogRepository } from 'server/repository/public/activityLog'
 
-export const create = async (
-  props: { user: User; assessment: Pick<Assessment, 'props'> },
-  client: BaseProtocol = DB
-): Promise<Assessment> => {
+type Props = { user: User; assessment: Pick<Assessment, 'props'> }
+
+export const create = async (props: Props, client: BaseProtocol = DB): Promise<AssessmentBase> => {
```
minotogna, line 11: what about getOne and getAll ?

sorja: They are widely used and there is optional prop meta cache

Removing the meta cache would require updating all calls to the function

I would suggest doing this incrementally, after we have `Redis.getSingleAssessment` in place and used

What do you think?

I only updated createAssessment, because it was the only one with single use

minotogna: ok sure. but call the redis repo method getAssessment or getOne

sorja: Of course

minotogna:
can please check why the tests-all is expected but it doesn't run ? should we just run all tests to all branches ?

minotogna (approved):

PR #4974, opened 2025-09-03
4534 - Fix tests

sorja:


src/server/controller/assessment/index.ts
```diff
@@ -17,6 +17,10 @@ export const AssessmentController = {
   create,
   getAll: AssessmentRepository.getAll,
   getOne: AssessmentRepository.getOne,
+  /**
+   * @deprecated
+   */
+  getOneWithCycleDeprecated: getOneWithCycle,
```
minotogna, line 23: this is not the right approach @sorja 
let's not add all this unnecessary code. 
what is the problem with tests ?

src/tools/generateCache/index.ts
```diff
@@ -9,7 +11,11 @@ import { Logger } from 'server/utils/logger'
 const exec = async (): Promise<void> => {
   await RedisData.getInstance().flushall()
 
-  await CacheController.generateAssessments()
+  await Promise.all(
+    [AssessmentNames.fra, AssessmentNames.panEuropean].map(
+      (assessmentName): Promise<Assessment> => CacheController.generateAssessment({ assessmentName })
+    )
+  )
```
minotogna, line 18: now that I think about , we can restore `await CacheController.generateAssessments()` that calls the getAssessmentsMap

src/tools/cycles/clone.ts
```diff
@@ -21,7 +21,7 @@ const assessmentName = AssessmentNames.fra
 const cycleName = '2025'
 const cycleNameClone = 'latest'
 
-const _addColumns = async (props: { assessment: Assessment; cycle: Cycle }): Promise<void> => {
+const _addColumns = async (props: { assessment: AssessmentBase; cycle: Cycle }): Promise<void> => {
```
minotogna, line 24: I think we should still pass the Assessment, taken from api context or external tools
in the majorities (if not all) of the repositories updated in this PR.

sorja: Assessment is extension of AssessmentBase so we can call them with either, but if we use Assessment we cannot call without first going through Redis

sorja: I think repositories shouldn't know about Redis?

src/tools/generateCache/index.ts
minotogna: I added it in my pr

src/tools/cycles/clone.ts
minotogna: they don't . they shouldn't even know the difference between AssessmentBase and Assessment. that's why we pass assessment and we know we get it from redis, not the repositories.

PR #4978, opened 2025-09-03
4534 - Migration step

sorja:


src/tools/migrations/steps/steps/20250903121613-step-cache-assessments.ts
```diff
@@ -0,0 +1,8 @@
+import { CacheController } from 'server/controller/cache'
+import { BaseProtocol, DB } from 'server/db'
+
```
minotogna, line 3: Initial ddl is missing ?

sorja: Weird, it didn't add it to commit, fixed now, thanks

minotogna (approved):

PR #4987, opened 2025-09-08
4979 - Dashboard 2025

### LINKED ISSUE #4979: Fra 2025 dashboard
- [ ] crate tool to Import data from csv (converted from OJ's worksheet )
- [ ] Add dashboard metadata and show UI elements
  - [x] OJ: Show only columns found in excel

sorja:


src/client/pages/CountryHome/FraHome/hooks/useSections.tsx
```diff
@@ -27,7 +26,8 @@ export const useSections = (): Array<CountryHomeSection> => {
 
     if (!cycle) return null
     const isCountry = Areas.isISOCountry(countryIso)
-    const showOverview = Cycles.isPublished(cycle) || Areas.isISOCountry(countryIso)
+    const showRegionDashboard = !Areas.isISOCountry(countryIso) && cycle.props.dashboard.region
```
minotogna: should we use `cycle.props.dashboard?.region`  . since it's an optional prop ?

src/tools/csvDashboardDataImport/index.ts
```diff
@@ -0,0 +1,140 @@
...
+  await DB.tx(async (client) => {
+    await _writeDb(values, client)
+
+    // Update total land area values
+    await _totalLandArea(client)
+
+    // Update data cache to display total land area updates around the platform (e.g. Table 1A)
+    Logger.info('Starting data cache update')
+    const { assessment, cycle } = await AssessmentController.getOneWithCycle({ assessmentName, cycleName }, client)
+    await CacheController.generateData({ assessment, cycle, force: true }, client)
```
minotogna: should we update the cache only for involved table and should we update the dependants if any ?

minotogna: totalLandArea has two dependants and they must be updated 
image

src/tools/csvDashboardDataImport/index.ts
```diff
@@ -0,0 +1,140 @@
...
+const processCSVFiles = async (): Promise<void> => {
+  // Inset new values
+  const fileNames = _getFileNames().filter((n) => n !== totalLandAreaFile)
+  const values = await _getValues(fileNames)
+
+  await DB.tx(async (client) => {
+    await _writeDb(values, client)
+
+    // Update total land area values
+    await _totalLandArea(client)
```
minotogna, line 133: I would suggest to use `src/tools/migrations/steps/steps/utils/updateTotalLandArea.ts`

minotogna: I think the script mentioned above doesn't include dependents update. maybe we should add it

src/tools/csvDashboardDataImport/index.ts
sorja: Call the new updateDependants

PR #5007, opened 2025-09-11
4997 - Edit Admin: At least one role is required error

### LINKED ISSUE #4997: Edit Admin: At least one role is required error
When editing an admin, changing the appellation twice gives an error `At least one role is required error ` which should not happen

sorja:


src/client/pages/User/hooks/useRolesValidationSchema.ts
```diff
@@ -5,18 +5,24 @@ import { z, ZodAny } from 'zod'
...
 export const useRolesValidationSchema = (): ZodAny => {
   const { t } = useTranslation()
 
   return useMemo<ZodAny>(() => {
     return z.any().check((ctx: ParsePayload) => {
       const values = ctx.value as UserEditFormRoles
+
       const emptyRoles = Object.values(values).every((value) => {
         if (Array.isArray(value)) return value.length <= 0
         return value !== true
```
minotogna: what about here ? the admin is already included in the every iteration. maybe the problem is there ?

minotogna: we could even change the iteration with entries and instead of checking by Array.isArray, we could check by entry name (which is the role name).

minotogna (approved):

src/client/pages/User/hooks/useRolesValidationSchema.ts
```diff
@@ -1,25 +1,42 @@
...
+// Roles with one or more countries
+const countryRoles = [
+  RoleName.REVIEWER,
+  RoleName.NATIONAL_CORRESPONDENT,
+  RoleName.ALTERNATE_NATIONAL_CORRESPONDENT,
+  RoleName.COLLABORATOR,
+  RoleName.VIEWER,
+]
+
+const roleSchema = z.object({
```
sorja, line 18: this looks like

```

z.object({
 [RoleName.ADMINISTRATOR]: z.boolean().optional()
 [RoleName.REVIEWER]: z.array(z.string()).optional(),
  ...,
  
})
```

minotogna:
@Mergifyio refresh

minotogna:
@mergifyio requeue

PR #5011, opened 2025-09-12
4995 - Add Collaborator - Duplicate/Disabled user Errors

### LINKED ISSUE #4995: Add Collaborator - Duplicate/Disabled user Errors
- [ ] Inviting users already with a country role to the same country the first time says: user added, the second time gives error
  -  In both cases BE should reject and give error: `The operation cannot be completed because the invited user already has a role for this country.`

- [ ]  When inviting a disabled user now backend gives error (duplicate key value...)

Image

Be should show error : `The operation cannot be completed because the invited user is disabled.`

sorja:


src/i18n/resources/en/other.json
```diff
@@ -251,6 +251,10 @@
     "review": {
       "commentDeleteNotOwner": "Error: User {{user}} tried to delete a comment that doesn’t own",
       "commentEnterResolvedIssue": "Error: User {{user}} tried to enter a comment for a resolved issue"
+    },
+    "request": {
```
minotogna: Should we use userInvitation as key ?

sorja: sure

minotogna (approved):

PR #5021, opened 2025-09-16
5020 - Update Forest Kids url

### LINKED ISSUE #5020: Kiosk - Update Forest Kids URL

yaguzmang:


src/client/pages/Kiosk/ForestKids/ForestKids.tsx
```diff
@@ -6,7 +6,7 @@ const ForestKids: React.FC = () => {
     <object
       aria-label="Forest Kids video game"
       className="kiosk-content__embedded-object"
-      data="https://www.fao.org/forest-resources-assessment/ForestKids/"
+      data="https://forestkids.herokuapp.com"
```
minotogna, line 9: @sorja has the url changed ? 
Or to be safe we should use the Heroku url ?

sorja: Seems that the FAO one is limited in size, I would use the Heroku one as changed in this PR.

👍

yaguzmang: @minotogna the previous url now has a security policy that doesn't allow to embed the page:
image

It also had these banners and titles
image

so better just link the game directly and it takes the entire space:

image

minotogna (approved):

PR #5029, opened 2025-09-18
5017 - Admin Links - Show locations with links

### LINKED ISSUE #5017: Admin Links: Show Locations with links
In Admin links page, in each row, next to the number indicating the number of locations, add a button show/hide.

- When clicking show, the row expands and below the number of locations show the list of locations.
each location should also link in a new tab.
- When location are shown, the button shows the label hide, that hides the locations

Example:

Image

yaguzmang:
https://github.com/user-attachments/assets/2e58c393-54fa-4d42-9ce0-aeab33e5ed48


src/tools/migrations/steps/steps/20250918030750-step-delete-all-links.ts
```diff
@@ -0,0 +1,15 @@
...
+import { AssessmentController } from 'server/controller/assessment'
+import { BaseProtocol, Schemas } from 'server/db'
+
+export default async (client: BaseProtocol): Promise<void> => {
+  const assessments = await AssessmentController.getAll({}, client)
+
+  await Promises.each(assessments, (assessment) =>
+    Promises.each(assessment.cycles, async (cycle) => {
+      const schemaCycle = Schemas.getNameCycle(assessment, cycle)
+      await client.query(`truncate table ${schemaCycle}.link`)
```
yaguzmang, line 12: Has the feature been used yet for checking/approving links? if not, we can delete them if any present and they will be recalculated once the verify job runs again. 
If yes, then I need to re do this function and calculate the url for each link here. Leaving PR in draft in case.

minotogna: yes, you can delete current links .

src/client/pages/AdminLinks/Locations/Locations.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <div className="locations">
+          {locations?.map((location, idx) => (
+            <Link
+              key={`${String(idx)}-${location.id}`}
+              className="locations-link"
+              rel="noreferrer"
+              target="_blank"
+              to={String(location.url)}
+            >
+              • {getLabel(location)}
```
minotogna: Here we can use an ol tag to show the number .
Also we could adjust the grid template columns to make the last column wider to avoid layout shift

src/meta/assessment/description/descriptions.ts
```diff
@@ -0,0 +1,32 @@
...
+      return 'description.estimationAndForecasting'
+    case CommentableDescriptionName.introductoryText:
+      return 'contactPersons.introductoryText'
+    default:
+      return ''
+  }
+}
+
+export const Descriptions = {
+  getI18nCommentableDescriptionLabelKey,
```
minotogna: I would shorten the name .
getLabelKey since the object is already called Descriptions

src/meta/cycleData/links/link.ts
```diff
@@ -1,6 +1,25 @@
...
+  colName: string
+  id: number
+  url: string
+}
+
+type DescriptionsLocation = LinkLocationBase & {
+  descriptionName: string
+  path: Array<string>
+  sectionName: string
+  tableName: 'descriptions'
```
minotogna: Do we need table name ? 

I think sectionName can be either the section name or originalDataPoint (in case of odp ) 
What do you think ?

src/meta/cycleData/links/link.ts
```diff
@@ -1,6 +1,25 @@
...
 
-export type LinkLocation = Record<string, Array<string> | number | string>
+type LinkLocationBase = {
+  colName: string
+  id: number
+  url: string
+}
+
+type DescriptionsLocation = LinkLocationBase & {
+  descriptionName: string
```
minotogna, line 10: Isn't this needed in odp as well ?
I think there we have comments for extent of forest , comments for forest characteristics, and references 
Links can be used in 3 different places , but please double check this

yaguzmang: For odp I see this one place for references links:
image
Maybe I'm missing something?

And the comments seem to be shared between  extent of forest and forest characteristics:
image

image

So for that example this is the links table:
image

To choose the label key for odp I use the colName:
```typescript
const descriptionLabelKey = colName === 'description' ? 'dataSource.comments' : 'nationalDataPoint.dataSources'
```
Should I keep it that way?

minotogna: I remember wrongly . comments is only one then.

However, users can insert links in Comments and References, in two places.

Using colName is a bit misleading. you have the custom type for ODP, use odpSection maybe ? or another name ? any suggestion ?

yaguzmang: Added it as odpSection :+1:

minotogna (approved):

PR #5039, opened 2025-09-19
4815 - Clean up nodeValues/hooks/data.ts dot dot dot notation -> use set in path

### LINKED ISSUE #4815: Clean up src/client/store/data/tableData/nodeValues/hooks/data.ts dot dot dot notation -> use set in path

sorja:
https://github.com/user-attachments/assets/f345b1c5-027a-49b5-b5b2-fb4f5f395890

### DIFF
--- a/src/client/store/data/tableData/nodeValues/hooks/data.ts
+++ b/src/client/store/data/tableData/nodeValues/hooks/data.ts
@@ -43,17 +43,11 @@ export const useRecordAssessmentDataWithOdp = (): RecordAssessmentData => {
       return recordTableData
     }
 
-    return {
-      [assessmentName]: {
-        ...(data[assessmentName] || {}),
-        [cycleName]: {
-          ...(data[assessmentName][cycleName] || {}),
-          [countryIso]: {
-            ...(data[assessmentName][cycleName][countryIso] || {}),
-            ...tableNames.reduce<RecordTableData>(reducer, {}),
-          },
-        },
-      },
-    }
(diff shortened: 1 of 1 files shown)


src/client/store/data/tableData/nodeValues/hooks/data.ts
```diff
@@ -43,17 +43,11 @@ export const useRecordAssessmentDataWithOdp = (): RecordAssessmentData => {
...
-            ...(data[assessmentName][cycleName][countryIso] || {}),
-            ...tableNames.reduce<RecordTableData>(reducer, {}),
-          },
-        },
-      },
-    }
+    const path = [assessmentName, cycleName, countryIso]
+    const newTableData = Objects.setInPath({ obj: {}, path, value: tableNames.reduce<RecordTableData>(reducer, {}) })
+
+    const tableData = Objects.cloneDeep(data)
```
minotogna, line 49: Would a simple spread work as well ? Cloning deep is too much ?

sorja: `"extentOfForest" is read-only`when spread notation: the object comes from state and is not mutable by setInPath

Needs to be Objects.cloneDeep unfortunately

If you don't think this PR brings value, it can be closed

What do you think?

minotogna: I think it can be closed since it's a double iteration rather than once as before

PR #5059, opened 2025-09-26
4005 - Admin countries sort by name - Add country/region labels

yaguzmang:
TODO next prs:
- [x] Add labels to assessment_cycle country/regions -> Update cache
- [x] Add migration step that calls the new tool

image


src/tools/addCountryRegionLabels/index.ts
```diff
@@ -0,0 +1,111 @@
...
+import { ToolsUtils } from 'tools/utils/toolsUtils'
+
+import { CountryIso, RegionCode } from 'meta/area'
+import { Lang, LanguageCodes } from 'meta/lang'
+
+import { BaseProtocol, DB } from 'server/db'
+
+const pgp = pgPromise()
+
+const _getPublicCountryIsos = async (client: BaseProtocol = DB): Promise<Array<CountryIso>> => {
```
sorja: is possible to use redis in these?

src/tools/addCountryRegionLabels/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    const labels = {} as Record<Lang, string>
+    LanguageCodes.forEach((lang) => {
+      labels[lang] = i18nInstances[lang].t(`area.${regionCode}.listName`)
+    })
+    return { labels, regionCode }
+  })
+
+  await DB.tx(async (client) => {
+    await _updateCountryTable(countryLabels, client)
+    await _updateRegionTable(regionLabels, client)
```
minotogna: Afterwards we should call CacheController to update redis cache .

Also , I don't see the migration step ?

src/tools/addCountryRegionLabels/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  return client.map(`select country_iso from public.country`, [], (row) => row.country_iso)
+}
+
+const _getPublicRegionCodes = async (client: BaseProtocol = DB): Promise<Array<RegionCode>> => {
+  return client.map(`select region_code from public.region`, [], (row) => row.region_code)
+}
+
+const _updateDDL = async (): Promise<void> => {
+  await Promise.all([
+    DB.none(`alter table public.country add column if not exists labels jsonb not null default '{}'::jsonb`),
```
minotogna: This should go to the migration step and it has to be added to the generate schema as well

src/tools/addCountryRegionLabels/index.ts
minotogna: Yes , just call AreaController

yaguzmang: Area controller gets the countries based on assessment/cycle. But I guess I can use assessment_fra_2025.country to get the same list that  is in public.country :+1:

yaguzmang: To note: I'm checking and for instance public.regions has 15 regions and assessment_fra_2025.region has 14. "forest_europe" is in public.region but not in assessment_fra_2025.region. I will add it manually so the labels are added for the public table.

yaguzmang: After checking out the migration step for country calling code, I decided to do something similar to get all country isos and regions (independent of cycle):
`src/tools/migrations/steps/steps/20250821062102-step-iso-2-calling-code.ts`:
```typescript
const countryISOs = await client.map(`select country_iso from public.country`, [], (row) => row.country_iso)
```

So now I just get all countries/regions from public, generate the labels, and then use them to update the tables in the public and cycle schemas.

Let me know how that looks. If we still want to use AreaController, maybe we should add a couple new functions that get the data from the public schema instead of the data for a specific cycle. What do you guys think?

src/tools/addCountryRegionLabels/index.ts
yaguzmang: Now that I added the code to add the labels to the cycle schema, I've also added the CacheController call :+1:

src/tools/addCountryRegionLabels/index.ts
```diff
@@ -0,0 +1,7 @@
+import '../scriptInit'
+
+import { ToolsUtils } from 'tools/utils/toolsUtils'
+
+import { addCountryRegionLabels } from './addCountryRegionLabels'
+
+ToolsUtils.exec(addCountryRegionLabels)
```
yaguzmang, line 7: From the task: 
> add migration step that calls the tool to update instances

Calling a function directly from this file would run this line `ToolsUtils.exec`, which inside closes the DB pool. That causes an error if called from a migration step. To avoid that, I moved the `addCountryRegionLabels` function to its own file.

minotogna: Yes ,  but you can have an inner file that exports a function that is called from both , the migration step and the tools

src/tools/addCountryRegionLabels/index.ts
minotogna: True , don't use specific cycle counties . Use the query you did or add a repository method

sorja: Good point! My comment can be ignored then and resolved

src/tools/migrations/steps/steps/20250926172049-step-add-country-region-labels.ts
```diff
@@ -0,0 +1,39 @@
...
+  await Promise.all(
+    allCycles.map(({ assessment, cycle }) => {
+      const schemaName = Schemas.getNameCycle(assessment, cycle)
+      return Promise.all([
+        DB.none(`alter table ${schemaName}.country add column if not exists labels jsonb not null default '{}'::jsonb`),
+        DB.none(`alter table ${schemaName}.region add column if not exists labels jsonb not null default '{}'::jsonb`),
+      ])
+    })
+  )
+}
```
minotogna: the labels should be added only to the public schema tables.

minotogna (approved):

PR #5089, opened 2025-10-14
4005 - Admin countries sort by name - Add BE/FE sorting

yaguzmang:
Switching UI languages:

https://github.com/user-attachments/assets/617b6838-b2ac-48b1-9c7d-f1dc93794211

Introduces `TablePaginatedActions.refetchData`, which uses the current filters/order and other table settings.

Coming next PR:
- [ ]  .
>"in UI, everywhere we use country label, use the [country|region].labels[lang] property"


src/client/pages/AdminCountries/hooks/useRefetchCountries.ts
```diff
@@ -0,0 +1,28 @@
...
+  const { assessmentName, cycleName } = useCycleRouteParams()
+  const lang = useLanguage()
+
+  const previousLang = usePrevious(lang, lang)
+
+  useEffect(() => {
+    if (previousLang === lang) return
+
+    const params = { assessmentName, cycleName, lang, limit: 30, path }
+    dispatch(TablePaginatedActions.refetchData(params))
```
minotogna: Cant we use already existing  actions ? 

Or pass a prop to table paginated refechOnUpdate : Array ?

yaguzmang: We could use the current `TablePaginatedActions.getData` but since it requires all params to be passed, we would end up having something very similar to the already existing `useFetchData` hook.

Passing a prop to refetch sounds good though, will try that :+1:

yaguzmang: Doing it that way, I would need to add `lang` as a param directly in tablepaginated `useFetchData`, so anytime we change the language, the tables would be refetched. Not sure if we want that for all tables.

Also we would need to add this effect:
```typescript
  useEffect(() => {
    if (!isInitialized) {
      previousrefetchOnUpdateRef.current = refetchOnUpdate
      return
    }

    const previous = previousrefetchOnUpdateRef.current
    const current = refetchOnUpdate

    const hasChanged =
      previous.length !== current.length || current.some((value, index) => !Object.is(value, previous[index]))

    previousrefetchOnUpdateRef.current = current

    if (!hasChanged) return

    if (counter?.show) {
      throttledGetCount(countParams)
    }
    throttledGetData(dataParams)
  }, [counter?.show, countParams, dataParams, isInitialized, refetchOnUpdate, throttledGetCount, throttledGetData])
``` 

Seems a bit more complicated to follow than dispatching the new action, no? what do you think

yaguzmang: Also, we have for example `useRefetchUsers` in country home. We could reuse the new action there since that refetch is triggered by sending an invitation, not necessarily by updating something as the `refetchOnUpdate` approach would imply

minotogna: I wouldn't touch the user in this pr.
And I would like to keep things as simple as possible .

Something to consider : when the language change , the table paginated should reset to its initial state : I.e. the data should then start from page 1.

To keep things simple, 2 things I would try : 
First : append lang to the path (path?lang=en)
Second : if first doesn't work easily , pass a prop useLang:Boolean (default false) that is handled internally by the table paginated .

What do you think ?

yaguzmang: The simplest was adding the lang to the path (and adding a bit of logic to handle that in the getCount action)

I removed the new action 👍

minotogna: good to hear @yaguzmang 
Please, cascade update the branches to development , as we have migrated the client/hooks import .

yaguzmang: Done :+1:

minotogna (approved):

PR #5099, opened 2025-10-15
ci(Mergify): configuration update

sorja:
This change has been made by @sorja from the Mergify config editor.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,5 +1,9 @@
+merge_queue:
+    max_parallel_checks: 1
+
 queue_rules:
   - name: default
+    batch_size: 1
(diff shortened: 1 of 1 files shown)


sorja:
Error message:

Configuration not compatible with a branch protection setting

The branch protection setting Require branches to be up to date before merging is not compatible with draft PR checks. To keep this branch protection enabled, update your Mergify configuration to enable in-place checks: set merge_queue.max_parallel_checks: 1, set every queue rule batch_size: 1, and avoid two-step CI (make merge_conditions identical to queue_conditions). Otherwise, disable this branch protection.

minotogna (approved):

PR #5102, opened 2025-10-16
4005 - Admin countries sort by name - Use DB labels in FE

yaguzmang:
From main task:

> in UI, everywhere we use country label, use the [country|region].labels[lang] property

### DIFF
--- a/src/meta/area/areas.ts
+++ b/src/meta/area/areas.ts
@@ -1,9 +1,10 @@
 import { i18n } from 'i18next'
 import { Strings } from 'utils/strings'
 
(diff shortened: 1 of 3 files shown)


src/meta/area/areas.ts
```diff
@@ -31,6 +32,17 @@ const getLocale = (isoCode: string): string => {
 
 const getListName = (isoCode: string, i18n: i18n): string => i18n.t(`area.${isoCode}.listName`)
 
+const getName = (area: Country | Region, lang: Lang, fallbackLang: Lang = Lang.en): string => {
```
minotogna: can we omit fallbackLang ? lang: Lang = Lang.en ?

PR #5124, opened 2025-10-20
3426 - Migrate arena-core types (part 1)

sorja:


src/meta/expressions/context.ts
```diff
@@ -0,0 +1,4 @@
+export interface ExpressionContext {
```
minotogna: Should we introduce generic for object type ?

src/meta/expressions/context.ts
```diff
@@ -0,0 +1,4 @@
+export interface ExpressionContext<T = any> {
```
minotogna: Can we use unknown?

sorja: Sure, should we use just unknown or Record<string, unknown>?

minotogna: not sure what the context can be, not necessarily a Record or is it ?

sorja: What I thought too, let's use `unknown` then

minotogna: Is it normal , tests don't pass ?

sorja: I think it's ok for now, build doesn't work because I haven't migrated the actual code, only types, so I think TS is confused.
This will be fixed in next PR 

TL;DR: Next PR build will be fixed

minotogna (approved):

PR #5125, opened 2025-10-21
3426 - Arena-Core migration and JSEP: Object

### LINKED ISSUE #3426: Arena-Core migration and JSEP: Object
Migrate away from arena-core
Add support for Object type for JSEP

https://github.com/openforis/arena-core/blob/774279b208306387e594dd554f045aea2beefea2/src/expression/javascript/parser/jsep.ts#L4

https://github.com/EricSmekens/jsep#jsep-provided-plugins

https://github.com/EricSmekens/jsep/tree/master/packages/object

sorja:
eslint sort order
update to latest version of jsep


minotogna (changes requested):
@sorja lib should be added to the simple-import-sort eslint configuration between client and docs

image

minotogna (approved):

PR #5140, opened 2025-10-23
4591 - Original Data Point Comments Separation - migration step

yaguzmang:


src/server/repository/assessmentCycle/originalDataPoint/create.ts
```diff
@@ -39,10 +40,9 @@ export const create = async (
           data_source_additional_comments,
           data_source_methods,
           data_source_references,
-          description,
           national_classes,
           values
-        ) values ($1, $2, $3, $4::jsonb, $5, $6, $7::jsonb, $8::jsonb) returning *;`,
+        ) values ($1, $2, $3, $4::jsonb, $5, $7::jsonb, $8::jsonb) returning *;`,
```
minotogna, line 45: Before there was an empty value passed in the description .
Description is still passed . 
We either pass both and add one value or we remove it from both .
I would suggest creating the two description with empty value

src/tools/migrations/steps/steps/20251023213324-step-separate-odp-comments.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (!hasForestCharacteristicsComments) {
+    await client.none(`alter table ${tableName} add column comments_forest_characteristics text`)
+    await client.none(`update ${tableName} set comments_forest_characteristics = comments_extent_of_forest`)
+  }
+}
+
+export default async (client: BaseProtocol): Promise<void> => {
+  const assessments = await AssessmentController.getAll({}, client)
+
+  await Promises.each(assessments, async (assessment) =>
```
minotogna, line 69: Only fra has the table if I remember correctly

src/tools/migrations/steps/steps/20251023213324-step-separate-odp-comments.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  const hasForestCharacteristicsComments = await _columnExists({
+    client,
+    columnName: 'comments_forest_characteristics',
+    schemaName,
+  })
+
+  if (!hasForestCharacteristicsComments) {
+    await client.none(`alter table ${tableName} add column comments_forest_characteristics text`)
```
minotogna: DDL should run outside transaction . 
For alter table use DB

src/tools/migrations/steps/steps/20251023213324-step-separate-odp-comments.ts
minotogna: no, also panEuropean has it. than this is fine

src/server/repository/assessmentCycle/originalDataPoint/create.ts
yaguzmang: I agree we can pass both. Can it be in the next PR? in it I will edit the ODP type and start adjusting where needed

minotogna (approved):

PR #5173, opened 2025-10-27
5151 - Admin edit self from country page -> Invalid role shown

### LINKED ISSUE #5151: Admin edit self from country page -> Invalid role shown
When admin edits himself from country page, a wrong form is shown.

Image

**Wrong behavior:** At the moment it shows the Role select field

Image

**Expected behavior:** Show the form fields like when user admin edits himself from admin page

Image

sorja:


src/client/pages/User/hooks/useRolesFields.ts
```diff
@@ -65,14 +73,14 @@ export const useRolesFields = (props: PropsFormDefinition): Returned => {
         shouldShow,
         defaultValue: userRoles.filter((role) => role.role === roleName).map((role) => role.countryIso),
         watches: {
-          clearIf: (props) => {
+          clearIf: (props): { shouldClear: boolean; clearValue: unknown } => {
```
minotogna: can we use return type from types definitions ? e.g. `ReturnType<typeof FieldDefinition['watches']['clearIf']>` ?

minotogna (commented):
code looks good. it will work. just a minor comment on the types.

minotogna (approved):

PR #5175, opened 2025-10-27
5169 - Remove country.config

### LINKED ISSUE #5169: DB: public.country -> remove config column

sorja:
Git diff is not shown for:
```
src/tools/migrations/public/steps/data/countries.ts
```

This file changed:
only country_iso is kept as array which is mapped to [{ country_iso }] for easy insert.


src/tools/migrations/steps/steps/20251027105336-step-drop-column-country-config.ts
```diff
@@ -0,0 +1,12 @@
+import { DB } from 'server/db'
+
+export default async (): Promise<void> => {
+  // drop paneuropean 2020 odp tables/views
+  await DB.none(`drop table if exists assessment_paneuropean_2020.original_data_point cascade;`)
+  await DB.none(`drop view if exists assessment_paneuropean_2020.original_data_point_data cascade;`)
+  // drop paneuropean 2025 odp tables/views
+  await DB.none(`drop table if exists assessment_paneuropean_2025.original_data_point cascade;`)
+  await DB.none(`drop view if exists assessment_paneuropean_2025.original_data_point_data cascade;`)
```
sorja, line 9: @minotogna 

I noticed that for some reason we have odp tables and views for panEuropean
They are empty, but had reference to public.country.config (error when dropping column)

Should we keep them (must be recreated) or discard?

minotogna: We can drop them @sorja

src/server/db/repository/assessmentCycle/cycle/create.ts
```diff
@@ -50,8 +50,8 @@ export const create = async (params: Props, client: BaseProtocol = DB): Promise<
   if (withCountries) {
     // Init countries
     await client.query(`
-      insert into ${schemaCycle}.country (country_iso, props)
-      select country_iso, config
+      insert into ${schemaCycle}.country (country_iso)
+      select country_iso
```
sorja, line 54: @minotogna 
Notable change

minotogna (approved):

PR #5179, opened 2025-10-27
5170: Move redis under src/server/cache with inner repository folder

### LINKED ISSUE #5170: Move redis under src/server/cache with inner repository folder

minotogna:
New server/cache folder structure
image

New eslint server  sort order 
image


sorja (approved):
Nice, should we also move repostiory(/controller) under DB like for cache?

PR #5216, opened 2025-10-29
5160 - assessment_cycle: replace assessment_id with assessment_uuid

sorja:


src/tools/migrations/steps/steps/20251028134533-step-generic-replace-ref-id-with-uuid.ts
```diff
@@ -0,0 +1,33 @@
+import { BaseProtocol } from 'server/db/db'
+
+// Update table_id to table_uuid for:
+// - assessment_cycle -> assessment
+
+export default async (client: BaseProtocol): Promise<void> => {
+  // Add new UUIDs
+  await client.none(`alter table public.assessment_cycle add column if not exists assessment_uuid uuid`)
```
minotogna: ddl with DB . here, line 19, and 30

sorja: Added:
```
const client: BaseProtocol = DB
```

Replaced all clients with this const
Added cache generation for assessments

minotogna (approved):

PR #5232, opened 2025-10-31
5189: Barrel files refactor: meta/api/endpoint

### LINKED ISSUE #5189: Barrel files refactor: meta/api/endpoint

minotogna:


src/meta/api/endpoint/_utils.ts
```diff
@@ -0,0 +1,3 @@
+export const joinPaths = (...parts: Array<string>): string => `/${parts.join('/')}`
+
+export const apiPath = (...parts: Array<string>): string => joinPaths('api', ...parts)
```
sorja, line 3: I would get rid of these, and just write the path in string e.g. `'/api/admin/countries'`.

This would make searching and grepping easier

sorja (approved):

src/meta/api/endpoint/_utils.ts
minotogna: separate task ? in the future ?

sorja: sure

PR #5255, opened 2025-11-07
5162 - TableSection: Refer to section via uuid

sorja:


src/server/controller/metadata/createTableSection.ts
```diff
@@ -9,7 +9,7 @@ import { ActivityLogRepository } from 'server/db/repository/public/activityLog'
 
 // Controller to create TableSection
 export const createTableSection = async (
-  props: { user: User; assessment: Assessment; tableSection: Pick<TableSection, 'props'> },
+  props: { user: User; assessment: Assessment; tableSection: Pick<TableSection, 'props' | 'sectionUuid'> },
```
sorja, line 12: Type safety change:
- Originally we never required sectionId for some reason, but added as it's still needed (required) in repository. Maybe it was forgotten in the original code

src/tools/migrations/steps/steps/20251028134533-step-generic-replace-ref-id-with-uuid.ts
```diff
@@ -94,5 +94,7 @@ export default async (): Promise<void> => {
     `)
 
     await client.none(`alter table ${schemaName}.table_section drop column if exists section_id`)
+
+    await CacheController.generateMetadata({ assessment }, client)
```
sorja, line 98: Added cache generation in migration step for metadata

minotogna (approved):

PR #5259, opened 2025-11-10
5163 - Table.tableSection.uuid: Server and tests

sorja:


src/tools/migrations/steps/steps/metadata/dashboard/utils/index.ts
```diff
@@ -145,7 +145,7 @@ export const getTable = (props: GetTableProps): Returned => {
   const { cols, cycle, rowMetadata, tableId, tableName } = props
   const table: Returned = {
     id: tableId,
-    tableSectionId: -1,
+    tableSectionUuid: '00000000-0000-0000-0000-000000000000' as UUID,
```
sorja, line 148: Unused property by dashboard - doesn't cause problems even though we dont migrate it now:
Next time we update dashboard, it will get handled automatically because of this change

minotogna (approved):

PR #5261, opened 2025-11-11
4591 - Original Data Point Comments Separation - Activity log labels & update cloneData

yaguzmang:
https://github.com/user-attachments/assets/b326e903-17a7-4cbd-8949-839a73dba421


src/meta/assessment/activityLogs/index.ts
```diff
@@ -21,8 +23,8 @@ const messageToKey: { [key in keyof typeof ActivityLogMessage]?: string } = {
   [ActivityLogMessage.originalDataPointCreate]: 'added',
   [ActivityLogMessage.originalDataPointRemove]: 'deleted',
   [ActivityLogMessage.originalDataPointUpdateDataSources]: 'updated',
-  [ActivityLogMessage.originalDataPointUpdateCommentExtentOfForest]: 'updated',
-  [ActivityLogMessage.originalDataPointUpdateCommentForestCharacteristics]: 'updated',
+  [ActivityLogMessage.originalDataPointUpdateCommentExtentOfForest]: 'edited',
+  [ActivityLogMessage.originalDataPointUpdateCommentForestCharacteristics]: 'edited',
```
sorja: Should the used values be enums?
Why did we introduce new one over update?

minotogna: Let's use updated in this case. 

and let's not refactor to enums for now. we'll look into this later on

yaguzmang: Yes we should look into it. The translation files don't have a key for 'updated', so this function later on returns the default "edited":
```typescript
const getLabelAction = (props: { activity: ActivityLog<unknown>; t: TFunction }): string => {
  const { activity, t } = props
  const labelActionKey = _getLabelActionKey(activity)
  const messageParams = _getLabelActionParams(activity, t)
  const label = t(labelActionKey, messageParams)
  return label !== labelActionKey ? label : t('landing.recentActivity.actions.edited')
}

```
I switched it back to 'updated' anyways as it still shows "edited ....", so everything is easier to follow later on when we look into it :+1:

minotogna (approved):

PR #5271, opened 2025-11-12
4591 - Original Data Point Comments Separation - Message topics

yaguzmang:
TODO: 
 - [x] Migrate all message topics with keys including "nationalDataPointComments" to "nationalDataPointExtentOfForestComments"
 

https://github.com/user-attachments/assets/e8cd0e23-a917-4780-8e0f-adc66b179ae7

### DIFF
--- a/src/client/pages/OriginalDataPoint/components/Comments/useCommentsActions.ts
+++ b/src/client/pages/OriginalDataPoint/components/Comments/useCommentsActions.ts
@@ -1,22 +1,45 @@
 import { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 
+import { OriginalDataPointCommentKey } from 'meta/assessment/originalDataPoint'
+import { SectionNames } from 'meta/assessment/section'
+import { TableNames } from 'meta/assessment/table'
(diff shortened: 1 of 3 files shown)


src/client/pages/OriginalDataPoint/components/Comments/useCommentsActions.ts
```diff
@@ -1,22 +1,47 @@
...
 
-export const useCommentsActions = (): Array<DataRowAction> => {
+type Props = {
+  field: OriginalDataPointCommentKey
+}
+
+const commentTopicSuffix: Record<OriginalDataPointCommentKey, string> = {
+  [TableNames.extentOfForest]: 'nationalDataPointExtentOfForestComments',
+  [TableNames.forestCharacteristics]: 'nationalDataPointForestCharacteristicsComments',
+}
```
yaguzmang, line 20: Should I add these as an enum or const under `src/meta/messageCenter/index.ts`?

minotogna: If they are used only here maybe no need for now

minotogna (approved):

PR #5272, opened 2025-11-12
5165 - migration/ddl/meta: col use row_uuid

sorja:


src/server/controller/cycleData/tableData/dropViews.ts
```diff
@@ -0,0 +1,24 @@
...
+import { BaseProtocol, DB } from 'server/db/db'
+import { TableRepository } from 'server/db/repository/assessment/table'
+import { Schemas } from 'server/db/schemas'
+
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+}
+
+export const dropViews = async (props: Props, client: BaseProtocol = DB): Promise<void> => {
```
sorja, line 13: util used in migration to drop existing data views

src/server/controller/cycleData/tableData/dropViews.ts
```diff
@@ -0,0 +1,24 @@
...
+  const { assessment, cycle } = props
+  const tables = await TableRepository.getMany({ assessment, cycle }, client)
+  const schemaCycle = Schemas.getNameCycle(assessment, cycle)
+
+  await Promise.all(
+    tables.map((table) => {
+      const tableName = table.props.name.toLowerCase()
+      return client.none(`drop view if exists ${schemaCycle}.${tableName} cascade`)
```
minotogna: We Should use DB for ddl

src/tools/migrations/steps/steps/20251028134533-step-generic-replace-ref-id-with-uuid.ts
```diff
@@ -143,7 +148,32 @@ export default async (): Promise<void> => {
...
+          on update cascade
+          on delete cascade
+    `)
+
+    await client.none(`alter table ${schemaName}.col drop column if exists row_id`)
+
     await CacheController.generateMetadata({ assessment }, client)
+
+    // recreate dropped views
+    await Promise.all(assessment.cycles.map((cycle) => TableData.refreshViews({ assessment, cycle }, client)))
```
minotogna, line 175: This before cache generation

src/server/controller/cycleData/tableData/dropViews.ts
sorja: const client: BaseProtocol = DB

src/tools/migrations/steps/steps/20251028134533-step-generic-replace-ref-id-with-uuid.ts
sorja: Shouldn't affect? We are generating metadata cache? I was 50/50 if we should also generate data cache after generating the views, but nothing should change in this case?

minotogna: It's just a logical order

sorja: sure, updated

src/server/controller/cycleData/tableData/dropViews.ts
minotogna: This is not guaranteed. From outside a transaction can be passed

sorja: Updated as so:

```
...
// use client to get the tables
  const tables = await TableRepository.getMany({ assessment, cycle }, client)
...

// use DB to drop the views
      return DB.none(`drop view if exists ${schemaCycle}.${tableName} cascade`)
...

```

minotogna (approved):

PR #5283, opened 2025-11-13
4591 - Original Data Point Comments Separation - Admin links update

yaguzmang:
https://github.com/user-attachments/assets/04fe1040-a763-48fb-ad74-2f320403c7f9

### DIFF
--- a/src/tools/migrations/steps/steps/20251023213324-step-separate-odp-comments.ts
+++ b/src/tools/migrations/steps/steps/20251023213324-step-separate-odp-comments.ts
@@ -78,38 +78,9 @@ const _updateOriginalDataPointTable = async (props: UpdateTableProps): Promise<v
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/tools/migrations/steps/steps/20251023213324-step-separate-odp-comments.ts
```diff
@@ -78,6 +78,40 @@ const _updateOriginalDataPointTable = async (props: UpdateTableProps): Promise<v
   }
 }
 
+const _updateOdpLinkLocations = async (props: UpdateTableProps): Promise<void> => {
+  const { client, schemaName } = props
+  const commentColumnExtent = ODPCommentColumns[TableNames.extentOfForest]
+
+  // Replace all odp link locations to use new comment columns instead of "description"
+  await client.none(
```
minotogna: Would it be easier to just remove all links and re run the job ?

minotogna (approved):

PR #5298, opened 2025-11-14
5166 - Simple import

sorja:


src/server/controller/metadata/index.ts
```diff
@@ -22,6 +23,7 @@ import { updateTableSection } from './updateTableSection'
 export const MetadataController = {
   // export
   exportAll,
+  importAll,
```
minotogna, line 26: @sorja this pr looks good to me.
However, originally I had in mind to import metadata and areas (and in the future data) through their respective controllers. however, looking at the current implementation, it's a generic table export/import. I think in a next pr we should move import and export methods.
Should we add under db a DBService with two methods importTables and exportTables ?
Any other suggestions ?

src/server/controller/metadata/export/importAll.ts
```diff
@@ -0,0 +1,9 @@
+import { ExportedTableData } from 'meta/metadata/export'
+
+import { BaseProtocol, DB } from 'server/db/db'
+
+import { importTable } from './importTable'
+
+export const importAll = async (allData: Array<ExportedTableData>, client: BaseProtocol = DB): Promise<void> => {
+  await Promise.all(allData.map((tableData) => importTable(tableData, client)))
```
minotogna, line 8: I also think, that there will be foreign keys order to respect. so an order should be respected when importing

minotogna (approved):

PR #5302, opened 2025-11-14
5166 - Database Service

sorja:
Public schema initialisation works on empty db


sorja:
@minotogna Should we remove public migration to create countries? Seems weird we have only that one, and missing e.g. regions - would be better to import?

src/server/service/databaseService/EXPORT_TABLES.ts
```diff
@@ -0,0 +1,23 @@
...
+// const EXPORT_ASSESSMENTS = [AssessmentNames.fra, AssessmentNames.panEuropean]
+// // Tables ordered by foreign key dependencies
+// const EXPORT_ASSESSMENT_TABLES = ['section', 'table_section', 'table', 'row', 'col']
+//
+// const ASSESSMENT_TABLES = EXPORT_ASSESSMENTS.flatMap((assessmentName) =>
+//   EXPORT_ASSESSMENT_TABLES.map((tableName) => ({
+//     schema: Schemas.getName({ props: { name: assessmentName } }),
+//     table: tableName,
+//   }))
+// )
```
sorja, line 12: Been thinking of how we should initialise the assessment schemas, maybe if schema name not found, we can run the initialisation DDL? Any ideas? Or before any imports, we initialise given set (fra, paneu) schemas

minotogna: in the e2e they should always be initialized through the import

src/server/service/databaseService/EXPORT_TABLES.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const EXPORT_TABLES: Array<ExportTableProps> = [
+  // ===== Schema: Public
+  { schema: 'public', table: 'assessment' },
+  { schema: 'public', table: 'assessment_cycle' },
+  // { schema: 'public', table: 'country' }, // omitted: import via public migration step
```
minotogna: should we remove that public step and export/import countries here as well ?

minotogna:
> @minotogna Should we remove public migration to create countries? Seems weird we have only that one, and missing e.g. regions - would be better to import?

yes @sorja I asked the same question before reading your comment 😅

src/server/service/databaseService/EXPORT_TABLES.ts
sorja: Definitely yes!

minotogna (approved):

PR #5305, opened 2025-11-15
5304: Fix Country Last published info: invalid when next cycle is published

### LINKED ISSUE #5304: Country Last published info: invalid when next cycle is published

minotogna:
image

image

image

### DIFF
--- a/src/server/db/repository/assessmentCycle/country/getCountryLastPublishedInfo.ts
+++ b/src/server/db/repository/assessmentCycle/country/getCountryLastPublishedInfo.ts
@@ -1,12 +1,20 @@
+import { Dates } from 'utils/dates'
+
 import { LastPublishedInfo } from 'meta/area/country'
 import { CountryIso } from 'meta/area/countryIso'
 import { CountryStatus } from 'meta/area/countryStatus'
 import { Assessment } from 'meta/assessment/assessment'
+import { Cycle } from 'meta/assessment/cycle'
+import { Cycles } from 'meta/assessment/cycles'
 
 import { BaseProtocol, DB } from 'server/db/db'
 import { Schemas } from 'server/db/schemas'
 
-type Props = { assessment: Assessment; countryIso?: CountryIso }
+type Props = {
+  assessment: Assessment
+  countryIso?: CountryIso
+  cycle: Cycle
+}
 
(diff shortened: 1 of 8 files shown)


src/tools/migrations/steps/steps/template.ts
```diff
@@ -1,5 +1,5 @@
 import { BaseProtocol } from 'server/db/db'
 
-export default async (client: BaseProtocol): Promise<void> => {
+export default async (_client: BaseProtocol): Promise<void> => {
```
sorja: this shouldnt be renamed

src/server/db/repository/assessmentCycle/country/getCountryLastPublishedInfo.ts
```diff
@@ -1,22 +1,42 @@
...
 export const getCountryLastPublishedInfo = async (
   props: Props,
   client: BaseProtocol = DB
 ): Promise<LastPublishedRecord> => {
-  const { assessment, countryIso } = props
+  const { assessment, countryIso, cycle } = props
+
+  const cycleDateCreated = Dates.parseISO(cycle.props.dateCreated)
+
+  let include = true
```
sorja, line 29: can we return true/false from fn without defining it outside?

src/tools/migrations/steps/steps/template.ts
minotogna: Tsc compilation problems . When we create a migration step we remove the _ 

Would that work ?

src/server/db/repository/assessmentCycle/country/getCountryLastPublishedInfo.ts
minotogna: That's iterative . A published cycle can have successor cycles not published . They all must be excluded

src/tools/migrations/steps/steps/template.ts
sorja: update `.eslintignore` is better like it was

minotogna: @sorja I had to exclude it from both tsconfig and eslintignore https://github.com/openforis/fra-platform/pull/5305/commits/d04d9b26db5f9eb0463576262ffd8ae84759cf20

sorja (approved):

sorja (approved):

PR #5309, opened 2025-11-17
5166 - Assessment Schema import/export

sorja:
image


src/tools/data/import.ts
```diff
@@ -1,26 +1,33 @@
...
 const INPUT_DIR = path.join(__dirname, 'fixtures')
 
 const exec = async (): Promise<void> => {
-  const files = await fs.promises.readdir(INPUT_DIR, { recursive: true })
+  // Create assessment schemas
+  await Promise.all(
+    [AssessmentNames.fra, AssessmentNames.panEuropean].map((assessmentName) => {
+      return AssessmentRepository.createAssessmentSchema({ assessmentName })
+    })
+  )
```
sorja, line 23: This is hardcoded for now - I don't think we ever really want to change this.
If we want, we can (later) extract the schemas from `fixtures/<schemaName>`

src/tools/data/import.ts
```diff
@@ -1,26 +1,33 @@
...
 
   await DB.tx(async (t) => {
-    // array of [[schema, table], ..]
-    const schemaTables = files.reduce((acc, file) => {
-      if (file.endsWith('json')) acc.push(file.split('/'))
-      return acc
-    }, [])
-    const allData = schemaTables.map(([schema, table]) => _readTableFromFile(schema, table, INPUT_DIR))
+    // Use EXPORT_TABLES order to respect foreign key dependencies
+    const allData = EXPORT_TABLES.map((tableConfig) => {
```
sorja, line 27: Instead of relying on the JSON file order, use static const `EXPORT TABLES` to import in correct order (to avoid FK issues)

minotogna (approved):

PR #5330, opened 2025-11-19
5290 - Upgrade ESlint to version 9

### LINKED ISSUE #5290: Upgrade ESlint to version 9 and use Multithread linting
https://eslint.org/blog/2025/08/multithread-linting/
https://eslint.org/docs/latest/use/configure/migration-guide

yaguzmang:
TODO:
- [x] Finish testing


sorja:
Hey, does this use the flat config format? If not, should we?

eslint.config.mjs
```diff
@@ -0,0 +1,204 @@
...
+          extensions: ['.js', '.jsx', '.ts', '.tsx'],
+        },
+        webpack: {
+          config: 'webpack.config.babel.js',
+        },
+        typescript: {},
+      },
+    },
+    rules: {
+      camelcase: [
```
minotogna, line 67: is this a new rule? it's a bit risky, we do want camel case properties and manually ignore (e.g. db objects)

minotogna: is there any new rule you added ? can you highlight them please ?

eslint.config.mjs
```diff
@@ -0,0 +1,204 @@
...
+        node: {
+          extensions: ['.js', '.jsx', '.ts', '.tsx'],
+        },
+        webpack: {
+          config: 'webpack.config.babel.js',
+        },
+        typescript: {},
+      },
+    },
+    rules: {
```
minotogna, line 66: @yaguzmang in current dev branch I have added this rule:
`'class-methods-use-this': 'off',`
Please add this as well.

yaguzmang:
> Hey, does this use the flat config format? If not, should we?

@sorja Yes we are using the new flat config format 👍

eslint.config.mjs
yaguzmang: Here is a summary:
- `camelcase`: was included in the airbnb rules which we dropped, so I manually added it. What it does:
image

We have a couple places in the code base where we ignored it. For example:

```typescript
export const SectionAdapter = (section: SectionDB): Section => {
  // eslint-disable-next-line camelcase
  const { props, sub_sections, ...restSection } = section

  return {
    ...restSection,
    props,
    // eslint-disable-next-line camelcase
    subSections: sub_sections?.map(SubSectionAdapter),
  }
}
```
Let me know if you want me to remove it. I think it's not supper necessary. cc @sorja 

The rest of the ones that I added manually were also already included with the airbnb pluggin:
- `global-require`
- `import/no-dynamic-require`
- `max-classes-per-file` 
- `no-alert`
- `no-await-in-loop` 
- `no-promise-executor-return` 
- `no-restricted-syntax`
- `react/jsx-props-no-spreading`

Regarding the other new imports in the flat config:
```
  js.configs.recommended,
  ...tseslint.configs.recommended,
  reactPlugin.configs.flat.recommended,
  jsxA11yPlugin.flatConfigs.recommended,
  ```
  - ` js.configs.recommended` : This is the flat-config version of `eslint:recommended` from `@eslint/js`. These were already in use because airbnb rules extended from them
  - `tseslint.configs.recommended`: We already had it.
  - `reactPlugin.configs.flat.recommended`: These ones also were already included by airbnb
  - `jsxA11yPlugin.flatConfigs.recommended`: Was also used by airbnb.
  
  So the new config is as close as I've been able to mimic the old one while dropping airbnb and not using any non-native packages/plugins. If anything we should now have fewer rules.

yaguzmang: In the newest commit:
- `'@typescript-eslint/no-explicit-any': 'warn'`: in the new version this rule throws an error, I set it to warn to match our previous config.
- `'@typescript-eslint/no-non-null-assertion': 'warn'`: In the new version this rule is off. I'm turning this rule on to warning to match our previous config too.
- `'func-names': 'warn',`, `'react/no-danger': 'warn'` and `'no-alert': 'warn',`: set it to match old 'warn' severity like in the previous version.
- `'react/display-name': 'off',`: This rule is set to error in the new react recommended rules. Examples of this rule:
image
The point of it is to have display names for debugging tools. Let me know if I should keep it. We would need to edit files like `InputNumber.tsx`, `InputPercent.tsx`, `DataGrid.tsx`, among others.

minotogna: Thanks very much @yaguzmang 🤩
To me it all looks great 👍 
@sorja you want to have a final review , resolve this comment and approve eventually ?

sorja (approved):

PR #5337, opened 2025-11-19
5159 - Remove AWS from test-all & use fixtures

sorja:


src/server/service/databaseService/EXPORT_TABLES.ts
```diff
@@ -65,6 +65,10 @@ const ASSESSMENT_CYCLE_TABLES = EXPORT_ASSESSMENTS.flatMap((assessmentName) =>
 
 export const EXPORT_TABLES: Array<ExportTableProps> = [
   // ===== Schema: Public
+  { schema: 'public', table: 'users', skipExport: true },
```
minotogna, line 68: Why adding these tables here , if they aren't exported ?
Isn't this the list of tables to export ?

src/test/e2e/healthcheck.sh
```diff
@@ -1,4 +1,3 @@
 #! /bin/bash
 
-result=$(psql -U fra -d fra -t -A -c "SELECT count(*) FROM users WHERE email = 'test@test.com'")
```
minotogna, line 3: Should we check robot instead ?

src/server/service/databaseService/EXPORT_TABLES.ts
sorja: Used in import - we have static fixture for user (test user)

src/test/e2e/healthcheck.sh
sorja: You are right - we need to add migration step to create robot user.

The problem:

- Robot user has no password (can't be used to test logging in etc)
- Need test user (test@test.com) to avoid publishing the password in repository
- Test user only created when running tests in CI

src/server/service/databaseService/EXPORT_TABLES.ts
minotogna: should we have another set of tables to import then ?

sorja: Maybe we can rename this variable better? It's used for both import and export (so we have single source to edit)

with skipExport: we expect to have static (manually added) fixtures, e.g. for the user

sorja:
Things to do in next PR:
  - Remove `ci:start`
  - `server/db/initDb`
  - Import and Export methods: Pass tables from outside
  - Add real uuid for test user
  - Remove concept of migration public

@minotogna

minotogna:
@sorja 

> * `server/db/initDb`

exportTables / importTables / initDatabse should be 3 methods exported from the same place under server/db.
We either have a DBService under server/db/service or three methods under server/db/{nameTODecide} (service?)
anyway under server/db

minotogna (approved):

PR #5361, opened 2025-11-20
5358 - DB: users_auth_provider -> reference user by uuid

sorja:


src/server/db/repository/public/user/getProfilePicture.ts
```diff
@@ -16,7 +16,7 @@ export const getProfilePicture = async (
     where = 'where lower(trim(u.email)) = trim(lower($1))'
     value = props.email
   } else if ('emailGoogle' in props) {
-    where = `where u.id = (select user_id from public.users_auth_provider where props->>'email' = $1)`
+    where = `where u.uuid = (select user_uuid from public.users_auth_provider where props->>'email' = $1)`
```
minotogna: @sorja can you please investigate if we still call this method with all 3 params ? id/email/emailGoogle ?
Maybe some of them can be removed if not used anymore ?

sorja: Can we do in separate issue? I am trying to finish the e2e and this should be one of the last things before it works :)

P.s.

This is the only usage:

```
  const userProfilePicture = await UserRepository.getProfilePicture({ id }, client)

```


Do you want me to remove the other usages in this PR?

minotogna: I guess it's a small change, isn't it ? if so please proceed, otherwise we'll address this later if you prefer

minotogna (approved):

PR #5374, opened 2025-11-21
5159 - service\databaseService: Pass tables from outside (intermediate pr)

sorja:
NEXT Pr will sort the db service and remove public migrations - this is just minor clean up


src/server/service/databaseService/export.ts
```diff
@@ -1,14 +1,18 @@
-import { ExportedTableData } from 'meta/metadata/export'
-
 import { BaseProtocol, DB } from 'server/db/db'
-import { EXPORT_TABLES } from 'server/service/databaseService/EXPORT_TABLES'
+
+export type ExportedTableData<T = unknown> = {
```
minotogna: can we have meaningful names for generics ? e.g. `ROW`

src/tools/data/EXPORT_TABLES.ts
```diff
@@ -4,6 +4,10 @@ import { CycleName } from 'meta/assessment/cycle'
 import { Schemas } from 'server/db/schemas'
 import { ExportTableProps } from 'server/service/databaseService/export'
 
+export type ExportTableConfig = ExportTableProps & {
+  skipExport?: boolean
```
minotogna, line 8: why do we want to still use the skipExport ?

sorja: I explained this yesterday in the meeting:

- We are using skipExport to skip the tables we have static fixtures for

minotogna: Yes , and I suggested to create a subset of tables

minotogna: Why do we pass tables to the export method we don't want to export ?

sorja: The subset of tables we discussed about is created here in `tools` which is passed to DB service:

image

Correct me if I'm wrong - what are you looking for exactly?

minotogna: I see . It's just more code to handle it ,  it's ok if you want it this way

minotogna (approved):

PR #5379, opened 2025-11-21
5159 - Introduce database tool and remove DBService from src/server

sorja:


eslint.config.mjs
```diff
@@ -71,6 +71,7 @@ export default defineConfig([
           ignoreDestructuring: false,
         },
       ],
+      'object-shorthand': ['warn', 'always', { avoidQuotes: true }],
```
minotogna: please, sort it alphabetically , after all @typescript

src/tools/db/initSchemas.ts
```diff
@@ -0,0 +1,11 @@
+import './scriptInit'
+
+import { ToolsUtils } from 'tools/utils/toolsUtils'
+
+import { DBService } from 'server/db/service/dbService'
+
+const exec = async (): Promise<void> => {
```
minotogna, line 6: let's move all db tools under src/tools/db (including importTables and exportTables).
what do you think ?

minotogna (commented):
@sorja we already have src/tools/db. would it makes sense to use this folder instead of  src/tools/database ?

src/tools/database/export.ts
```diff
@@ -4,18 +4,18 @@ import * as path from 'path'
 import { ToolsUtils } from 'tools/utils/toolsUtils'
 
 import { DB } from 'server/db/db'
-import { DatabaseService } from 'server/service/databaseService'
 
-import { _copyStaticFixtures } from './_copyStaticFixtures'
-import { _writeTableToFile } from './_writeTableToFile'
-import { EXPORT_TABLES } from './EXPORT_TABLES'
+import { _copyStaticFixtures } from './utils/copyStaticFixtures'
+import { DBService } from './utils/dbService/dbService'
```
minotogna: I wouldn't create a utils folder. plus we can use absolute import. 
it can be something like 
`import {DBService } from 'tools/db/service'` (note not the double service folder )
the write and read methods can go under 'tools/db/io' . what do you think ?

sorja: sure

minotogna (approved):

PR #5400, opened 2025-11-22
5399: Client/styles cleanup: link

### LINKED ISSUE #5399: Client/styles cleanup: link

minotogna:
- [x] Created client/Links folder with: 
  -   new common Link component with isolated link style
  -   And moved DefinitionLink, CountryLink, LinkHome 
- [x] DefinitionLink and CountryLink uses new Link style
- [x] Start of removing redundant client/styles folder (in favor of new client/style) -> step 1: moved and rename Main.scss
- [x] AdminLinks uses new Link component 
- [x] CountryHome RecentActivity uses new Link component 
- [x] DataTable ODPHeaderCell uses new Link component 
- [x] DataTable CellHeader uses new Link component 
- [x] DataTable SDGHints uses new Link component 

https://github.com/user-attachments/assets/67d57f29-cbd9-4527-92b1-7eceb1477ec8

https://github.com/user-attachments/assets/3ef43fea-d294-44c1-bae3-5b869bae16da


src/client/components/Links/Link/Link.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Link as ReactRouterLink } from 'react-router-dom'
+
+import { LinkProps } from 'client/components/Links/Link/types'
+
+import { useLinkClassName } from './hooks/useLinkClassName'
+
+const Link: React.FC<LinkProps> = (props) => {
+  const { children, className, color, oneLine, rel, target, to } = props
+  // tooltip props
+  const { 'data-tooltip-content': tooltipContent, 'data-tooltip-id': tooltipId } = props
```
sorja, line 11: i wouldnt do this, its clear from the name

one possibility also to pass 

`Props.tooltip = { id, content }`

minotogna: The component is just a wrapper . Using custom names , will require refactoring props if we switch component.
Wrapper components is suggested to have the primitive prop names . 
What do you think @sorja ?

minotogna: Even data col has the same prop names .
We should be consistent , this refactor shouldn't include how to consistently handle data tooltip props . If we really want to have different names , we should refactor all app , with even custom Div component to be consistent .

minotogna: Can we address this discussion in the daily if we need @sorja i have another pr coming 🙏

sorja (approved):

src/client/components/Links/Link/Link.tsx
sorja: Sure, my comment was more on

```
const {} = props
const {} = props
```

PR #5434, opened 2025-11-30
4200 - Vite: initial configuration 

minotogna:


index.html
```diff
@@ -22,5 +22,6 @@
     <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=G-H1GZKMJDG7" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
     <!-- End Google Tag Manager (noscript) -->
     <div id="main"></div>
+    <script type="module" src="/src/client/Main.tsx"></script>
```
sorja, line 25: This file index.html can be moved to src/client/.

minotogna: it's more trouble than a benefit. it requires playing with vite build process. if we leave it there, by default it will be part of the build output and it works nicely in dev and prod builds

minotogna: in your pr @sorja , I didn't see how you handled it in the custom build process.

sorja (approved):

PR #5469, opened 2025-12-04
5363 - Request a password reset

sorja:
https://github.com/user-attachments/assets/f03eaf2a-d768-4bfb-bbe6-a2358cd1afa0


src/client/pages/Login/validationSchemas.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from 'zod'
+
+export const emailValidationSchema = (t: (key: string) => string): z.ZodString => {
+  return z.string().min(1, { message: t('login.emptyEmail') })
```
minotogna: Is there an email validation as well (other than empty )?

src/i18n/resources/en/login.json
```diff
@@ -34,7 +34,11 @@
   "password": "Password",
   "passwordChanged": "Password has been changed",
   "passwordError": "Password must contain six characters or more and have at least one lower case and one upper case alphabetical character and one number",
+  "passwordLowercase": "Password must contain at least one lowercase letter",
```
minotogna, line 37: @sorja im also wondering, isn't it better for the users to see all requirements in one message (like before )? Otherwise while they tap , they might see always different messages ...

sorja: Sure, the message was too long though, what do you suggest?
Should we show the message as separate?

image

src/client/pages/Login/validationSchemas.ts
sorja: Weird, it didn't commit it - check now please

src/i18n/resources/en/login.json
minotogna: ok , let's show them separate. but please , after merging this to dev, update the other languages through crowdin. ok @sorja ?

sorja: Sure thing

minotogna (approved):

PR #5481, opened 2025-12-05
5366 - Move Links APIs under Admin

### LINKED ISSUE #5366: Move Links APIs under Admin
Move Links apis from cycle data to admin

yaguzmang:


src/meta/api/endpoint/cycleData.ts
```diff
@@ -31,15 +31,6 @@ export const CycleData = {
     },
   },
 
-  Links: {
```
minotogna, line 34: This structure is more flexible 😅

yaguzmang: Ah okay, I was following the pattern in admin.ts. I will change it back 👍

minotogna (approved):

PR #5563, opened 2025-12-15
5548 - [1] Regional Focal Point Role: Add Role+label+invitation

minotogna:
pr # 1

image

image


sorja:
We should remove the possibility to write when selecting a role in the invitation (and other similar select fields).
@minotogna

minotogna:
> We should remove the possibility to write when selecting a role in the invitation (and other similar select fields).
> 
> @minotogna 

@sorja why ?

sorja:
> > We should remove the possibility to write when selecting a role in the invitation (and other similar select fields).
> > @minotogna
> 
> @sorja why ?

Why would you want to write when there are 4 or 5 options? I think Select and Searchable select should be separate, but this in another topic

sorja (approved):

minotogna:
> > > We should remove the possibility to write when selecting a role in the invitation (and other similar select fields).
> 
> > > @minotogna
> 
> > 
> 
> > @sorja why ?
> 
> 
> 
> Why would you want to write when there are 4 or 5 options? I think Select and Searchable select should be separate, but this in another topic

Maybe you're right @sorja 
But yes , let's address this in the daily ? (Please don't forget 😊)

PR #5578, opened 2025-12-17
5546 - Invitation - Accept

sorja:
image


src/client/pages/Authentication/Invitation/Accept/hooks/useOnAccept.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    setIsLoading(true)
+    try {
+      const config = { params: { invitationUuid } }
+      const { data } = await axios.post<{ user: User }>(ApiEndPoint.User.invitationAccept(), {}, config)
+      // On accept, update received user to store and navigate to role assessment/cycle/country
+      dispatch(UserActions.setUser(data.user))
+      navigate(Routes.Country.generatePath({ assessmentName, countryIso, cycleName }))
+    } catch (error) {
+      setIsLoading(false)
```
minotogna, line 41: should this be set to false within `finally` ?

src/client/pages/Authentication/Invitation/Invitation.tsx
```diff
@@ -0,0 +1,22 @@
...
+const Invitation: React.FC = () => {
+  const user = useUser()
+  const data = useData()
+
+  if (!data) return null
+  if (!user) return <Register />
+
+  if (data.user.uuid !== user.uuid) {
+    return null
+  }
```
minotogna, line 17: thinking. if a user is logged with another account, what should happen ? Register as well ?

src/client/pages/Authentication/Invitation/Accept/hooks/useOnAccept.ts
sorja: Sure, not sure it matters here because we redirect so this should get unmounted, right? What do you prefer?

src/client/pages/Authentication/Invitation/Invitation.tsx
sorja: I was thinking about this too, but not sure. It would be weird if you can access someone elses invitation (although, then you can access it when you log out)

minotogna: well users can have multiple accounts

src/client/pages/Authentication/Invitation/Accept/hooks/useOnAccept.ts
minotogna: same

src/client/pages/Authentication/Invitation/Invitation.tsx
sorja: Sure, let's redirect to registration then. Next PR is ok (where the Registration is implemented)

minotogna (approved):

PR #5595, opened 2025-12-19
5369 - Dependencies upgrade: i18n - Add all plurals

yaguzmang:


src/i18n/resources/en/admin.json
```diff
@@ -1,7 +1,11 @@
 {
   "admin": "Admin",
+  "country_few": "{{count}} Countries",
+  "country_many": "{{count}} Countries",
   "country_one": "{{count}} Country",
   "country_other": "{{count}} Countries",
+  "country_two": "{{count}} Countries",
+  "country_zero": "{{count}} Countries",
```
minotogna: any chance to use `$t(admin.country_other, {{count}})` (and for all others).
This will avoid replicating the same word. what do you think? can you please verify this will work ?

yaguzmang: Yes better 👍 
i18n passes the param {{count}} to the referenced translation so no need to add it.
Russian also uses more plurals (one, few, many, other) so for that one I only simplified `*_two` → `*_few` and `*_zero` → `*_many`

minotogna (approved):

PR #5606, opened 2025-12-23
5491 - Admin links job on separate Heroku dyno - Add Job status payload

yaguzmang:


src/server/worker/job/jobLock.ts
```diff
@@ -26,18 +56,22 @@ export class JobLock {
...
   }
 
+  public static async disconnect(): Promise<void> {
+    await JobLock.#redlock.quit()
+  }
+
   public async acquireLock(): Promise<void> {
-    this.#lock = await JobLock.#redlock.acquire([`lock:${this.#name}`], 5 * 60 * 1000)
-    await this.setStatus(JobStatus.running)
+    this.#lock = await JobLock.#redlock.acquire([`lock:${this.#name}`], this.#lockDurationMs)
```
minotogna: is there a real need now to customize the duration lock ?

yaguzmang: I've noticed the job can take more than 5 minutes sometimes, so I added the option to make the lock longer

minotogna: ok. but should we use another default value then ? maybe 10 minutes ?

yaguzmang: Okay :+1:

minotogna (approved):

PR #5614, opened 2025-12-24
5491 - Admin links job on separate Heroku dyno - Add VerifyLinksJob

yaguzmang:
The plan:
- Queue verify‑links jobs in the main web process and keep them using Redis, then spin up a dedicated Heroku dyno to consume them.
- Only one dyno should run at a time. If one is already active, new requests are just added to the queue.
- The worker runs with concurrency=1, so the jobs are run sequentially. There will be a single queue for *all* link jobs, regardless of assessment/cycle. Then the worker shuts itself down after the queue stays empty for a brief idle grace period.
- The job status is stored in Redis keyed by assessment/cycle so the admin status endpoint can report queued/running/success/failed per assessment/cycle. That's also necessary so we can disable the Verify Links button per assessment.  (We will add a countryIso to the key in the future).

### DIFF
--- a/src/server/controller/cycleData/links/visitCycleLinks/queueFactory.ts
+++ b/src/server/controller/cycleData/links/visitCycleLinks/queueFactory.ts
@@ -1,17 +1,15 @@
-import { Job, Queue, QueueOptions, Worker } from 'bullmq'
+import { Job, Queue, QueueOptions } from 'bullmq'
 import IORedis from 'ioredis'
 
 import { Assessment } from 'meta/assessment/assessment'
 import { Cycle } from 'meta/assessment/cycle'
 
 import { ProcessEnv } from 'server/utils'
-import { Logger } from 'server/utils/logger'
 
 import { VisitCycleLinksProps } from './props'
-import { WorkerFactory } from './workerFactory'
 
-const queues: Record<string, Queue<VisitCycleLinksProps>> = {}
-const workers: Record<string, Worker<VisitCycleLinksProps>> = {}
+const queueName = 'verifyLinks'
+let queue: Queue<VisitCycleLinksProps> | undefined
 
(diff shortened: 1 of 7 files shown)


src/server/controller/cycleData/links/visitCycleLinks/queueFactory.ts
```diff
@@ -26,35 +24,29 @@ const opts: QueueOptions = {
...
-  queue = new Queue<VisitCycleLinksProps>(key, opts)
-  queues[key] = queue
-
+  queue = new Queue<VisitCycleLinksProps>(queueName, opts)
   return queue
 }
 
-const getActiveJobs = async (props: Props): Promise<Array<Job<VisitCycleLinksProps>>> => {
-  const queue = getInstance(props)
+const getActiveJobs = async (_props: Props): Promise<Array<Job<VisitCycleLinksProps>>> => {
```
yaguzmang, line 34: This will be removed/edited next PR

src/server/worker/tasks/verifyLinksWorkerPresence.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const clearWorkerLock = async (): Promise<void> => {
+  await redis.del(workerPresenceKey)
+}
+
+const disconnect = async (): Promise<void> => {
+  await redis.quit()
+}
+
```
minotogna, line 31: this pr looks good 🎉. however, it might be useful to add some comments, since it's a complicate process :P 
also, all files related to the visitLinks job should be moved under worker/tasks/verifyLinks folder.
the worker/tasks folder should include only the main file that starts the job.

yaguzmang: Added 🙌 
Moving files in following PRs 👍

minotogna (approved):

PR #5616, opened 2025-12-29
5615 - GEO: Extent of Tree cover by source-> reported to Fra last published value

### LINKED ISSUE #5615: GEO: Extent of Tree cover by source-> reported to Fra last published value
- [ ] In the chart and in the table , the value reported to fra should use the last published value fetched through the explorer api . 
- [ ] the sentence ‘reported to fra’ (in both table and chart) should not include the 2020 anymore .

yaguzmang:
image

image

### DIFF
--- a/src/server/api/explorer/getData.ts
+++ b/src/server/api/explorer/getData.ts
@@ -15,9 +15,6 @@ type GetDataParams = {
 }
 type GetDataRequest = Request<never, never, Body, GetDataParams>
 
-const excludeOdpTable = true
-const mergeOdp = true
-
 export const getData = async (req: GetDataRequest, res: Response): Promise<void> => {
   try {
     const { assessment } = req.context
@@ -26,8 +23,6 @@ export const getData = async (req: GetDataRequest, res: Response): Promise<void>
       assessment,
       columns,
       countryISOs,
-      excludeOdpTable,
-      mergeOdp,
       tableNames,
       variables,
     })
--- a/src/server/api/explorer/getData.ts
+++ b/src/server/api/explorer/getData.ts
@@ -2,12 +2,7 @@ import { Request, Response } from 'express'
 
 import { CountryIso } from 'meta/area/countryIso'
 import { AssessmentName } from 'meta/assessment/assessment'
-import { Assessments } from 'meta/assessment/assessments'
-import { CycleName } from 'meta/assessment/cycle'
-import { RecordAssessmentData } from 'meta/data/recordData'
-import { RecordAssessmentDatas } from 'meta/data/recordDatas'
 
-import { AreaController } from 'server/controller/area'
 import { CycleDataController } from 'server/controller/cycleData'
 import Requests from 'server/utils/requests'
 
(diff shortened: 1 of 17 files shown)


src/server/controller/geo/_getLastPublishedFra1aForestArea.ts
```diff
@@ -0,0 +1,61 @@
...
+  })
+
+  const lastYear = Object.keys(tableData)
+    .reduce((years, colName) => {
+      const year = Number.parseInt(colName, 10)
+      if (Number.isFinite(year)) years.push(year)
+      return years
+    }, [])
+    .sort((a, b) => a - b)
+    .at(-1)
```
yaguzmang: Avoiding hardcoding "2025" here

src/server/controller/geo/getForestEstimations.ts
```diff
@@ -13,7 +16,17 @@ import { AssetsController } from './assets'
...
 export const getForestEstimations = async (props: Props, client: BaseProtocol = DB): Promise<ForestEstimations> => {
-  return ForestEstimationsRepository.getOne(props, client)
+  const { countryIso, year } = props
+
+  const forestEstimations = await ForestEstimationsRepository.getOne({ countryIso, year }, client)
+
+  if (!forestEstimations) return forestEstimations
+
+  const assessment = await AssessmentController.getOne({ assessmentName: AssessmentNames.fra }, client)
+  const lastPublishedFra1aForestArea = await _getLastPublishedFra1aForestArea({ assessment, countryIso }, client)
```
minotogna, line 26: can we reuse the code from `src/server/api/explorer/getData.ts` ? we can export that code into the CycleDataController.getLastPublishedData . what do you think ?

yaguzmang: Sounds good, added it 👍

src/server/api/explorer/getData.ts
```diff
@@ -25,55 +20,17 @@ const mergeOdp = true
...
-        return acc
-      },
-      { [assessmentName]: { [publishedCycleName]: {} } }
-    )
+    const { columns, countryISOs, tableNames, variables } = req.query
+    const data = await CycleDataController.getLastPublishedData({
+      assessment,
+      columns,
+      countryISOs,
+      excludeOdpTable,
```
minotogna: This is always true in this context . No need to pass it from outside

src/server/api/explorer/getData.ts
```diff
@@ -25,55 +20,17 @@ const mergeOdp = true
...
+      mergeOdp,
```
minotogna: As well as this .

yaguzmang: do you mean that here in `getData.ts`we do:
```
...
      excludeOdpTable: true,
      mergeOdp: true,
...
```
or that we don't even pass those to `getLastPublishedData`, so they are set to true directly inside `getLastPublishedData`?

yaguzmang: I imagine it's the first option so we keep `getLastPublishedData` generic. Added the changes to the latest commit 👍

src/server/controller/cycleData/getLastPublishedData.ts
```diff
@@ -0,0 +1,73 @@
...
+import { BaseProtocol, DB } from 'server/db/db'
+
+import { getTableData } from './getTableData'
+
+type Props = {
+  assessment: Assessment
+  columns?: Array<string>
+  countryISOs: Array<CountryIso>
+  excludeOdpTable?: boolean
+  mergeOdp?: boolean
```
minotogna: These two parameters should not be passed from outside . They are always true when getting this part published data . 😅

src/server/controller/geo/_getLastPublishedFra1aForestArea.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+
+  const lastYear = Object.keys(tableData)
+    .reduce<Array<number>>((years, colName) => {
+      const year = Number.parseInt(colName, 10)
+      if (Number.isFinite(year)) years.push(year)
+      return years
+    }, [])
+    .sort((a, b) => a - b)
+    .at(-1)
```
minotogna: Why is this needed ?

src/server/controller/cycleData/getLastPublishedData.ts
yaguzmang: Fixed😅

src/server/controller/geo/_getLastPublishedFra1aForestArea.ts
yaguzmang: I didn't see any explicit sorting in the backend, so I think in general the keys are sorted because of how postgres manages jsonbs. I added this here to not to rely on that and get the last year

minotogna: Oh I see 
@yaguzmang to get the col name (the 2025 basically atm ) you should get it from last published cycle (dynamically ) 
Does it make sense ?

src/server/controller/geo/_getLastPublishedFra1aForestArea.ts
```diff
@@ -0,0 +1,54 @@
...
+  const lastPublishedCycleName = Assessments.getLastPublishedCycle(assessment).name
+  const variables = ['forestArea']
+  const tableNames = [TableNames.extentOfForest]
+
+  const data = await CycleDataController.getLastPublishedData(
+    { assessment, countryISOs: [countryIso], tableNames, variables },
+    client
+  )
+
+  const tableData = RecordAssessmentDatas.getTableData({
```
minotogna: And you can use get datum directly

src/server/api/explorer/getData.ts
minotogna: The parameters 

excludeOdpTable: true,
      mergeOdp: true

Should not be passed to get last published data . They are implicitly true . They are passed internally

src/server/controller/geo/_getLastPublishedFra1aForestArea.ts
yaguzmang: Yes that makes sense. Should we use the country last published cycle? or simply the assessment last published cycle

minotogna: the last published cycle. because the published cycles are 2020, 2025, 2030 and the colName exists. 
you cannot use the last published country, since it can be latest and the colName latest doesn't exist.
Does it make sense @yaguzmang ?

src/server/controller/geo/_getLastPublishedFra1aForestArea.ts
```diff
@@ -0,0 +1,57 @@
...
+      tableNames: [tableName],
+      variables: [variableName],
+    },
+    client
+  )
+
+  const rawValue =
+    RecordAssessmentDatas.getDatum({
+      assessmentName,
+      colName: countryLastPublishedCycleName,
```
minotogna: see my comment, you cannot use last published country cycle

yaguzmang: Fixed 👍

minotogna (approved):

PR #5618, opened 2025-12-30
5491 - Admin links job on separate Heroku dyno - Add worker trigger 

yaguzmang:
TODO:
- [x] Move remaining files under worker/tasks/verifyLinks


src/server/start.ts
```diff
@@ -1,6 +1,16 @@
...
 
 serverInit()
+
+if (ProcessEnv.nodeEnv === NodeEnv.development) {
+  void import('./worker/tasks/verifyLinks/triggerVerifyLinksWorker')
+    .then(({ triggerVerifyLinksWorker }) => triggerVerifyLinksWorker())
+    .catch((error) => {
+      Logger.error(`[verifyLinks] failed to start local worker: ${JSON.stringify(error)}`)
+    })
+}
```
yaguzmang, line 16: Starting the verify-links worker automatically so local dev env doesn’t get stuck in a stale state: the server restarts, the worker dies, but Redis still says a worker is “active”. When that happens, new verify requests won’t spin up a worker and the queue just sits there until you manually restart the worker or clear Redis.

We might need something similar for prod, we could use a watchdog mechanism that checks in an interval if there are queued jobs but no dyno worker. Though could be good to test first as it is to see if that's necessary at all.

minotogna: why do we need this in server start ?

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/mergeLinks.ts
```diff
@@ -0,0 +1,25 @@
+import { LinkToVisit } from 'meta/cycleData/links/link'
```
yaguzmang, line 1: Moved this file from `src/server/controller/cycleData/links/visitCycleLinks/utils/mergeLinks.ts`. No logic changes.

src/server/start.ts
yaguzmang: In dev we don’t have the same Heroku lifecycle (enqueue job → start dedicated dyno → dyno consumes queue), we run the worker in the same main process. So I'm starting it here so it cleans up any stale jobs that might happen because of server restarts during development. 

Should we add a separate `start.dev.ts` maybe? what do you think?

minotogna (approved):

PR #5642, opened 2026-01-15
5547 - Reviewer Links Verification process - Cycle api endpoints & requireVerifyLinks middleware

yaguzmang:


src/server/api/cycleData/index.ts
```diff
@@ -20,6 +20,11 @@ import { getDescription } from './descriptions/getDescription'
 import { getDescriptionsHistory } from './descriptions/getDescriptionsHistory'
 import { removeDataSource } from './descriptions/removeDataSource'
 import { upsertDescription } from './descriptions/upsertDescription'
+import { getLinksCount } from './links/getLinksCount'
+import { getManyLinks } from './links/getManyLinks'
+import { isVerificationInProgress } from './links/isVerificationInProgress'
+import { updateLink } from './links/updateLink'
+import { verifyLinks } from './links/verifyLinks'
```
yaguzmang: These endpoints are virtually the same as the Admin ones. Now that links have the country dimension also, maybe the admin endpoints should just use these cycle ones?
or, we keep both and duplicate a bit.

minotogna: Better not to duplicate . The auth middleware should be smart to authorize admin only if country iso is WO , otherwise based on user role

src/server/api/cycleData/index.ts
```diff
@@ -146,6 +151,13 @@ export const CycleDataApi = {
     express.get(ApiEndPoint.CycleData.Review.status(), AuthMiddleware.requireView, getReviewStatus)
     express.get(ApiEndPoint.CycleData.Review.summary(), AuthMiddleware.requireView, getReviewSummary)
 
+    // Country Links
+    express.get(ApiEndPoint.CycleData.Links.count(), AuthMiddleware.requireVerifyLinks, getLinksCount)
+    express.get(ApiEndPoint.CycleData.Links.many(), AuthMiddleware.requireVerifyLinks, getManyLinks)
+    express.patch(ApiEndPoint.CycleData.Links.one(), AuthMiddleware.requireVerifyLinks, updateLink)
+    express.post(ApiEndPoint.CycleData.Links.verify(), AuthMiddleware.requireVerifyLinks, verifyLinks)
+    express.get(ApiEndPoint.CycleData.Links.verifyStatus(), AuthMiddleware.requireVerifyLinks, isVerificationInProgress)
```
minotogna: Can we call the function verify status as well ?

src/server/api/cycleData/index.ts
yaguzmang: Okay, so I will adjust the middleware and then import here the api endpoint from admin, for example 

`src/server/api/cycleData/index.ts:`
```typescript
import { getLinksCount } from 'server/api/admin/links/getLinksCount
...
express.get(ApiEndPoint.CycleData.Links.count(), AuthMiddleware.requireVerifyLinks, getLinksCount)
```
so we reuse the admin functions

minotogna (approved):

src/server/api/cycleData/index.ts
minotogna: @yaguzmang i think we should move back the endpoint under cycle data and use only those . What do you think ?

PR #5653, opened 2026-01-19
5649 - Bulk download - Desk study column

### LINKED ISSUE #5649: Bulk download - Desk study column (High priority)
@minotogna 

**Describe the bug**
The desk study column is missing from the downloaded files. It should be added to each .csv file.

minotogna:


sorja (approved):
Was it added to every file, if so, should we default to true and remove param?

minotogna:
> Was it added to every file, if so, should we default to true and remove param?

I don't know @sorja they keep changing idea, and this way is more flexible

PR #5654, opened 2026-01-19
5547 - Reviewer Links Verification process - Add common LinksTable

yaguzmang:
Adding LinksTable, to be used in the country home and admin.


src/client/pages/AdminLinks/AdminLinks.tsx
```diff
@@ -1,67 +1,14 @@
...
-          className="verify-links-button"
-          disabled={verifyLinksInProgress ?? true}
-          label={t('admin.verifyLinks')}
-          onClick={handleVerifyLinks}
-        />
-      </div>,
-    ],
-    [handleVerifyLinks, t, verifyLinksInProgress]
-  )
+  useInjectSlice({ reducerPath: AdminSliceName, reducer: AdminSlice })
```
yaguzmang, line 10: TODO: Since links aren't used only in admin, we should move the redux definitions, similar to what we did with the api endpoints, admin -> cycle data.

src/client/components/LinksTable/LinksTable.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const verifyLinksInProgress = useIsVerificationInProgress(assessmentName, cycleName, countryIso)
+  useLinksChangeListener({ countryIso, path })
+  useListenLinksVerificationEvents({ countryIso, path })
+
+  useEffect(() => {
+    dispatch(LinksActions.getIsVerificationInProgress({ assessmentName, cycleName, countryIso }))
+  }, [assessmentName, countryIso, cycleName, dispatch])
+
+  const extraActions = useMemo<Array<React.ReactElement>>(
```
minotogna: Better separate hook

src/client/pages/AdminLinks/AdminLinks.tsx
minotogna: Yes and it should be injected with if the LinksTable directly , not here .
We should put it under client / store directly ?

yaguzmang: Yes, in store/links
Can it be next PR? then I will inject it from the LinksTable.

minotogna (approved):

PR #5661, opened 2026-01-22
5647 - Add from Repository: Confirmation message when adding private file

### LINKED ISSUE #5647: Add from Repository: Confirmation message when adding private file
@minotogna 

Many users may not notice the change in status if they do not read carefully.

Remove this sentence (“A file added as reference will become publicly accessible”) from the page below and add it  as a pop-up message when the user selects a private file as a reference. The user must then confirm the change in status from private to public.

Image

sorja:
https://github.com/user-attachments/assets/0d690eac-8d86-438c-a7c5-78bd835ca097

Note: Fixes auth problem when editing (private -> public) repository items for reviewers.

### DIFF
--- a/src/meta/auth/authorizer/canEditRepositoryItem.ts
+++ b/src/meta/auth/authorizer/canEditRepositoryItem.ts
@@ -1,8 +1,17 @@
 import { Country } from 'meta/area/country'
 import { Cycle } from 'meta/assessment/cycle'
 import { canEditCountryProps } from 'meta/auth/authorizer/canEditCountryProps'
+import { Authorizer } from 'meta/auth/authorizer/index'
+import { CollaboratorEditPropertyType } from 'meta/user/role/collaborator'
 import { User } from 'meta/user/user'
+import { Users } from 'meta/user/users'
 
 export const canEditRepositoryItem = (props: { cycle: Cycle; country: Country; user: User }): boolean => {
+  const { country, cycle, user } = props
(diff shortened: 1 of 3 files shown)


src/meta/auth/authorizer/canEditRepositoryItem.ts
```diff
@@ -1,8 +1,17 @@
...
 import { User } from 'meta/user/user'
+import { Users } from 'meta/user/users'
 
 export const canEditRepositoryItem = (props: { cycle: Cycle; country: Country; user: User }): boolean => {
+  const { country, cycle, user } = props
+  if (Users.isReviewer(user, country.countryIso, cycle)) {
+    const permission = CollaboratorEditPropertyType.descriptions
+    return Authorizer.canEditSectionData({ country, cycle, permission, section: undefined, user })
+  }
+
```
sorja, line 19: Notable change in authorizer

minotogna: Why is this needed @sorja ?
Also, please  , update the translations after merging this pr

minotogna: Oh I see . Should this also apply to RFC ?

sorja: I think yes, the RFC should have same rights as reviewer in this case, right?

(Editing descriptions at any country status)

minotogna (approved):

PR #5665, opened 2026-01-23
5547 - Reviewer Links Verification process - Last executed date

yaguzmang:
image

image

### DIFF
--- a/src/client/components/LinksTable/LinksTable.scss
+++ b/src/client/components/LinksTable/LinksTable.scss
@@ -1,3 +1,19 @@
+@use 'src/client/style/partials' as *;
+
+.verify-links {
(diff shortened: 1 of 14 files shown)


src/client/components/LinksTable/LinksTable.scss
```diff
@@ -1,3 +1,19 @@
...
+}
+
 .verify-links-button {
+  white-space: nowrap;
   width: max-content;
 }
+
+.verify-links__last-executed {
+  font-size: $font-xs;
+  font-weight: 600;
```
yaguzmang, line 17: Would look like this without the font-weight:
image

minotogna (approved):

PR #5683, opened 2026-02-02
5547 - Reviewer Links Verification process - Disable export, filters, hide pagination during verification

### LINKED ISSUE #5628: Admin links: when job is running disable filters/csv + remove pagination

yaguzmang:
Resolves #5628 

https://github.com/user-attachments/assets/bf9c8707-9137-475c-81c4-7dd4a7bc5bc5

### DIFF
--- a/src/client/components/LinksTable/LinksTable.tsx
+++ b/src/client/components/LinksTable/LinksTable.tsx
@@ -56,7 +56,6 @@ const LinksTable: React.FC<Props> = (props) => {
       extraActions={extraActions}
       filters={filters}
       gridTemplateColumns="2fr minmax(min-content, 1fr) minmax(264px, 1fr)" // 264px: Location width with padding
-      hidePagination={isVerificationInProgress}
       path={path}
     />
   )
--- a/src/client/components/LinksTable/LinksTable.tsx
+++ b/src/client/components/LinksTable/LinksTable.tsx
@@ -6,6 +6,7 @@ import { CountryIso } from 'meta/area/countryIso'
 
 import { useAppDispatch, useInjectSlice } from 'client/store/hooks'
 import { LinksActions } from 'client/store/links/actions'
+import { useIsVerificationInProgress } from 'client/store/links/hooks/verification'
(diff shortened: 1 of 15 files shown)


src/client/components/LinksTable/LinksTable.tsx
```diff
@@ -46,10 +49,14 @@ const LinksTable: React.FC<Props> = (props) => {
...
     <TablePaginated
       columns={columns}
+      counter={{ show: !isVerificationInProgress }}
+      disableExport={isVerificationInProgress}
+      disableFilters={isVerificationInProgress}
       export
       extraActions={extraActions}
       filters={filters}
       gridTemplateColumns="2fr minmax(min-content, 1fr) minmax(264px, 1fr)" // 264px: Location width with padding
+      hidePagination={isVerificationInProgress}
```
minotogna: But if there's a loader , can we use that logic instead of this prop ?

yaguzmang: The skeleton is shown during verification because when it is in progress we set the data to undefined (using the reducer `resetData`). I could add a new reducer `resetCount` that does the same for the count, and add the necessary logic to avoid refetching it. What do you think?

minotogna: should resetData reset the count at the same time ? I would say so. what do you think ?

src/client/components/TablePaginated/hooks/useFetchData.tsx
```diff
@@ -76,8 +76,8 @@ export const useFetchData = (props: Props): void => {
       const _props = { assessmentName, countryIso, cycleName, filters, limit, orderBy, page, path, sectionName }
       throttledGetData(_props)
 
-      return () => {
-        dispatch(TablePaginatedActions.resetData({ path }))
+      return (): void => {
+        dispatch(TablePaginatedActions.resetData({ path, resetCount: false }))
```
yaguzmang, line 80: Prevents hiding the paginator (and creating layout shift) when, for example, ordering by a column.

minotogna (approved):

PR #5700, opened 2026-02-12
5692 - bulk download: use getLastPublishedData - pt1

sorja:


src/server/controller/cycleData/getBulkDownload/_data.ts
```diff
@@ -38,14 +44,39 @@ const getNames = (props: {
...
+  if (!Cycles.isPublished(cycle)) {
+    return getTableData({ assessment, countryISOs, cycle, mergeOdp: true, tableNames })
+  }
+
+  const tableData = await getLastPublishedData({ assessment, countryISOs, tableNames })
+
+  // Re-key to the requested cycle name for CSV generation
+  const { name: assessmentName } = assessment.props
+  const lastPublishedCycleName = Assessments.getLastPublishedCycle(assessment)?.name
+  if (lastPublishedCycleName && lastPublishedCycleName !== cycle.name) {
```
minotogna: Are you sure this is needed ? 
The getLastPublishedData method already returns the record with only one cycle name (the one you pass from outside )

sorja: We don't pass the cycleName or cycle from outside - so this is true only when the last published cycle is the cycle for which we are fetching data (normally, 2025 now) - but what if we fetch data for 2020? Or when 2030 is released, we fetch for 2025? Is this a realistic scenario?

sorja: @minotogna so in short, should we support more than one published cycle?

minotogna: @sorja maybe we need two different endpoints or a custom parameter to the url . See data export .

sorja: DataExport doesn't use cycleName, it just has assessmentName, then calls getLastPublishedData which internally figures out the last published cycle.

We are not using DataExport for 2020.

So the question remains, should we support one or many published cycles?

sorja: In short - again, if we want to support multiple cycles I think we need to rekey it, if only one published cycle then it should be ok.

We need to do probably same for other data types too (odp, desc)

Let me know what you prefer

sorja: Thanks for clarifying, simplified it now

minotogna (approved):

PR #5722, opened 2026-02-26
5648 - Cycle/Repository: Migration & DDL (1/4)

sorja:


src/tools/migrations/steps/steps/20260225100134-step-move-repository-to-public-schema.ts
```diff
@@ -0,0 +1,108 @@
...
+        // only update props for country-specific items; global items keep props from the oldest cycle
+        if (item.country_iso) acc[item.uuid].props = item.props
+      } else {
+        acc[item.uuid] = { ...item, cycleUuid }
+      }
+    })
+  })
+
+  return Object.values(acc)
+}
```
sorja, line 49: This is the actual core of the migration, all rest is GIGO.

Here we iterate starting from oldest cycle.

Then:

If the item is first met, save the cycle_uuid (latest known cycle uuid for given repository item, e.g. 2020)

Then:

If the repository item is met _again_ , update the props from the new cycle, but don't alter the cycle_uuid (e.g. props from 2025, cycle_uuid still 2020)

This is only TRUE if it's not an global item: global items have the original state from the original cycle.

...meaning that: latest won't have any it's own global items anymore.

-- Some developers notes & thoughts --

...general wondering for future PRs: should we omit cycle_uuid for global items...
...we also have hidden items...

src/tools/migrations/steps/steps/utils/tableExists.ts
```diff
@@ -0,0 +1,18 @@
+import { BaseProtocol } from 'server/db/db'
+
+type Props = {
+  schema: string
+  tableName: string
+}
+
+export const exists = (props: Props, client: BaseProtocol): Promise<{ exists: boolean }> => {
```
minotogna: Should we call this method tableExists ?

src/tools/migrations/steps/steps/20260225100134-step-move-repository-to-public-schema.ts
```diff
@@ -0,0 +1,108 @@
...
+    await _insertItems(resolved, client)
+  }
+
+export default async (client: BaseProtocol): Promise<void> => {
+  // create public table if not exists
+  const { exists } = await tableExists({ schema: 'public', tableName: 'repository' }, client)
+  if (!exists) await DB.none(getRepositoryDDL())
+
+  const assessments = await AssessmentController.getAll({}, client)
+  await Promises.each(assessments, _migrateAssessment(client))
```
minotogna, line 107: Do we remove all cycle repos tables afterwards?

minotogna (commented):
👍

src/tools/migrations/steps/steps/20260225100134-step-move-repository-to-public-schema.ts
minotogna: This looks good @sorja 

I would say that for global items for now let's leave the cycle where they were created , I.e. follow the same algorithm used to migrate country items but we don't group them in the ui . Would that work ?

src/tools/migrations/steps/steps/20260225100134-step-move-repository-to-public-schema.ts
sorja: It will be my last pr (4/4) where I add that so I can do some testing

minotogna: Sure

minotogna (approved):

PR #5726, opened 2026-02-27
5723 - Regional focal point : disable users

### LINKED ISSUE #5723: Regional focal point : disable users
Allow regional focal point to disable users in user form .
He can disable Reviewers , Alternate NC, NC, Collaborators and Viewers

sorja:
kuva

- Added util to create user and users
- Added authorizer src/meta/auth/authorizer/canDisableUser.ts
- TODO: Check why unchecking the box causes bug


src/tools/user/createUsers.ts
```diff
@@ -0,0 +1,22 @@
+import '../scriptInit'
+
+import { RoleName } from 'meta/user/role/name'
+import { ToolsUtils } from 'tools/utils/toolsUtils'
+
+import { createUser } from './createUser'
+
+const countryIso = 'X01'
+
+const USERS = [
```
minotogna, line 10: why this tool? this looks more like a test ?

src/tools/user/createUser.ts
```diff
@@ -0,0 +1,81 @@
...
+import { AssessmentNames } from 'meta/assessment/assessment'
+import { CycleNames } from 'meta/assessment/cycle/names'
+import { RoleName } from 'meta/user/role/name'
+import { ToolsUtils } from 'tools/utils/toolsUtils'
+
+import { passwordHash } from 'server/api/auth/utils/passwordUtils'
+import { AssessmentController } from 'server/controller/assessment'
+import { DB } from 'server/db/db'
+import { Logger } from 'server/utils/logger'
+
```
minotogna, line 12: should this be under src/tools/users/ (generic users folder)

src/tools/user/createUsers.ts
sorja: This is to create multiple users easily to test features on multiple roles - just so I don't need to write createUser multiple times

src/tools/user/createUser.ts
sorja: Sure, there was createUser folder only with createAdmin - I'll move all under user

sorja:
@minotogna
Note: I made the PR so that the user can be re-enabled by RFC:

If RFC disables -> saves -> undisables user - this is ok.
If RFC disables -> saves -> leaves page -> he cannot re-enable

Note that disabled users are not accessibile, so once the RFC _leaves the edit user page_ s/he can obviously not re-enable the user (because s/he cannot find him). Admin is required to re-enable the user in this case.

minotogna (approved):

PR #5737, opened 2026-03-04
5648 - Repository: getMany

sorja:


src/server/controller/cycleData/repository/getMany.ts
```diff
@@ -0,0 +1,28 @@
...
+  countryIso?: AreaCode
+  global: boolean
+}
+
+export const getMany = async (props: Props): Promise<Array<RepositoryItem>> => {
+  const { assessment, countryIso, cycle, global } = props
+
+  const items = await RepositoryRepository.getMany({ assessment, countryIso, global })
+
+  // For global items return [this cycle items] + [previous cycle items]
```
minotogna: Should it be even if not global ? 
And , better to add the filter in the query directly ?

sorja: For countries, the split is done in frontend (all repository items are fetched) tablePaginated component

Regarding countries:

When we are in 2025, should we fetch `latest` and `2020` or just `2020`?

sorja: Maybe I should do in the query something like cycle_uuid in (...)

minotogna: For countries in 2020 we do not want to see latest and 2025.
In 2025 we don't want to see latest .
Can you please check it's the case @sorja ?

sorja: Updated it as:
- Repository takes cycle and finds the previous cycle uuids
- Return repository items based on this for both global and countries

src/server/db/repository/assessmentCycle/repository/getMany.ts
```diff
@@ -1,38 +1,36 @@
...
   const { assessment, countryIso, cycle, global } = props
-  const schemaCycle = Schemas.getNameCycle(assessment, cycle)
-
-  let condition = 'country_iso = $1'
-
-  if (global) {
-    condition = 'country_iso is null'
-  }
+  const assessmentUuid = assessment.uuid
+  const cycleUuids = Cycles.getPreviousCycles({ assessment, cycle }).map((c) => c.uuid)
```
minotogna: Does this method return the cycle uuid passes as props ? If yes , should we rename it to getPreviousAndSelfCycles ?

minotogna (approved):

PR #5747, opened 2026-03-09
5648 - Compare RepositoryItems

sorja:


src/tools/repository/compareItems.ts
```diff
@@ -0,0 +1,159 @@
+import 'tsconfig-paths/register'
+import 'dotenv/config'
+
```
minotogna, line 3: should we put this under tools/repository/compareItems.ts ?

sorja: sure, moved

minotogna (approved):

PR #5749, opened 2026-03-09
5741 - Subregions - Bulk download / API

yaguzmang:
image

Files in slack

----

API Areas:
image

### DIFF
--- a/src/server/controller/cycleData/getBulkDownload/_countries.ts
+++ b/src/server/controller/cycleData/getBulkDownload/_countries.ts
@@ -2,7 +2,6 @@ import { Areas } from 'meta/area/areas'
 import { Country } from 'meta/area/country'
 import { RegionCode } from 'meta/area/regionCode'
 import { RegionGroupName } from 'meta/area/regionGroup'
-import { SubRegionCode } from 'meta/area/subRegionCode'
 import { Lang } from 'meta/lang'
 
 import { AreaRedisRepository } from 'server/cache/repository/area'
@@ -17,17 +16,12 @@ export const getCountries = async (props: PropsBulkDownload): Promise<Array<Coun
   ])
   const fraRegionGroup = regionGroups.find((rg) => rg.name === RegionGroupName.fra2020)
   const allowedRegionCodes = new Set<RegionCode>(fraRegionGroup.regions.map((region) => region.regionCode))
-  const allowedSubregionCodes = new Set<SubRegionCode>(
-    fraRegionGroup.regions.flatMap((region) => region.subRegions?.map((subRegion) => subRegion.regionCode) ?? [])
-  )
 
   return countries
     .reduce<Array<Country>>((acc, country) => {
       if (!Areas.isAtlantis(country.countryIso)) {
         const regionCodes = country.regionCodes.filter((regionCode) => allowedRegionCodes.has(regionCode))
-        const subregionCodes = (country.subregionCodes ?? []).filter((subregionCode) =>
-          allowedSubregionCodes.has(subregionCode)
-        )
+        const subregionCodes = country.subregionCodes ?? []
 
         acc.push({ ...country, regionCodes, subregionCodes })
       }
--- a/src/server/controller/cycleData/getBulkDownload/_countries.ts
+++ b/src/server/controller/cycleData/getBulkDownload/_countries.ts
@@ -1,6 +1,8 @@
 import { Areas } from 'meta/area/areas'
 import { Country } from 'meta/area/country'
+import { RegionCode } from 'meta/area/regionCode'
 import { RegionGroupName } from 'meta/area/regionGroup'
+import { SubRegionCode } from 'meta/area/subRegionCode'
 import { Lang } from 'meta/lang'
 
 import { AreaRedisRepository } from 'server/cache/repository/area'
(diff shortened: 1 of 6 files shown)


src/server/controller/cycleData/getBulkDownload/csvContent/_row.ts
```diff
@@ -35,14 +35,16 @@ export const getCSVRow = (props: Props): CSVRow => {
...
   const { name: cycleName } = cycle
-  const { countryIso, countryIso2, m49, regionCodes } = country
+  const { countryIso, countryIso2, m49, regionCodes, subregionCodes = [] } = country
   const { colForestArea, colNodes, colYear, includeClimaticDomain, includeDeskStudy } = options
 
   const row: CSVRow = []
 
-  const regionLabels = regionCodes.map((code) => getAreaLabel({ code, i18n })).join(',')
+  const regionLabels = (regionCodes ?? []).map((code) => getAreaLabel({ code, i18n })).join(',')
+  const subRegionLabels = subregionCodes.map((code) => i18n.t(`area.${code}.listName`)).join(',')
```
sorja: Is there a reason they are defaulted differently? If not, maybe better to have a single way to set the default value, e.g. in the extraction.

src/server/db/repository/assessmentCycle/region/getRegionGroups.ts
```diff
@@ -24,13 +23,39 @@ export const getRegionGroups = async (
...
                              'regions', jsonb_agg(
-                                     jsonb_build_object('region_code', r2.region_code, 'name', r2.name, 'sort_index', r2.sort_index)
+                                     jsonb_strip_nulls(
+                                       jsonb_build_object(
+                                         'region_code', r2.region_code,
+                                         'name', r2.name,
+                                         'sort_index', r2.sort_index,
+                                         'sub_regions',
+                                         case
+                                           when rg.name = '${RegionGroupName.fra2020}'
```
sorja: Is it needed to be explicit about fra2020 here?

src/server/controller/cycleData/getBulkDownload/csvContent/_row.ts
minotogna: not sure I understood 😅this is where the sub region codes get converted to labels

src/server/controller/cycleData/getBulkDownload/_countries.ts
```diff
@@ -14,14 +16,22 @@ export const getCountries = async (props: PropsBulkDownload): Promise<Array<Coun
     AreaRedisRepository.getManyRegionGroups({ assessment, cycle }),
   ])
   const fraRegionGroup = regionGroups.find((rg) => rg.name === RegionGroupName.fra2020)
-  const allowedRegionCodes = fraRegionGroup.regions.map((r) => r.regionCode)
+  const allowedRegionCodes = new Set<RegionCode>(fraRegionGroup.regions.map((region) => region.regionCode))
+  const allowedSubregionCodes = new Set<SubRegionCode>(
```
minotogna: I think all subregions are allowed at the moment, right ? therefore I would remove the logic of allowed sub regions if all sub regions are allowed. we will add this when and if it will be needed.

src/server/db/repository/assessmentCycle/region/getRegionGroups.ts
minotogna: maybe not

minotogna (commented):
also, do we need to regenerate the area cache ?

src/server/controller/cycleData/getBulkDownload/csvContent/_row.ts
sorja: > not sure I understood 😅this is where the sub region codes get converted to labels

the variable defaults ( `= []` ) are set in two different ways:

```const {  subregionCodes = [] } = country ```

and

```
  const regionLabels = (regionCodes ?? []).map(...)
```

File: src/server/db/repository/assessmentCycle/region/getRegionGroups.ts (hunk 2 above)
Comment (yaguzmang) [reply src/server/db/repository/assessmentCycle/region/getRegionGroups.ts]: Not needed at the moment, since only "fra2020" regions have subregions now, I will remove it 👍 . We just need to keep that in mind if we add any other subregions to the other main regions

**Comment by yaguzmang** [comment]
> also, do we need to regenerate the area cache ?

@minotogna in the migration step `20260306161636-step-add-subregions.ts`, we call the tool `await addCountryRegionLabels(client)` at the end, and in that tool we do:

```typescript
  // update cache
  await Promises.each(allCycles, async ({ assessment, cycle }) => {
    await CacheController.generateArea({ assessment, cycle }, client)
  })
```
So we are covered 👍

src/server/controller/cycleData/getBulkDownload/csvContent/_row.ts
yaguzmang: Yes better to keep it consistent 👍 , thank you!

minotogna:
> > also, do we need to regenerate the area cache ?
> 
> 
> 
> @minotogna in the migration step `20260306161636-step-add-subregions.ts`, we call the tool `await addCountryRegionLabels(client)` at the end, and in that tool we do:
> 
> 
> 
> ```typescript
> 
>   // update cache
> 
>   await Promises.each(allCycles, async ({ assessment, cycle }) => {
> 
>     await CacheController.generateArea({ assessment, cycle }, client)
> 
>   })
> 
> ```
> 
> So we are covered 👍 

Thank you @yaguzmang

minotogna (approved):

PR #5752, opened 2026-03-10
5741 - Fix Admin collaborators : errors on submit

### LINKED ISSUE #5751: Admin collaborators : errors on submit
1. Was logged in as admin, went to the admin page
2. Selected collaborators and found Samy
3. Deselected him as reviewer for Iraq, selected “Submit” and received the following message “Cannot destructure property 'countryIso' of 'country' as it is undefined.”

End result = no changes made in the platform.

yaguzmang:
Fixes issues when changing languages and editing users from non-country routes
image

("User" is there to remove the name from the screenshot)


src/server/api/user/updateUser.ts
```diff
@@ -12,8 +12,8 @@ export const updateUser = async (req: UserEditRequest, res: Response): Promise<v
     const userEditForm = req.body
     const profilePicture = req.file
 
-    const { assessment, country, cycle } = req.context
-    const { countryIso } = country
+    const { assessment, cycle } = req.context
+    const { countryIso } = req.query
```
yaguzmang: @sorja Now extracting from query instead

sorja: I mean should be taken from country as it  was before, now the safe guard should handle it?

yaguzmang: Updated 👍

sorja:
Looks good to me!

minotogna (approved):

PR #5757, opened 2026-03-11
5756 - Accept invitation: Add data protection link

### LINKED ISSUE #5756: Accept invitation: data protection link
In the accept invitation page only, add at the end:

By accepting this invitation, you confirm that you hereby accept the terms and conditions described in the FAO Data Protection and Privacy Notice. [Click here to view the FAO Data Protection and Privacy Notice](https://www.fao.org/contact-us/data-protection-and-privacy/en/).

mockup is here (the text is different though)

Image

yaguzmang:
image

image


minotogna:
> 
> 
> image
> 
> 
> 
> 
> 
> 
> 
> 
> 
> image
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 

@yaguzmang any chance to try to make it fit in two rows (at least in English ) ? What do you think ?

yaguzmang:
> @yaguzmang any chance to try to make it fit in two rows (at least in English ) ? What do you think ?

@minotogna 
Updated:
image

src/client/pages/Authentication/Invitation/Register/Register.tsx
```diff
@@ -73,6 +74,7 @@ const Register: React.FC = () => {
       >
         <Icon name="video" /> {t(videoResources[1].labelKeyShort)}
       </a>
+      <PrivacyNotice />
```
minotogna: Don't the accept and register share the same root to render this only once ?

yaguzmang: Yes those share the same root, however both components `<Register />` and `<Accept />`define the container `<div className="login-form">` within them. So it will need a small refactor. Moving the login-form container div up to `Invitation.tsx` and doing something like this:


image

Let me know if that's prefered, I can make the update

minotogna: Yes definitely.

However don't use the ? : 
But the &&

{Condition && <Component/> }

{!Condition && <Component/> }

yaguzmang: Added 👍

minotogna (approved):

PR #5761, opened 2026-03-12
5750 - Admin Collaborators CSV export: User fields

### LINKED ISSUE #5750: Admin Collaborators CSV export: User fields
In the admin collaborators page, the CSV export should include all user fields.
For the fields to include, see csvAllUsers/index https://github.com/openforis/fra-platform/blob/development/src/tools/csvAllUsers/index.ts

sorja:
Note:

We have now more than one row per user:

We used to have columns for each role, now we destruct the role properties per role - each role gets it's own row

Before:

| name | Role1 | Role2 |
|------|-------|-------|
| Foo Bar | Country1 | Country2 |

Now

| name | Country | Role | ...Role props.... |
|------|-------|-------|---------|
| Foo Bar | Country1 | Role1 | ...Address, etc..|
| Foo Bar | Country2 | Role2 | ...Preferred contact, etc...|


src/tools/csvAllUsers/index.ts
```diff
@@ -18,79 +17,26 @@ const cycleName = '2025'
...
-const _escapeCSVField = (field: unknown): string => {
-  const str = String(field ?? '')
-  // If field contains comma, quote, or newline, wrap in quotes and escape existing quotes
-  if (str.includes(',') || str.includes('"') || str.includes('\n') || str.includes('\r')) {
-    return `"${str.replace(/"/g, '""')}"`
+  const rawRows = await client.any(query, queryParams)
+  const fixQuotes = ([k, v]: [string, string]): Array<string> => {
+    const containsSpecialChars = v.includes(',') || v.includes('"')
+    const fixedV = containsSpecialChars ? `"${v.replace(/"/g, '""')}"` : v
+    return [k, fixedV]
```
sorja: The script (deprecated? should we delet) has been simplified to use the server fn

Some logic from the script was moved to server (e.g. fieldsToRemove, flattening)

minotogna: yes @sorja . let's remove the script as it;s useless now . thank you

minotogna (approved):

PR #5771, opened 2026-03-16
5734 - Email: Use recipient language

### LINKED ISSUE #5734: Email: Use recipient language
Create a generic i18n method to get i18n instance with user language

SomeI18nUtil.getI18n({ user })

sorja:
image

- Clean up I18n
- Test emails (all pre-translated)
- Add missing recipient / lang

### DIFF
--- a/src/i18n/i18nConfig.ts
+++ b/src/i18n/i18nConfig.ts
@@ -1,4 +1,4 @@
-import i18next, { InitOptions, TFunction } from 'i18next'
+import { InitOptions } from 'i18next'
 
 import { Lang } from 'meta/lang'
 
@@ -49,14 +49,3 @@ export const createParams = (lang: Lang): InitOptions => ({
     },
   },
 })
-
-export const createI18nPromise = (lang: Lang): Promise<{ language: Lang; t: TFunction }> =>
-  new Promise((resolve, reject) => {
-    i18next.createInstance(createParams(lang), (err, t) => {
-      if (err) {
-        reject(err)
-        return
-      }
-      resolve({ language: lang, t })
-    })
-  })
--- a/src/server/controller/cycleData/getBulkDownload/getBulkDownload.ts
+++ b/src/server/controller/cycleData/getBulkDownload/getBulkDownload.ts
@@ -1,18 +1,16 @@
-import { i18n as i18nType } from 'i18next'
-
 import { getCountries } from 'server/controller/cycleData/getBulkDownload/_countries'
 import { getData } from 'server/controller/cycleData/getBulkDownload/_getData'
 import { getCSVContentFile } from 'server/controller/cycleData/getBulkDownload/csvContent/getCSVContentFile'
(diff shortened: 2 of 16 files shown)


src/server/utils/i18n.ts
```diff
@@ -0,0 +1,18 @@
...
+  user?: User
+  lang?: Lang
+}
+
+const get = async ({ lang, user }: Props = {}): Promise<i18nType> => {
+  const instance = i18next.createInstance(createParams(user?.props.lang ?? lang ?? Lang.en))
+  await instance.init()
+  return instance
+}
+
```
sorja: This used to be in i18n factory - but only used by server.

Also cleaned the code.

Old implementation for ease of comparison:

```
export const createI18nPromise = (lang: Lang): Promise<{ language: Lang; t: TFunction }> =>
  new Promise((resolve, reject) => {
    i18next.createInstance(createParams(lang), (err, t) => {
      if (err) {
        reject(err)
        return
      }
      resolve({ language: lang, t })
    })
  })
```

sorja: Also - note how we now correctly return the i18nType instead of { lang, t }

src/server/controller/cycleData/getBulkDownload/getBulkDownload.ts
```diff
@@ -1,20 +1,16 @@
...
 import { CSVContent, PropsBulkDownload } from 'server/controller/cycleData/getBulkDownload/types'
+import { I18n } from 'server/utils'
 
 type Props = Omit<PropsBulkDownload, 'i18n'> & { includeClimaticDomain?: boolean }
 
 export const getBulkDownload = async (props: Props): Promise<Array<CSVContent>> => {
   const { assessment, cycle, includeClimaticDomain } = props
 
-  const i18n = (await createI18nPromise(Lang.en)) as i18nType
+  const i18n = await I18n.get({})
```
sorja: Note that because of the refactor, we don't need to cast it anymore incorrectly

src/i18n/i18nConfig.ts
```diff
@@ -49,14 +49,3 @@ export const createParams = (lang: Lang): InitOptions => ({
     },
   },
 })
```
sorja, line 51: Removed code that was only used by server

src/server/controller/cycleData/getBulkDownload/getBulkDownload.ts
minotogna: Should we call the method getInstance?

src/server/utils/index.ts
```diff
@@ -1,2 +1,3 @@
+export { I18n } from './i18n'
```
minotogna: I wouldn't add this to index . We should avoid barrel files

minotogna:
Also @sorja have you checked if all emails are translated in all languages ?

sorja:
> Also @sorja have you checked if all emails are translated in all languages ?

Yes, I think they got "autotranslated" with Crowdin and merged in previous translation PRs.

minotogna (approved):

PR #5777, opened 2026-03-17
5758 - Navigation: Data errors - Validations ContextFactory

yaguzmang:
In a similar fashion as `controller/cycleData/updateDependencies/context/context.ts`, we will have `validations/context/context.ts`, which will produce the affected tables and external node updates through a dependant chain, so we can validate through it.

Later I will add `dataContextBuilder.ts`, which will get the actual table data we can use to validate.

### DIFF
--- a/src/server/controller/cycleData/validations/context/baseContextBuilder.ts
+++ b/src/server/controller/cycleData/validations/context/baseContextBuilder.ts
@@ -1,0 +1,9 @@
+import { ContextBuilderProps } from './contextBuilderProps'
+
+export abstract class BaseContextBuilder {
+  protected readonly props: ContextBuilderProps
+
+  protected constructor(props: ContextBuilderProps) {
+    this.props = props
+  }
+}
--- a/src/server/controller/cycleData/validations/context/context.ts
+++ b/src/server/controller/cycleData/validations/context/context.ts
@@ -1,0 +1,66 @@
+import { CountryIso } from 'meta/area/countryIso'
+import { Assessment, RecordAssessments } from 'meta/assessment/assessment'
+import { Cycle } from 'meta/assessment/cycle'
+import { TableName } from 'meta/assessment/table'
+import { NodeUpdates } from 'meta/data/nodeUpdates'
+import { RecordAssessmentData } from 'meta/data/recordData'
+
+type ConstructorProps = {
+  assessment: Assessment
+  assessments: RecordAssessments
+  countryIso: CountryIso
+  cycle: Cycle
+  data: RecordAssessmentData
+  externalNodeUpdates: Array<NodeUpdates>
+  tableNames: Array<TableName>
+}
+
+export class Context {
+  readonly #assessment: Assessment
+  readonly #assessments: RecordAssessments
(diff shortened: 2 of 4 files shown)


src/server/cache/repository/keys.ts
```diff
@@ -17,6 +17,9 @@ enum Assessments {
 
 enum Data {
   data = 'data',
+  validations = 'data:validations',
+  validationsStale = 'data:validationsStale',
+  validationsSummary = 'data:validationsSummary',
```
minotogna: Why storing stale and summary ?
Should stale be computed on the fly ? 
Have you tried it ?

minotogna: Also ,
Should we start by validating first and then storing it ? I would follow then logical flow to avoid adding unknown code

yaguzmang: Okay, I will start then with the code that actually validates and then we can get back to this redis part 👍

minotogna:
@yaguzmang wouldn't be better to generalize and have one context factory only ?

sorja:
> @yaguzmang wouldn't be better to generalize and have one context factory only ?

True - I would align calculations and validations as close as possible, the only difference is the output type

yaguzmang:
> > @yaguzmang wouldn't be better to generalize and have one context factory only ?
> 
> True - I would align calculations and validations as close as possible, the only difference is the output type

@minotogna @sorja I agree they should stay as close as possible structurally, though I wouldn't merge `updateDependencies/context/contextFactory.ts` with this validation factory as it is at the moment. Both factories are fundamentally different, `updateDependencies` not only has a different output type but also has different traversal logic, row-level context, `includeSourceNodes` / ODP-specific behavior, and row fetching responsibilities.  `validations/context/contextFactory.ts` just gets affected tables and external node updates. If we merged them now, we would likely end up with a more generic factory with multiple flags and branches. That would make the flow harder to follow and review.

If by "generalize" we mean a full re-write of `updateDependencies contextFactory.ts`, then yeah I can look into that, redoing it now considering we also want validation context

minotogna (approved):

minotogna:
> > > @yaguzmang wouldn't be better to generalize and have one context factory only ?
> 
> > 
> 
> > True - I would align calculations and validations as close as possible, the only difference is the output type
> 
> 
> 
> @minotogna @sorja I agree they should stay as close as possible structurally, though I wouldn't merge `updateDependencies/context/contextFactory.ts` with this validation factory as it is at the moment. Both factories are fundamentally different, `updateDependencies` not only has a different output type but also has different traversal logic, row-level context, `includeSourceNodes` / ODP-specific behavior, and row fetching responsibilities.  `validations/context/contextFactory.ts` just gets affected tables and external node updates. If we merged them now, we would likely end up with a more generic factory with multiple flags and branches. That would make the flow harder to follow and review.
> 
> 
> 
> If by "generalize" we mean a full re-write of `updateDependencies contextFactory.ts`, then yeah I can look into that, redoing it now considering we also want validation context 
> 
> 

Ok @yaguzmang well look into that later on maybe . Let's move
On 🙂

PR #5779, opened 2026-03-18
5778 - Disassociate user role

### LINKED ISSUE #5778: Disassociate user role
@minotogna 

**Describe the bug**
The roles in FRA 2025 and the latest FRA are interconnected. Updating the contact in the FRA latest will automatically change the corresponding person in FRA 2025, which is not the intended behavior.

**Example**
Indonesia, Bangladesh and Jamaica as mentioned in the email.

sorja:
image

image


minotogna:
@sorja the confirm message better : 
Are you sure you want to disassociate $user-name as $role from .... ?

sorja:
> @sorja the confirm message better : Are you sure you want to disassociate $user-name as $role from .... ?

Sure, sounds good, here's a screenshot:

image

src/client/pages/User/RemoveRole/hooks/useOnClick.ts
```diff
@@ -0,0 +1,39 @@
...
+  return useCallback(async () => {
+    const name = Users.getFullName(targetUser)
+    const country = t(Areas.getTranslationKey(countryIso))
+    const cycleLabel = t('common.cycleLabel', { assessmentName, cycleName })
+    const countryRole = Users.getRole(targetUser, countryIso, cycle)
+    const role = t(Users.getI18nRoleLabelKey(countryRole.role))
+    // eslint-disable-next-line no-alert
+    if (!window.confirm(t('editUser.disassociateRoleConfirm', { name, country, role, cycleLabel }))) return
+    const params = { assessmentName, cycleName, countryIso, userUuid: targetUser.uuid }
+    await axios.delete(ApiEndPoint.User.role(), { params })
```
minotogna, line 39: Do we want to have a confirm message afterwards: 'User $name successfully disassociated from country assessmentName cyclename' 
What do you think ? Or is it useless ?

src/client/pages/User/User.scss
```diff
@@ -0,0 +1,10 @@
+@use 'src/client/style/partials' as *;
+
+.user-form {
+  max-width: 800px;
+  margin: 0 auto;
+
+  .disassociate-role {
```
minotogna: What about putting a Flex around form and Disassociate to avoid this margin top ?
And the div user-form becomes Flex

src/i18n/resources/en/editUser.json
```diff
@@ -11,6 +11,8 @@
   "deactivate": "Deactivate",
   "demoteToUser": "Are you sure you want to remove Admin privileges?",
   "disabled": "Disabled",
+  "disassociateRole": "Disassociate User Role",
+  "disassociateRoleConfirm": "Are you sure you want to disassociate {{name}} as {{role}} from {{country}} {{cycleLabel}}?",
```
minotogna: Let's not forget the translations afterwards @sorja

src/server/controller/user/disassociateRole.ts
```diff
@@ -0,0 +1,40 @@
...
+  return client.tx(async (t) => {
+    const role = Users.getRole(target, countryIso, cycle)
+    await UserRoleRepository.remove({ uuid: role.uuid }, t)
+
+    // @ts-ignore
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { _: roles, ...userNoRoles } = user
+
+    const activityLog = {
+      target: { user: userNoRoles, countryIso },
```
minotogna: Should we put the deleted role in the target as well? Otherwise we won't know which role was deleted

src/server/middleware/auth/user/disassociateRole.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { CountryRequest } from 'meta/api/request/country'
+import { CountryIso } from 'meta/area/countryIso'
+import { Authorizer } from 'meta/auth/authorizer'
+
+import { UserController } from 'server/controller/user'
+import { _next } from 'server/middleware/auth/_next'
+import { Requests } from 'server/utils'
+
+export const requireDisassociateUserRole = async (
```
minotogna: Since we use the disable user authorizer , should we use the middleware as well ? (And we'll remove this one )

src/client/pages/User/RemoveRole/hooks/useOnClick.ts
sorja: I would almost allow undoing this action - replace the button with "Undo"  after the action?

And also we should show the notification

src/server/controller/user/disassociateRole.ts
sorja: True, now we just saved the user and country without knowing which role he had.

src/server/middleware/auth/user/disassociateRole.ts
sorja: There was no such middleware, that's why this was introduced.

I made it specific to disassociate, so we can make it more generic when there is another similar use to this

What do you think? Do you prefer directly a more generic one? How would you call it?

src/client/pages/User/RemoveRole/hooks/useOnClick.ts
minotogna: no no , let's not undo

src/i18n/resources/en/editUser.json
```diff
@@ -11,6 +11,8 @@
   "deactivate": "Deactivate",
   "demoteToUser": "Are you sure you want to remove Admin privileges?",
   "disabled": "Disabled",
+  "disassociateRole": "Disassociate User Role",
```
minotogna: just spoke with AP. he prefers 'Remove User Role'

sorja: Sure, sounds better (and clearer/easier to understand)

I'll update accordingly ( - will also updated methods / code)

src/i18n/resources/en/editUser.json
```diff
@@ -11,6 +11,9 @@
   "deactivate": "Deactivate",
   "demoteToUser": "Are you sure you want to remove Admin privileges?",
   "disabled": "Disabled",
+  "removeRole": "Remove User Role",
+  "removeRoleConfirm": "Are you sure you want to remove {{name}} as {{role}} from {{country}} {{cycleLabel}}?",
+  "removeRoleSuccess": "User {{name}} successfully removed from {{assessmentName}} {{cycleName}}",
```
minotogna: Country is missing . Without country message can be misleading

sorja: Sure, updated both (confirm, toaster) to use same params now:

```
    const i18nParams = { name, country, role, cycleLabel }

```

minotogna (approved):

PR #5782, opened 2026-03-19
5758 - Navigation: Data errors - Add validations DataContextBuilder

yaguzmang:
Adding `DataContextBuilder`, which can be optionally instantiated from `ContextFactory`.
Now `ContextFactory` supports either getting the tables to validate only, or getting both the tables and the data required to validate them.
The idea is that from `scheduleUpdateDependencies.ts` we only need to know which tables must be validated, so we can invalidate the cache, and no table data is needed there.
From the worker, we will need both the affected tables and the data, so we can run the actual validations.

### DIFF
--- a/src/server/controller/cycleData/validations/context/dataContextBuilder.ts
+++ b/src/server/controller/cycleData/validations/context/dataContextBuilder.ts
@@ -1,0 +1,133 @@
+import { Assessment, AssessmentName, RecordAssessments } from 'meta/assessment/assessment'
+import { Assessments } from 'meta/assessment/assessments'
+import { Cycle, CycleName } from 'meta/assessment/cycle'
+import { VariableCache } from 'meta/assessment/metaCache'
+import { AssessmentMetaCaches } from 'meta/assessment/metaCaches'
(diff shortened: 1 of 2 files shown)


src/server/controller/cycleData/validations/context/dataContextBuilder.ts
```diff
@@ -0,0 +1,133 @@
+import { Assessment, AssessmentName, RecordAssessments } from 'meta/assessment/assessment'
```
yaguzmang, line 1: Based on what was mentioned in previous comments, here it could make sense to add a common `baseDataContextBuilder.ts‎` that manages the assessments/tables registries and data fetching, and we separate the dependencies logic: `updateDependencies` which is variable-driven , `validations` which is table-driven

`src/server/controller/cycleData/updateDependencies/context/`: 
- `baseDataContextBuilder.ts`
- `dataContextBuilder.ts`

`src/server/controller/cycleData/validations/context`:
- `dataContextBuilder.ts` -> extends updateDependencies' `baseDataContextBuilder.ts`

minotogna (approved):

PR #5785, opened 2026-03-21
5758 - Navigation: Data errors - Node level validations Context

yaguzmang:
Now mirroring closer `src/server/controller/cycleData/updateDependencies/context` so we validate at the Node level, not at the table level.


src/server/controller/cycleData/validations/context/contextResult.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export class ContextResult {
+  readonly #setValidations: RecordTableValidationsState
+  readonly #unsetValidations: ValidationUnsetRecord
+
+  constructor() {
+    this.#setValidations = {}
+    this.#unsetValidations = {}
+  }
```
minotogna: What do set and unset represent ?

src/server/controller/cycleData/validations/context/contextFactory.ts
```diff
@@ -180,42 +175,43 @@ export class ContextFactory extends BaseContextBuilder {
...
       data,
       externalNodeUpdates: Array.from(this.#externalNodeUpdatesByCycle.values()),
+      queue: [...this.#queue],
+      rows,
       tableNames: Array.from(this.#tableNames),
     })
   }
 
-  static async collect(props: Props): Promise<Returned> {
+  static async collect(props: ContextBuilderProps): Promise<Returned> {
```
minotogna: Are you ever going to use this  collect method ?

src/server/controller/cycleData/validations/context/contextResult.ts
minotogna: I suspect the result should contain only the entire validations object + tables to update .

yaguzmang: Since we will only cache invalid cells like we said in the meeting, `setValidations` will be used to set in the cache and `unsetValidations` will be used to unset the cells that were invalid before and are valid now

src/server/controller/cycleData/validations/context/contextFactory.ts
yaguzmang: Yes I was planning to use it as a preflight quick check in a method `updateValidations`, but I can do without it 👍

src/server/controller/cycleData/validations/context/contextResult.ts
minotogna: The cache is a record . During the process of validations those results should be set or unset from the main object , not kept in separate objects .
What do you think ?

yaguzmang: This is how I was planning to use them:

1. Traverse updated nodes and dependants to find which ones to evaluate.
2. In a function `validateNodeUpdates` evaluate the nodes, and save the ones to set and unset in the context result (`setValidations, unsetValidations`).
3. In a separate function persist those validations results:

```typescript
export const persistResults = async (props: Props, _client: BaseProtocol = DB): Promise<Returned> => {
  const { context } = props
  const { assessment, countryIso, cycle, tableNames } = context
  const { setValidations, unsetValidations } = context.result // result after evaluating the nodes

  const { tableHasErrors, tableValidations } = await ValidationRedisRepository.applyTableValidationPatch({
    affectedTables: tableNames,
    assessment,
    countryIso,
    cycle,
    setValidations,
    unsetValidations,
  })

  ...
  
```

Perhaps I should rename them `validationsToSet` and `validationsToUnset`.

Or, do you mean that during node evaluation we should update the cache directly? e.g:
```typescript
export const validateNodeUpdates = (props: Props): Context['result'] => {
    // traverse the dependants
    while (context.queue.length !== 0) {
      ...
      const result = ExpressionEvaluator.evalFormula(...)
      if (result.valid === false) {
        // set in cache 
      } else {
        // remove from cache
      }
    }
}
```

Let me know I can make the adjustments 👍

src/server/controller/cycleData/validations/context/context.ts
```diff
@@ -1,38 +1,51 @@
...
 
   constructor(props: ConstructorProps) {
-    const { assessment, assessments, countryIso, cycle, data, externalNodeUpdates, tableNames } = props
+    const { assessment, assessments, country, cycle, data, externalNodeUpdates, queue, rows, tableNames } = props
 
     this.#assessment = assessment
     this.#assessments = assessments
-    this.#countryIso = countryIso
+    this.#country = country
     this.#cycle = cycle
```
sorja, line 40: Question:

Should we pass only assessments, assessmentName and cycleName?

yaguzmang: We need `assessment` and `cycle` at the `contextFactory.ts` level because the builders already use those resolved objects before the final `Context` exists. Inside `Context` we could derive them from `assessments` plus the names, but that would only change the call to `new Context(...)` in `#createContext()`. It would not simplify any other consumer, and `contextFactory.ts` already has them resolved

src/server/controller/cycleData/validations/context/contextResult.ts
minotogna: Yes , during evaluation you update the cache .

yaguzmang: Ok, I removed this file 👍

minotogna (approved):

PR #5787, opened 2026-03-23
5767 - Repository Folders: Backend

sorja:


src/server/db/repository/assessmentCycle/repository/getMany.ts
```diff
@@ -19,18 +18,15 @@ export const getMany = async (props: Props, client: BaseProtocol = DB): Promise<
...
-  }
+  const condition = global ? 'country_iso is null' : 'country_iso = $1'
 
   return client.map<RepositoryItem>(
     `
-      select * from ${schemaCycle}.repository
-      where
-        ${condition} and (props ->> 'hidden')::boolean is not true
-      order by id -- TODO: order by rowIndex
+      select *
```
minotogna: Should we make the hierarchical query ?

sorja: I was thinking of it, but we have so little files it's not really needed - it's really fast in frontend

minotogna: I think with this query , FE has complexity n * m which is to avoid .
Better a hierarchical query here

sorja: Sure

I think I originally tested O(n log n) implementation on FE - which would've been ok for our use case (<1k items/country).

I introduce new endpoint 'getManyTree` which will replace (will delete later, when all features in place) getMany.

sorja: Also doing this in backend not only streamlines the frontend implementation, but will be useful when constructing the zip folder.

👍

src/meta/api/endpoint/cycleData.ts
```diff
@@ -71,6 +71,7 @@ export const CycleData = {
 
     many: (): string => apiPath('cycle-data', 'repository', 'items'),
     one: (): string => apiPath('cycle-data', 'repository', 'items', 'item'),
+    tree: (): string => apiPath('cycle-data', 'repository', 'items', 'tree'),
```
minotogna, line 74: Should we keep using the many ? Anyway I think we shouldn't be using table paginated anymore . And one endpoint only .
What do you think ?

src/server/db/repository/assessmentCycle/repository/getManyTree.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Schemas } from 'server/db/schemas'
+
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+  countryIso?: AreaCode
+  global: boolean
+}
+
+export const getManyTree = async (props: Props, client: BaseProtocol = DB): Promise<Array<RepositoryItemTree>> => {
```
minotogna, line 17: Same as before . One controller / repos method . Which one do you prefer ?

src/meta/api/endpoint/cycleData.ts
sorja: Yes sure - but I would prefer to keep them separate now as some FE is still depending and I want to keep it working.

I can remove it with FE PR - but should we call it tree or just many? `many` is ok (i will rename it after)

src/server/db/repository/assessmentCycle/repository/getManyTree.ts
sorja: I agree - I will rename to getMany(Tree?) as final when no dependencies.
Currently the bulk file download depends on this.

Do you prefer getMany or getManyTree? 

It will return { repositoryItem & children: Array<repositoryItem> }

src/meta/api/endpoint/cycleData.ts
minotogna: Many is ok @sorja . I'll approve this now then

minotogna (approved):

src/meta/api/endpoint/cycleData.ts
sorja: Noted

PR #5788, opened 2026-03-25
5767 - Render tree

sorja:
image

Purpose: Render tree and remove TablePaginated

Next: Add functionality + clean up style


src/client/pages/CountryHome/Repository/RepositoryList/hooks/useItems.ts
```diff
@@ -0,0 +1,28 @@
...
+    ? `${ApiEndPoint.CycleData.Repository.tree()}?global=true`
+    : ApiEndPoint.CycleData.Repository.tree()
+
+  useEffect((): void => {
+    const limit: number = undefined
+    const page: number = undefined
+    dispatch(TablePaginatedActions.getData({ assessmentName, countryIso, cycleName, limit, page, path }))
+  }, [assessmentName, countryIso, cycleName, dispatch, path])
+
+  return useTablePaginatedData<RepositoryItemTree>({ path }) ?? []
```
minotogna, line 27: Better to separate hook to fetch from this .

The getter can be used in the component directly . But I guess we'll need a custom slice ?

sorja: Not to use TablePaginated store at all?
Sure - should I do it in this PR?

minotogna: Or actually we can use just the store . But the getter should be moved in the component directly

minotogna: Or not . That's fine let's see later on

minotogna (approved):

PR #5794, opened 2026-03-25
5758 - Navigation: Data errors - Add validateNodeUpdates

yaguzmang:


src/server/cache/repository/validation/getTableValidations.ts
```diff
@@ -0,0 +1,46 @@
...
+export const getTableValidations = async (props: Props): Promise<RecordTableValidationsState> => {
+  const { assessment, countryIso, cycle, tableNames } = props
+
+  const redis = RedisData.getInstance()
+  const key = getKeyCountry({ assessment, countryIso, cycle, key: Keys.Data.validations })
+
+  if (tableNames?.length) {
+    const values = await redis.hmget(key, ...tableNames)
+
+    return tableNames.reduce<RecordTableValidationsState>((acc, tableName, index) => {
```
minotogna, line 39: Is this reduce needed ? Isn't just a parse enough ?

src/server/cache/repository/validation/getTableValidations.ts
```diff
@@ -0,0 +1,46 @@
...
+      acc[tableName] = _parseTableValidations(values[index])
+      return acc
+    }, {})
+  }
+
+  const tableValidations = await redis.hgetall(key)
+
+  return Object.entries(tableValidations).reduce<RecordTableValidationsState>((acc, [tableName, validations]) => {
```
minotogna, line 48: And here as well

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,64 @@
...
+      }
+      const evalProps = { assessmentName, assessments, colName, country, countryIso, cycleName, data, formula, row }
+      return ExpressionEvaluator.evalFormula<NodeValueValidation>(evalProps)
+    })
+
+    const validation = NodeValueValidations.merge(validations)
+    const validationProps = { assessment, colName, countryIso, cycle, tableName, validation, variableName }
+
+    if (validation.valid) {
+      await ValidationRedisRepository.unsetValidation(validationProps)
```
minotogna: We want to call the repository only at the end and once .

Here we should keep track of the table names + update the validation object

yaguzmang: Okay that's very similar to my approach from before so I can just revert to that. Will upload in a couple of mins.

src/server/cache/repository/validation/getTableValidations.ts
yaguzmang: Since we store the validations at table-level, in redis they are like this:


```typescript 
redisKey = validations:<assessment>:<cycle>:<country>

{
  "table_1": "{\"colA\":{\"var1\":{\"valid\":false}}}",
  "table_2": "{\"colB\":{\"var2\":{\"valid\":false}}}"
}
```


 `hmget(key, 'table_1', 'table_2')` returns:

```typescript
[
  '{"colA":{"var1":{"valid":false}}}',
  '{"colB":{"var2":{"valid":false}}}'
]
```
so we need to parse the result for each table to get:

```typescript
{
  table_1: {...},
  table_2: {...}
}

```

We do something similar in `area/getCountriesMap.ts` for example

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
yaguzmang: Done. Now we patch the tables once at the end 👍

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,112 @@
...
+  context: Context
+}
+
+export const validateNodeUpdates = async (props: Props): Promise<void> => {
+  const { context } = props
+  const { assessment, assessments, country, countryIso, cycle, data, rows, tableNames } = context
+  const { name: assessmentName } = assessment.props
+  const { name: cycleName } = cycle
+  const queue = context.queue.splice(0)
+  const tableValidations = await ValidationRedisRepository.getTableValidations({
```
minotogna: Should this be initialized in the context ?

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,112 @@
...
+    assessment,
+    countryIso,
+    cycle,
+    tableNames,
+  })
+  const touchedTableNames = new Set<TableName>()
+
+  await Promises.each(queue, async (variable: VariableCache) => {
```
minotogna: Maybe you don't need a promise here ?

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,112 @@
...
+      path: [tableName, colName, variableName],
+      value: validation,
+    })
+  })
+
+  if (touchedTableNames.size === 0) {
+    return
+  }
+
+  await ValidationRedisRepository.setTableValidations({
```
minotogna: Should we call the repos outside ? Because also we'll need to return the results and pass the updated tables to clients via websockets .
I would leave this method pure , in the sense it will validate only

src/server/cache/repository/validation/setTableValidations.ts
```diff
@@ -0,0 +1,48 @@
...
+    const tableValidation = tableValidations[tableName] ?? {}
+
+    if (!Objects.isEmpty(tableValidation)) {
+      acc[tableName] = JSON.stringify(tableValidation)
+    }
+
+    return acc
+  }, {})
+
+  const tableNamesToDelete = tableNames.filter((tableName) => Objects.isEmpty(tableValidations[tableName] ?? {}))
```
minotogna: What if we just set an empty object if the validation is valid ?

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
yaguzmang: I could rollback to a previous flow mentioned in the last PR, without `contextResult.ts`:

`validateNodeUpdates.ts` -> returns `setValidations` and `unsetValidations`
with that result, we call outside `ValidationRedisRepository.applyTableValidationPatch` which inside gets the current cached validations and updates them (so we can get rid of line 44 here as well).
From there we can safely return the results which we know are already in cache:

```typescript
const { tableValidations } = await ValidationRedisRepository.applyTableValidationPatch({
    setValidations,
    unsetValidations,
    ...
 })
```

what do you think?

minotogna: Better to update directly the validation object  and return the updated table names .

yaguzmang: Done 👍

src/server/controller/cycleData/validations/context/contextFactory.ts
```diff
@@ -159,8 +160,16 @@ export class ContextFactory extends BaseContextBuilder {
...
+    const { countryIso } = country
+    const tableNames = Array.from(this.#tableNames)
     const { assessments, data } = await this.#dataContextBuilder.getData()
     const rows = await RowRedisRepository.getRows({ assessment, rowKeys: Array.from(this.#rowKeys) })
+    const tableValidations = await ValidationRedisRepository.getTableValidations({
+      assessment,
+      countryIso,
+      cycle,
+      tableNames,
+    })
```
minotogna, line 172: Any chance to use Promise.all ?

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,95 @@
...
+import { Promises } from 'utils/promises'
+
+import { Context } from 'server/controller/cycleData/validations/context/context'
+import { shouldSkipValidationFormula } from 'server/controller/cycleData/validations/shouldSkipValidationFormula'
+
+type RemoveValidationProps = {
+  colName: string
+  tableName: TableName
+  tableValidations: RecordTableValidationsState
+  variableName: string
```
minotogna: We also have types for ColName and VariableName

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,95 @@
...
+export const validateNodeUpdates = async (props: Props): Promise<Array<TableName>> => {
+  const { context } = props
+  const { assessment, assessments, country, countryIso, cycle, data, rows, tableValidations } = context
+  const { name: assessmentName } = assessment.props
+  const { name: cycleName } = cycle
+  const queue = context.queue.splice(0)
+  const touchedTableNames = new Set<TableName>()
+
+  await Promises.each(queue, (variable: VariableCache) => {
+    if (Objects.isNil(variable)) {
```
minotogna: Can this happen ?

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,95 @@
...
+      return
+    }
+
+    const { colName, tableName, variableName } = variable
+
+    if (Objects.isEmpty(colName)) {
```
minotogna: And this ?

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (Objects.isNil(row) || Objects.isNil(col)) {
+      _removeValidation({ colName, tableName, tableValidations, variableName })
+      return
+    }
+
+    if (Objects.isEmpty(validateFns) || Objects.isEmpty(row.props.variableName)) {
+      _removeValidation({ colName, tableName, tableValidations, variableName })
+      return
+    }
+
```
minotogna, line 62: These two remove validation cases above are fishy .

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
yaguzmang: in `VariableCache` colName is defined with 

` colName?: ColName // TODO: will colName become mandatory when handling dependencies by col ?`

So I had added the check in case, but at this point it in the flow it is redundant. Removed it :+1:

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
yaguzmang: Updated to:

- if row/col can't be resolved -> throw an error since metadata/cache is inconsistent
- if validateFns is empty -> remove the cached validation because the node no longer has validation rules

src/server/cache/repository/validation/setTableValidations.ts
yaguzmang: Sure that works too and will help a bit in the FE. Added 👍

src/server/controller/cycleData/validations/validateNodeUpdates/validateNodeUpdates.ts
minotogna: Nice 😊 thank you

minotogna (approved):

PR #5803, opened 2026-03-30
5767 - Repository: Toolbar and Zebra-coloring

sorja:
https://github.com/user-attachments/assets/1ea2e90e-7b94-4f3a-9a7b-f2004154aac5

Scope of PR:
- Add toolbar
- Add zebra

Next PR:
- Sort on click

image


sorja:
I know that the UI is not aligned perfectly (e.g. badges: they are too close to the border...) 
I am planning to do a final UI clean up when all building pieces are in place.

For UI missing: Another toolbar on top with search/filter

Do we need badges? Should it just be text? Even colorless text...?

minotogna (approved):

PR #5804, opened 2026-03-30
5758 - Navigation: Data errors - Update validations

yaguzmang:


minotogna (commented):
@yaguzmang at the beginning we don't want a worker , but to validate the nodes after they get updated

yaguzmang:
> @yaguzmang at the beginning we don't want a worker , but to validate the nodes after they get updated

@minotogna in `src/server/controller/cycleData/updateDependencies/updateDependents.ts`, after running the calculations it will call this worker through `scheduleUpdateValidations(...)` . I can include that in this PR, would that work?

minotogna:
> > @yaguzmang at the beginning we don't want a worker , but to validate the nodes after they get updated
> 
> 
> 
> @minotogna in `src/server/controller/cycleData/updateDependencies/updateDependents.ts`, after running the calculations it will call this worker through `scheduleUpdateValidations(...)` . I can include that in this PR, would that work? 

Even at the next pr @yaguzmang . Let me take another look at this pr 😉

minotogna:
> > @yaguzmang at the beginning we don't want a worker , but to validate the nodes after they get updated
> 
> 
> 
> @minotogna in `src/server/controller/cycleData/updateDependencies/updateDependents.ts`, after running the calculations it will call this worker through `scheduleUpdateValidations(...)` . I can include that in this PR, would that work? 

Do we need a queue @yaguzmang ? We're not using it for the calculations atm ? If not , let's not add this code then

yaguzmang:
> > > @yaguzmang at the beginning we don't want a worker , but to validate the nodes after they get updated
> > 
> > 
> > @minotogna in `src/server/controller/cycleData/updateDependencies/updateDependents.ts`, after running the calculations it will call this worker through `scheduleUpdateValidations(...)` . I can include that in this PR, would that work?
> 
> Do we need a queue @yaguzmang ? We're not using it for the calculations atm ? If not , let's not add this code then

@minotogna `updateDependencies` uses the queue to schedule external dependant updates. Since we wanted to mirror that structure, I added a queue here too to do the same with external validations. What do you think?

minotogna:
> > > > @yaguzmang at the beginning we don't want a worker , but to validate the nodes after they get updated
> 
> > > 
> 
> > > 
> 
> > > @minotogna in `src/server/controller/cycleData/updateDependencies/updateDependents.ts`, after running the calculations it will call this worker through `scheduleUpdateValidations(...)` . I can include that in this PR, would that work?
> 
> > 
> 
> > Do we need a queue @yaguzmang ? We're not using it for the calculations atm ? If not , let's not add this code then
> 
> 
> 
> @minotogna `updateDependencies` uses the queue to schedule external dependant updates. Since we wanted to mirror that structure, I added a queue here too to do the same with external validations. What do you think? 

Those are external dependencies. 
We need to validate after the normal flow . It will a synchronous process . 
I doubt we need a queue now @yaguzmang 
After updating dependencies , we need to validate

src/server/controller/cycleData/updateDependencies/updateDependents.ts
yaguzmang: Can a node have validation dependants while not having any calculation dependants? if so, we need to call update validation before this early return as well.

yaguzmang:
@minotogna PR updated based on comments 👍

src/server/controller/cycleData/updateDependencies/updateDependents.ts
minotogna: True . Can you investigate the best place where to run the validations on value update ?

yaguzmang: Checked redis and yes, there are multiple examples
- `fra-2025`: `disturbances.other` validates `insects`, `diseases`, `severe_weather_events`, `other`
- `fra-2025`: `contactPersons.expectedYearForNextCountryReportUpdate` validates `reportLastUpdate.reportLastUpdate`
etc.

Added a call to `updateValidations` before resolving when there are no calculation dependants 👍, now both cases covered.
This is still the best place to call the validations

src/server/controller/cycleData/updateDependencies/updateDependents.ts
```diff
@@ -16,18 +16,20 @@ type Props = Omit<UpdateDependenciesProps, 'client'> & {
...
-  const dependants = nodes.flatMap(({ tableName, variableName }) => {
-    return AssessmentMetaCaches.getCalculationsDependants({ assessment, cycle, tableName, variableName })
+  // avoid executing worker when nodes have no calculation dependents.
+  const hasCalculationDependants = nodes.some(({ tableName, variableName }) => {
+    return AssessmentMetaCaches.getCalculationsDependants({ assessment, cycle, tableName, variableName }).length > 0
   })
-  if (dependants.length === 0 && !includeSourceNodes) {
+  if (!hasCalculationDependants && !includeSourceNodes) {
+    // nodes with no calculation dependents may have validation dependents
+    await updateValidations({ assessment, country, cycle, nodeUpdates: props.nodeUpdates, notifyClients })
```
minotogna: It is best if we call this only once , because after there will be socket events to emit as well , it's better not to duplicate

yaguzmang: I edited the function structure a bit so it's easy to follow and so we call this once 👍

src/server/controller/cycleData/updateDependencies/updateDependents.ts
```diff
@@ -1,57 +1,69 @@
...
   // 3. schedule external assessment/cycle updates
-  await Promises.each(externalDependants, async (externalNodeUpdates) => {
-    Logger.debug(
-      `[updateDependencies] [job-${job.id}] scheduling ${externalNodeUpdates.nodes.length} external dependents`
-    )
+  await Promises.each(workerResult.externalDependants, async (externalNodeUpdates) => {
+    Logger.debug(`[updateDependencies] [${logKey}] scheduling ${externalNodeUpdates.nodes.length} external dependents`)
     await updateExternalDependents({ countryIso, nodeUpdates: externalNodeUpdates, notifyClients, user }, client)
   })
+  // TODO: Update external validations
```
minotogna, line 66: @yaguzmang this is not a use case at the moment and it can be skipped for simplicity

minotogna (approved):

PR #5808, opened 2026-03-31
5767 - Repository/Filters: - add filter by name

sorja:
https://github.com/user-attachments/assets/c12b8929-2d84-43b5-8490-b0d972276b07

Scope of PR:
- Add filter by name
- Split useItems to smaller components 

Next PR:
- Add other filters

### DIFF
--- a/src/client/pages/CountryHome/Repository/Repository.tsx
+++ b/src/client/pages/CountryHome/Repository/Repository.tsx
@@ -18,17 +18,13 @@ const Repository: React.FC = () => {
     <div className="repository">
       {!isPanEuropean && (
         <>
-          <div className="repository__header">
-            <h3>{t('landing.links.links')}</h3>
-          </div>
+          <h3>{t('landing.links.links')}</h3>
           <RepositoryList isGlobal />
           <Hr />
         </>
       )}
 
-      <div className="repository__header">
-        <h3>{t('landing.links.repository')}</h3>
-      </div>
+      <h3>{t('landing.links.repository')}</h3>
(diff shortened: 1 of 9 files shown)


minotogna:
On the left side , let's use a filter icon ,  not sort icon @sorja

minotogna:
Also , let's remove the download all from title

sorja:
> Also , let's remove the download all from title

About the title: Should we leave it as it is or align left/something else?

For now, I'll remove & fix the icon

minotogna:
> > Also , let's remove the download all from title
> 
> 
> 
> About the title: Should we leave it as it is or align left/something else?
> 
> 
> 
> For now, I'll remove & fix the icon

Try to align the title to the left , maybe it's better

sorja:
> > > Also , let's remove the download all from title
> > 
> > 
> > About the title: Should we leave it as it is or align left/something else?
> > For now, I'll remove & fix the icon
> 
> Try to align the title to the left , maybe it's better

image


Sure - either way let's leave like this and clean up in another PR in case

sorja:
@minotogna I will align in future PR the sorting icon in the table header to the right of the cell, what do you think? Is it too far in the first column?

minotogna:
> @minotogna I will align in future PR the sorting icon in the table header to the right of the cell, what do you think? Is it too far in the first column?

I think it's too far @sorja

minotogna (commented):
@sorja please check current pan European as well .
Behavior should not change

minotogna (approved):

PR #5815, opened 2026-04-01
5811 - Dependencies upgrade - eslint

### LINKED ISSUE #5811: Dependencies Upgrade: eslint & co
eslint 9.39.2 -> 10.1.0
globals 16.5.0 -> 17.4.0 
typescript-eslint 8.50.0 -> 8.58.0

yaguzmang:
- `no-useless-assignment` -> new ESLint 10 core rule -> flags placeholder assignments like `let x = ''` or `let asset = {}` when that value is always overwritten before being used.
- `preserve-caught-error` -> new ESLint 10 core rule -> flags `throw new Error(...)` inside `catch` blocks unless the original error is preserved with `new Error(message, { cause: error })`.
- `@typescript-eslint/no-var-requires` -> deprecated/outdated `typescript-eslint` rule -> the relevant rule is now `@typescript-eslint/no-require-imports`.


minotogna (approved):
💪

PR #5818, opened 2026-04-02
5767 - Repository: Filter by linked, unlinked public and private

sorja:
image

Scope of current PR:
- Fix panEuropean repository (odp)
- implement all column filters (expect date created)

Next PR:
- Filter by file type

### DIFF
--- a/src/client/pages/CountryHome/Repository/RepositoryList/Filters/Filters.scss
+++ b/src/client/pages/CountryHome/Repository/RepositoryList/Filters/Filters.scss
@@ -18,9 +18,5 @@
   .filter-multiselect__tooltip-trigger,
   .table-paginated-filter-input {
     height: 30px;
-
-    div.select__control {
-      min-height: 30px;
-    }
   }
 }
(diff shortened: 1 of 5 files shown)


sorja:
Screenshot has an example selection of filters, by default none are selected

minotogna:
@sorja shiuld linked / unlinked and public / private be two selects ?

sorja:
@minotogna sure, single/multiselect?

src/client/pages/CountryHome/Repository/RepositoryList/Filters/Filters.scss
```diff
@@ -15,7 +15,12 @@
     width: 1px;
   }
 
+  .filter-multiselect__tooltip-trigger,
   .table-paginated-filter-input {
     height: 30px;
+
+    div.select__control {
+      min-height: 30px;
```
minotogna: Do we need this ? Or is there a size param we can pass to the select ?

sorja: Not needed

Adding the selector in the above rule fixed it, this was redundant

minotogna (approved):

PR #5822, opened 2026-04-03
5767 - Repository: Folder navigation with breadcrumbs

sorja:
image

image

Scope of this PR:
- Folder navigation with breadcrumbs
- Create files under folders (parent_uuid passed)

Next PR:
- Finalize UI
  - Update the toolbar bar with back button
  - Style the toolbar
  - Order components (toolbar)

Still coming (for UI):
- use Form in the panel

Rest can be found in the main PR

### DIFF
--- a/src/client/pages/CountryHome/Repository/RepositoryList/RepositoryList.tsx
+++ b/src/client/pages/CountryHome/Repository/RepositoryList/RepositoryList.tsx
@@ -23,6 +23,7 @@ const RepositoryList: React.FC<Props> = (props) => {
   return (
     <div className="repository-list">
       <Filters isGlobal={isGlobal} />
+      <Filters isGlobal={isGlobal} parentUuid={parentUuid} />
       <Toolbar isGlobal={isGlobal} />
       {items.map((item) => (
         <RepositoryListItem key={item.uuid} collapsed={collapsed} item={item} onToggle={onToggle} />
--- a/src/client/pages/CountryHome/Repository/RepositoryList/RepositoryList.tsx
+++ b/src/client/pages/CountryHome/Repository/RepositoryList/RepositoryList.tsx
@@ -3,9 +3,12 @@ import React from 'react'
 
 import RepositoryListItem from 'client/pages/CountryHome/Repository/RepositoryList/RepositoryListItem'
 
+import { useFolderNavigation } from './hooks/useFolderNavigation'
 import { useGetItems } from './hooks/useGetItems'
 import { useItems } from './hooks/useItems'
 import { useOnToggle } from './hooks/useOnToggle'
+import Breadcrumb from './Breadcrumb'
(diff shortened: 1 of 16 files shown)


src/client/pages/CountryHome/Repository/RepositoryList/RepositoryList.tsx
```diff
@@ -19,15 +22,19 @@ const RepositoryList: React.FC<Props> = (props) => {
...
 
   return (
-    <div className="repository-list">
-      <Filters isGlobal={isGlobal} />
-      <Toolbar isGlobal={isGlobal} />
-      {items.map((item) => (
-        <RepositoryListItem key={item.uuid} collapsed={collapsed} item={item} onToggle={onToggle} />
-      ))}
-    </div>
+    <RepositoryListContext.Provider value={contextValue}>
```
minotogna, line 28: How often do these values change ? If often , this is not optimal

src/client/pages/CountryHome/Repository/RepositoryList/hooks/useFolderNavigation.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+type Returned = {
+  contextValue: RepositoryListContextValue
+  visibleItems: Array<RepositoryItemTree>
+}
+
+export const useFolderNavigation = (props: Props): Returned => {
+  const { collapsed, items, onToggle } = props
+  const [folderTarget, setFolderTarget] = useState<string | null>(null)
```
minotogna: Null is a value in JavaScript. Should this be undefined ?

src/client/pages/CountryHome/Repository/RepositoryList/RepositoryList.tsx
sorja: What do you mean by not optiomal?

`contextValue` is wrapped in use memo so it should be ok

Context was introduced to avoid prop drilling and to share the state inside the component tree

minotogna: If the context values updates often , it is not optimal to use context , that's why my question .
Context are inefficient

sorja: The context rerenders affect only the subtree and the updates happen only on user interaction (=rare)

So in this case I would say it's ok to have it like this, if you prefer, I can always split the context to stable values (onNavigate, onToggle) and for the more frequently changing (e.g. collapsed, folderPath and parentUuid)?

But having 2 contexts seem overkill for so small component

What do you think/prefer?

sorja: tldr: updates are user triggered and rare

minotogna (approved):

PR #5827, opened 2026-04-08
5758 - Navigation: Data errors - Update summary on table validations update

yaguzmang:


src/client/pages/Country/hooks/useTableValidationsListener.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useAppDispatch } from 'client/store/hooks'
+import { useCountryRouteParams } from 'client/hooks/routeParams'
+import { SocketClient } from 'client/service/socket/client'
+
+export const useTableValidationsListener = (): void => {
+  const { assessmentName, countryIso, cycleName } = useCountryRouteParams<CountryIso>()
+  const dispatch = useAppDispatch()
+
+  useEffect(() => {
+    const eventName = Sockets.getNodeValidationsUpdateEvent({ countryIso, assessmentName, cycleName })
```
minotogna: We forgot here and in the summary , only users with edit access can fetch and subscribe to all validations

src/client/store/data/tableData/validations/slice/extraReducers/setSummaryReducer.ts
```diff
@@ -0,0 +1,41 @@
+import { ActionReducerMapBuilder } from '@reduxjs/toolkit'
+
+import { Objects } from 'utils/objects'
+
+import { setNodeValueValidations } from 'client/store/data/tableData/validations/actions/setNodeValueValidations'
+import { ValidationsState } from 'client/store/data/tableData/validations/state'
+
+export const setSummaryReducer = (builder: ActionReducerMapBuilder<ValidationsState>): void => {
```
minotogna: Should we call this updateSummaryReucer ?

src/client/pages/Country/hooks/useTableValidationsListener.tsx
minotogna: Should we call this tableValidations instead of node validations ?

src/server/controller/cycleData/validations/updateValidations.ts
```diff
@@ -30,4 +35,14 @@ export const updateValidations = async (props: Props): Promise<void> => {
     tableNames: updatedTableNames,
     tableValidations,
   })
+
+  if (notifyClients && updatedTableNames.length > 0) {
+    const eventName = Sockets.getNodeValidationsUpdateEvent({ assessmentName, countryIso, cycleName })
+    const updatedTableValidations = updatedTableNames.reduce<RecordTableValidationsState>((acc, tableName) => {
```
minotogna, line 41: Why this mapping ? Shouldn't we just send the result of the validation?

src/client/pages/Country/hooks/useInitValidationSummary.tsx
```diff
@@ -4,13 +4,17 @@ import { CountryIso } from 'meta/area/countryIso'
...
 
 export const useInitValidationSummary = (): void => {
   const { assessmentName, countryIso, cycleName } = useCountryRouteParams<CountryIso>()
+  const canEditData = useCanEditCycleData()
   const dispatch = useAppDispatch()
 
   useEffect(() => {
+    if (!canEditData) return
+
     dispatch(ValidationsActions.getSummary({ assessmentName, cycleName, countryIso }))
```
minotogna: On return we should remove the validations

src/client/pages/Country/hooks/useValidations.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    const listener = (args: [{ tableValidations: RecordTableValidationsState }]): void => {
+      const [{ tableValidations }] = args
+      dispatch(ValidationsActions.setNodeValueValidations({ assessmentName, cycleName, countryIso, tableValidations }))
+    }
+
+    SocketClient.on(eventName, listener)
+
+    return (): void => {
+      SocketClient.off(eventName, listener)
```
minotogna, line 39: On return we should remove the validations

src/server/controller/cycleData/validations/updateValidations.ts
yaguzmang: We have the updatedTableNames from 
`const updatedTableNames = await validateNodeUpdates({ context })` and the tableValidations from `context`.

So here we are building the object with those two, using only revalidated tables, which the FE can use. I could change `validateNodeUpdates({ context })` so it returns that shape already too... let me know

src/client/pages/Country/hooks/useInitValidationSummary.tsx
```diff
@@ -4,13 +4,21 @@ import { CountryIso } from 'meta/area/countryIso'
...
-  }, [assessmentName, countryIso, cycleName, dispatch])
+
+    return (): void => {
+      dispatch(ValidationsActions.removeValidations({ assessmentName, cycleName, countryIso }))
```
minotogna: Is this the same action dispatched in the table validations hook ?
Would it make sense to dispatch it only once and maybe merge these two hooks ?

src/client/store/data/tableData/validations/slice/index.ts
```diff
@@ -1,6 +1,8 @@
 import { createSlice } from '@reduxjs/toolkit'
 
 import { getSummaryReducer } from 'client/store/data/tableData/validations/slice/extraReducers/getSummaryReducer'
+import { removeValidationsReducer } from 'client/store/data/tableData/validations/slice/extraReducers/removeValidationsReducer'
+import { updateSummaryReducer } from 'client/store/data/tableData/validations/slice/extraReducers/setSummaryReducer'
```
minotogna: Should we rename the file as well ?

src/server/controller/cycleData/validations/updateValidations.ts
minotogna: Right 👍

minotogna (commented):
Also , the tests do not pass

src/client/pages/Country/hooks/useInitValidationSummary.tsx
yaguzmang: Yes that makes sense, added 👍

minotogna (approved):

PR #5833, opened 2026-04-10
5758 - Navigation: Data errors - Make validators be language agnostic - 1

yaguzmang:
Next PR will introduce the actual parser that uses the new Node message param `validatorName`. Will look something like this (with other helpers):

```typescript
export const translateValidationMessage = (t: TFunction, message: NodeValueValidationMessage): string => {
  const { key, params, validatorName } = message

  if (Objects.isEmpty(params)) {
    return t(key)
  }

  if (validatorName === validatorSumEqualTo.name) {
    return t(key, getTranslatedValidatorSumEqualToParams(t, params as ValidatorSumEqualToParams))
  }

  if (
    validatorName === validatorSumSubCategoriesNotEqualToParent.name ||
    validatorName === validatorSumSubCategoriesNotGreaterThanParent.name
  ) {
    return t(key, getTranslatedValidatorSumSubCategoriesParams(t, params as ValidatorSumSubCategoriesParams))
  }

  return t(key, translateParams(t, params))
}
```


src/meta/expressionEvaluator/functions/subcategories/utils.ts
```diff
@@ -19,32 +17,3 @@ export const calculateCategoriesSum = (
...
-      key: `generalValidation.${label}`,
-      params: {
-        parentLabel,
-        parentValue: Numbers.format(Number(parentValue)),
-        categoriesSum: Numbers.format(categoriesSum),
-        categoryLabels: categoryLabelKeys.map((labelKey) => t(labelKey)).join(', '),
-      },
-    },
-  ]
-}
```
yaguzmang, line 50: This helper becomes irrelevant because we will have one parser handling this and returning the translation

src/meta/expressionEvaluator/functions/subcategories/utils.ts
```diff
@@ -1,50 +1,8 @@
...
-    (acc: Array<CategoryInfo>, value: string | undefined, currentIndex: number) => {
-      if (!Objects.isEmpty(value)) acc.push({ labelKey: categoryLabelKeys[currentIndex], value })
-      return acc
-    },
-    []
-  )
-
-  return Numbers.sum(nonEmptyCategories.map(({ value }) => value))
-}
-
```
yaguzmang, line 22: Removing this acc that was pushing `labelKey`s and didn't use them afterward

src/meta/assessment/nodeValueValidation/index.ts
```diff
@@ -1,6 +1,7 @@
 export type NodeValueValidationMessageParam = string | number | Array<string> | Array<number>
 
 export interface NodeValueValidationMessage {
+  validatorName: string
```
minotogna: Can this just be name ?

src/meta/expressionEvaluator/functions/subcategories/validatorSumSubCategoriesNotEqualToParent.ts
```diff
@@ -32,19 +31,24 @@ export const validatorSumSubCategoriesNotEqualToParent: ExpressionFunction<Conte
...
-        parentLabelParams,
-        parentColLabelKey
-      )
-
-      return { valid, messages }
+      return {
+        valid,
+        messages: [
+          {
+            validatorName: validatorSumSubCategoriesNotEqualToParent.name,
```
minotogna: Should we introduce an enum ? ( and use it also in the name , e.g. see line 11 this file )

yaguzmang: Yes, added 👍

minotogna (approved):

PR #5835, opened 2026-04-13
5767 - Repository: Add and Edit Folder

sorja:
Edit folder:

Display 'Edit {folderName}' tooltip

image

Display 'Edit item' tooltip'
image

PR Scope:
- Add and edit folders
- (Omitted styling edits in this PR)

Next PR:
- Finish styling (form, breadscrumbs etc)

After:
- Modal


src/i18n/resources/en/common.json
```diff
@@ -3,6 +3,9 @@
   "added": "Added",
   "accepted": "Accepted",
   "add": "Add",
+  "addFolder": "Add folder",
+  "editFolder": "Edit {{name}}",
+  "editItem": "Edit item",
```
minotogna: Should we have just one edit label ? Why folder with name and item ? Just 'Edit' for both ?

Also , can we make lighter the hover color ?

@sorja

sorja: Hmm, just wanted a way to distinguish on hover between the edit for folder and edit for items, but maybe not needed?

sorja: image

I think it's better like you said

minotogna (approved):

PR #5837, opened 2026-04-13
5758 - Navigation: Data errors - Make validators be language agnostic - 2

yaguzmang:


src/meta/validations/translateValidationMessage.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  return {
+    categoryLabels: categoryLabelKeys.map((labelKey) => t(labelKey)).join(', '),
+    categoriesSum: _translateParam(t, categoriesSum),
+    parentLabel,
+    parentValue: _translateParam(t, parentValue),
+  }
+}
+
+export const translateValidationMessage = (t: TFunction, message: NodeValueValidationMessage): string => {
```
minotogna: I would create a 
Src/meta/validations/MessageParser.getMessage

And sub files for specific validations  parser 
This way if we add custom messages , this file won't grow

yaguzmang: Added 👍

minotogna (approved):

PR #5838, opened 2026-04-13
5758 - Navigation: Data errors - New Validation cache types - 1

yaguzmang:
New PR scope:
- Removing cross-assessment dependencies handler because that doesn't exist for validations like they do for node updates.

- Introducing new type 
```ts
export type ValidationTargetsBySource = Record<
  TableName,
  Record<VariableName, Record<ColName, Array<ValidationTargetCache>>>
>
```

The idea is that the record itself is the validation source, and it contains an array of targets.
For example, for cell `extentOfForest`.`forestArea`.`2025`, the dependencies are
- `forestAreaChange`.`forestAreaNetChange`.`2020-2025`
- `forestAreaWithinProtectedAreas `.`forest_area_within_protected_areas `.`2025`, 
- and `forestAreaWithinProtectedAreas`.`forest_area_with_long_term_management_plan`.`2025`.

so the `ValidationTargetsBySource` object here would be:

```ts
{
  extentOfForest: {
    forestArea: {
      '2025': [
        {
          tableName: 'forestAreaChange',
          variableName: 'forestAreaNetChange',
          colName: '2020-2025',
        },
        {
          tableName: 'forestAreaWithinProtectedAreas',
          variableName: 'forest_area_within_protected_areas',
          colName: '2025',
        },
        {
          tableName: 'forestAreaWithinProtectedAreas',
          variableName: 'forest_area_with_long_term_management_plan',
          colName: '2025',
        },
      ],
    },
  },
}
```

The entire type is:
```ts
export type ValidationDependencyCache = {
  dependencies: DependencyRecord
  dependants: ValidationTargetsBySource
}
```

Dependencies are still kept as `DependencyRecord` because their current consumers only need to know what source variables/tables/cycles must be loaded, not which exact target cell triggered them.

### DIFF
--- a/src/meta/assessment/metaCache/index.ts
+++ b/src/meta/assessment/metaCache/index.ts
@@ -28,16 +28,7 @@ export type VariablesCache = Record<TableName, Record<VariableName, VariableCach
 export type TableDependencyRecord = Record<VariableName, Array<VariableCache>>
 export type DependencyRecord = Record<TableName, TableDependencyRecord>
 
-export type ValidationTargetCache = {
-  colName: ColName
-  tableName: TableName
-  variableName: VariableName
-}
-
-export type ValidationTargetsBySource = Record<
-  TableName,
-  Record<VariableName, Record<ColName, Array<ValidationTargetCache>>>
->
+export type ValidationTargetsBySource = Record<TableName, Record<VariableName, Record<ColName, Array<VariableCache>>>>
 
 export type DependencyCache = {
   dependencies: DependencyRecord
--- a/src/meta/assessment/metaCache/index.ts
+++ b/src/meta/assessment/metaCache/index.ts
@@ -28,14 +28,30 @@ export type VariablesCache = Record<TableName, Record<VariableName, VariableCach
 export type TableDependencyRecord = Record<VariableName, Array<VariableCache>>
 export type DependencyRecord = Record<TableName, TableDependencyRecord>
 
+export type ValidationTargetCache = {
+  colName: ColName
+  tableName: TableName
+  variableName: VariableName
+}
+
+export type ValidationTargetsBySource = Record<
+  TableName,
+  Record<VariableName, Record<ColName, Array<ValidationTargetCache>>>
+>
+
 export type DependencyCache = {
   dependencies: DependencyRecord
   dependants: DependencyRecord
 }
 
+export type ValidationDependencyCache = {
+  dependencies: DependencyRecord
+  dependants: ValidationTargetsBySource
+}
+
 export interface AssessmentMetaCache {
   calculations: DependencyCache
-  validations: DependencyCache
+  validations: ValidationDependencyCache
   enablers?: DependencyCache
   variablesByTable: VariablesCache
 }
--- a/src/server/controller/cycleData/validations/context/contextFactory.ts
+++ b/src/server/controller/cycleData/validations/context/contextFactory.ts
@@ -1,6 +1,7 @@
+import { ColName } from 'meta/assessment/col'
 import { VariableCache } from 'meta/assessment/metaCache'
 import { AssessmentMetaCaches } from 'meta/assessment/metaCaches'
-import { RowCacheKey } from 'meta/assessment/rowCache'
+import { RowCache, RowCacheKey } from 'meta/assessment/rowCache'
 import { RowCaches } from 'meta/assessment/rowCaches'
 import { TableName } from 'meta/assessment/table'
 import { NodeUpdate, NodeUpdates } from 'meta/data/nodeUpdates'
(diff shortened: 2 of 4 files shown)


src/server/controller/cycleData/validations/context/contextFactory.ts
```diff
@@ -109,25 +148,23 @@ export class ContextFactory extends BaseContextBuilder {
...
         cycleName: dependant.cycleName ?? cycle.name,
-        colName: dependant.colName ?? variable.colName,
+        colName: dependant.colName,
         tableName: dependant.tableName,
         variableName: dependant.variableName,
-      }
-
-      if (Objects.isEmpty(candidate.colName)) {
-        return
-      }
```
yaguzmang, line 122: Instead of returning early here we should get the entire row cols, so the validations behave the same as in `useValidate.ts`

src/server/controller/cycleData/validations/context/dataContextBuilder.ts
```diff
@@ -40,44 +37,33 @@ export class DataContextBuilder extends BaseContextBuilder {
...
+        tableNames: new Set<TableName>(),
+      }
     }
 
-    return this.#assessments[assessmentName]
-  }
-
-  async #ensureTablesFetch(assessmentName: AssessmentName, cycleName: CycleName): Promise<TablesFetch> {
-    if (!this.#tables[assessmentName]?.[cycleName]) {
-      // Validation dependencies may point to another assessment/cycle
```
yaguzmang, line 63: Removing this since cross-assessment dependencies don't exist for validations like they do for node updates.

src/server/controller/cycleData/validations/context/contextFactory.ts
```diff
@@ -54,6 +55,44 @@ export class ContextFactory extends BaseContextBuilder {
     return assessmentName === this.props.assessment.props.name && cycleName === this.props.cycle.name
   }
 
+  async #getRow(variable: VariableCache): Promise<RowCache | undefined> {
```
minotogna: why do we need this ?

yaguzmang: In the soon to be deleted `useValidate` we  go over the validation like this:
```typescript
rowsData.forEach((row) => {
  const { variableName } = row.props
  row.cols.forEach((col) => {
    const validateFns = col.props.validateFns?.[cycle.uuid] ?? row.props.validateFns?.[cycle.uuid]

    if (validateFns?.length) {
      const { colName } = col.props
      // evaluate and store validation at [tableName, colName, variableName]
    }
  })
})
```

```
const validateFns = col.props.validateFns?.[cycle.uuid] ?? row.props.validateFns?.[cycle.uuid]
```

There if the col doesn't have a validateFns, we use the row validateFns. And we go over all cols. 
Here in the new backend, some validation dependants arrive without a colName. That happens when they come from a row-level validation rather than a specific column. In that case, we need the row so we can expand that dependant into the row’s columns and queue one validation target per colName, the same way useValidate is doing on the frontend.

minotogna: in the current use cases, we either validate a cell , or the initial validation (the table).
maybe we don't need this ?

yaguzmang: We still need this because when getting the dependants, this 
```ts
const dependants = AssessmentMetaCaches.getValidationsDependants({
  assessment,
  cycle,
  tableName: variable.tableName,
  variableName: variable.variableName,
})
```
can return only `{ tableName, variableName }` and in `validateNodeUpdates.ts` we expect also the `colName` to have a concrete cell:
```ts
    const row = rows[RowCaches.getKey({ tableName, variableName })]
    const col = row?.cols?.find((candidate: Col) => candidate.props.colName === colName)
    const validateFns = col?.props.validateFns?.[cycle.uuid] ?? row?.props.validateFns?.[cycle.uuid]

    if (Objects.isNil(row) || Objects.isNil(col)) {
      throw new Error(`Could not resolve validation target ${tableName}.${variableName}.${colName}`)
    }
```

This error happened to me while testing before this fix. I got: `Could not resolve validation target forestAreaChange.forestAreaNetChange.2025` because instead of getting each concrete col I was incorreclty doing this when building the dependants: `colName: dependant.colName ?? variable.colName`. 

In short, if `AssessmentMetaCaches.getValidationsDependants` doesn't give us a concrete node, we have to get the entire row columns to build the node references. 

There is an old comment in the definition of `VariableCache`:
```ts
export interface VariableCache {
  assessmentName?: AssessmentName
  cycleName?: CycleName
  tableName: TableName
  variableName: VariableName
  colName?: ColName // TODO: will colName become mandatory when handling dependencies by col ?
}
```

Maybe we should look into making colName mandatory now? let me know what you think

minotogna: Yes I know . I'm not saying the opposite .
I'm only saying we don't validate per row anymore , we validate per cell .
The row associated to the cell must be retrieved

yaguzmang: Ah I see what you are saying. Yes, here we are still effectively validating cells. In `#expandVariable` we only look up the row when the dependant does not come back with a colName, so we can resolve that variable into cell targets. So yes, we go over the row to build the cells to validate, not to validate the row as a unit.

For example, these do not return a colName in the validation dependants (`AssessmentMetaCaches.getValidationsDependants`):

- `extentOfForest.forestArea` -> `{ tableName: 'forestAreaChange', variableName: 'forestAreaNetChange' }`
- `extentOfForest.forestArea` -> `{ tableName: 'disturbances', variableName: 'insects' }`
- `extentOfForest.otherLand` -> `{ tableName: 'otherLandWithTreeCover', variableName: 'agroforestry' }`

For those cases, even if we backfill all table validations at first, after a user edit we still need to resolve them into specific [tableName, variableName, colName] targets for the incremental validation step.

minotogna: But why do we get all rows when we get one row ? Can't the rows be part of the context ? I'm having trouble in understanding the scope of this pr .

yaguzmang: Removed the code for expanding row cells 👍

src/meta/assessment/metaCache/index.ts
```diff
@@ -28,14 +28,30 @@ export type VariablesCache = Record<TableName, Record<VariableName, VariableCach
...
 
+export type ValidationTargetCache = {
+  colName: ColName
+  tableName: TableName
+  variableName: VariableName
+}
+
+export type ValidationTargetsBySource = Record<
+  TableName,
+  Record<VariableName, Record<ColName, Array<ValidationTargetCache>>>
```
minotogna: let's still use VariableCache. we'll check it later on , no need to introduce ValidationTargetCache

minotogna (approved):

PR #5842, opened 2026-04-15
5758 - Navigation: Data errors - Get table validations

yaguzmang:
https://github.com/user-attachments/assets/a338e67c-8af3-4b45-83ef-147d0fab9168

Final TODO: 
- Fix icons issue (maybe bundle icons with the app?)
- E2E Tests expansion for validations


src/client/pages/Section/hooks/useGetTableData/useGetTableData.ts
```diff
@@ -24,14 +25,21 @@ export const useGetTableData = (props: Props): void => {
...
       // fetch internal dependencies
-      if (tableNames.size > 0) {
+      if (tableNamesSet.size > 0) {
         const propsFetch = { assessmentName, cycleName, countryIso, mergeOdp: !tableWithOdp }
+        const tableNames = Array.from(tableNamesSet)
 
-        dispatch(NodeValuesActions.getTableData({ ...propsFetch, tableNames: Array.from(tableNames) }))
+        dispatch(NodeValuesActions.getTableData({ ...propsFetch, tableNames }))
+
+        if (canEdit) {
```
minotogna: The validations should be fetched in their own hook . 

This hook is getTableData and should just do that ; single responsibility of the software components is better

src/client/store/data/tableData/validations/actions/getTableValidations.ts
```diff
@@ -0,0 +1,31 @@
...
+    const params = { assessmentName, countryIso, cycleName, sectionName, tableNames }
+
+    const { data } = await axios.get<RecordTableValidationsState>(ApiEndPoint.CycleData.Validations.tableData(), {
+      params,
+    })
+
+    dispatch(setNodeValueValidations({ assessmentName, cycleName, countryIso, tableValidations: data }))
+
+    return data
+  }
```
minotogna, line 28: Why dispatching and returning data ? One of the two should be enough ?

yaguzmang: True 👍 , kept dispatch only

yaguzmang:
~~Tests on this PR fail due to the validation dependencies not being correct yet. So we need to hold this one until we implement the new validation dependents cache.~~

minotogna (approved):

PR #5852, opened 2026-04-18
5758 - Navigation: Data errors - Show nav validation error indicator

yaguzmang:
image

https://github.com/user-attachments/assets/d347c788-0783-40a8-96f6-0c05c9791711

RTL:
image

### DIFF
--- a/src/client/components/Navigation/NavAssessment/Section/NavigationSection.tsx
+++ b/src/client/components/Navigation/NavAssessment/Section/NavigationSection.tsx
@@ -79,7 +79,7 @@ const NavigationSection: React.FC<Props> = (props) => {
         tabIndex={0}
       >
         <div className="nav-section__order">{prefix}</div>
-        <Flex alignItems="center" display="inline-flex" gap="4">
+        <Flex alignItems="center">
           <div className="nav-section__label">{sectionLabel}</div>
         </Flex>
         {!expanded && !isDataExport && (
--- a/src/client/components/Navigation/NavAssessment/Section/NavigationSection.tsx
+++ b/src/client/components/Navigation/NavAssessment/Section/NavigationSection.tsx
@@ -12,6 +12,7 @@ import { useCycle } from 'client/store/meta/hooks/cycles'
 import { useSectionReviewSummary } from 'client/store/review/hooks/review'
 import { useCountryIso } from 'client/hooks/country'
 import { useIsDataExportView } from 'client/hooks/dataExport'
+import Flex from 'client/components/Layout/Flex'
 import SectionItemLink from 'client/components/Navigation/NavAssessment/Section/SectionItemLink'
 import ReviewSummaryIndicator from 'client/components/ReviewSummaryIndicator'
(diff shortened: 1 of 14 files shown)


src/client/components/Navigation/NavAssessment/Section/Section.scss
```diff
@@ -25,6 +25,12 @@
   font-size: $font-s;
 }
 
+.nav-section__label-content {
```
minotogna: Can we use a Flex element ?

src/client/components/ValidationErrorIndicator/ValidationErrorIndicator.tsx
```diff
@@ -0,0 +1,27 @@
...
+  uuid: UUID
+}
+
+const ValidationErrorIndicator: React.FC<Props> = (props) => {
+  const show = useShow(props)
+
+  if (!show) return null
+
+  return (
+    <div className="validation-error-indicator">
```
minotogna: And here a Flex element ?

src/client/components/Navigation/NavAssessment/Section/Section.scss
yaguzmang: Yes. I now added an inline-flex display option to Flex 👍

minotogna: Why do we need inline flex
Here ? Wasn't flex
Enough ?

src/client/components/Navigation/NavAssessment/Section/NavigationSection.tsx
```diff
@@ -77,10 +79,13 @@ const NavigationSection: React.FC<Props> = (props) => {
         tabIndex={0}
       >
         <div className="nav-section__order">{prefix}</div>
-        <div className="nav-section__label">{sectionLabel}</div>
+        <Flex alignItems="center" display="inline-flex" gap="4">
```
minotogna: We don't need gap with one element only

src/client/components/Navigation/NavAssessment/Section/SectionItemLink/SectionItemLink.tsx
```diff
@@ -61,10 +63,13 @@ const SectionItemLink: React.FC<Props> = (props) => {
       <div className="nav-section__order">
         {t(SubSections.getAnchorLabel({ assessment, cycle, subSection }), SubSections.getAnchor({ cycle, subSection }))}
       </div>
-      <div className="nav-section__label">{Labels.getCycleLabel({ cycle, labels: subSection.props.labels, t })}</div>
+      <Flex alignItems="center" display="inline-flex" gap="4">
```
minotogna: And here?

src/client/components/Navigation/NavAssessment/Section/Section.scss
yaguzmang: you are right. With the new layout inline-flex was not necessary, flex does 👍

minotogna (approved):

PR #5857, opened 2026-04-21
5758 - Navigation: Data errors - New Validation cache types - 2

yaguzmang:
These changes introduce the new way we will get the validation dependants:

1. Before this change, validation dependants were variable-level, not cell-level.

So for something like:

```ts
extentOfForest.forestArea
-> forestAreaChange.forestAreaNetChange
```

The system could calculate that `forestAreaNetChange` depended on `forestArea`, but it did **not** know which target cell to validate.

That caused errors after editing some cells:

```ts
extentOfForest.forestArea['2025']
```

was resolving this target:

```ts
forestAreaChange.forestAreaNetChange['2025']
```

which is invalid, because that row only has cols:

- `1990-2000`
- `2000-2010`
- `2010-2015`
- `2015-2020`
- `2020-2025`

In this PR we make validation dependants include specific colNames, so we know the exact target cell(s) to revalidate.

---

**2. We changed only the validation dependants to be source-cell keyed**

In `src/meta/assessment/metaCache/index.ts`, validations have this shape:

```ts
export type ValidationTargetsBySource = Record<
  TableName,
  Record<VariableName, Record<ColName, Array<VariableCache>>>
>

export type ValidationDependencyCache = {
  dependencies: DependencyRecord
  dependants: ValidationTargetsBySource
}
```
Meaning:
- `dependencies` -> The input this validation target depends on.
- `dependants` -> When this specific source cell changes, these are the exact validation target cells to re-run.

The `dependants` shape is:

```ts
dependants[sourceTable][sourceVariable][sourceCol] = Array<targetVariables>
```

Example:

```ts
dependants['extentOfForest']['forestArea']['2025'] = [
  { tableName: 'forestAreaChange', variableName: 'forestAreaNetChange', colName: '2020-2025' }
]
```

---
## Code changes

**3. `src/server/cache/repository/metaCache/generateMetaCache/index.ts` now evaluates validations per target cell**

Before, validation dependency generation did not run with a concrete target-column context.

Now:

- row-level `validateFns` are evaluated once per actual `row.col`
- col-level `validateFns` are also evaluated with `col` in context
- structural cols with blank `colName` are skipped

So when generating the validation for:

```ts
forestAreaChange.forestAreaNetChange['2020-2025']
```

we pass the evaluator the current target col,  `'2020-2025'`.
By passing that target col we can store the exact source-cell -> target-cell mapping, instead of only a source-variable -> target-variable mapping.

---

**4. `src/server/cache/repository/metaCache/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts` ** Here we write the source-cell -> target-cell mapping.

Only for validations, `MemberEvaluator` now has special logic for building the dependant graph.

**`#getValidationSourceVariable(...)`**-> resolves what the source cell is for a validation dependant. Considering:

- if the member already has `colName`, keep it
- if it has no `colName`, use the current target `col.props.colName`
- if it points to another cycle/assessment, do not create a dependant edge

**`#addValidationDependant(...)`** -> writes the reverse graph.

It:
- resolves the source validation cell
- reads the current target `colName`
- stores:

```ts
dependants[sourceTable][sourceVariable][sourceCol].push({
  tableName: targetTable,
  variableName: targetVariable,
  colName: targetCol,
})
```

In practice, it stores mappings like:

```ts
extentOfForest.forestArea['2025']
-> forestAreaChange.forestAreaNetChange['2020-2025']
```

### Note
In `#addDependency(...)`, for validations we strip `colName` before storing dependencies:

```ts
const dependencyVariable = type === 'validations' ? { ...variable, colName: undefined } : variable
```
Because for validation dependencies we don't need the exact cells to do data loading. 

---

**5. `src/meta/assessment/metaCaches/index.ts` now reads validations with the new shape**

---

Next PR(s): 
- Introduce a tool to backfill all cache for all countries in all cycles


src/meta/assessment/metaCaches/index.ts
```diff
@@ -23,13 +26,15 @@ type TableProps = CycleProps & {
 type VariableProps = TableProps & {
   variableName: VariableName
 }
+type ValidationVariableProps = VariableProps & {
```
minotogna: should this be called ColProps ?

src/server/cache/repository/metaCache/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
```diff
@@ -101,9 +164,9 @@ export class MemberEvaluator extends ExpressionNodeEvaluator<Context, MemberExpr
         dependencies = AssessmentMetaCaches.getValidationsDependencies(propsDependency)
       }
 
-      if (!_includesVariableCache(dependencies, variable)) {
+      if (!_includesVariableCache(dependencies, dependencyVariable)) {
         const path = [type, 'dependencies', tableName, variableName]
-        Objects.setInPath({ obj: metaCache, path, value: [...dependencies, variable] })
+        Objects.setInPath({ obj: metaCache, path, value: [...dependencies, dependencyVariable] })
```
minotogna: it seems this file become over complicated and long to maintain .
maybe another Member implementation for validations ?

minotogna (approved):

PR #5858, opened 2026-04-21
5767 - Repository/Migration: created_at

sorja:
image


src/tools/migrations/steps/steps/20260421060924-step-populate-repository-created-at.ts
```diff
@@ -0,0 +1,22 @@
+import { AssessmentController } from 'server/controller/assessment'
+import { BaseProtocol } from 'server/db/db'
+import { Schemas } from 'server/db/schemas'
+
+export default async (client: BaseProtocol): Promise<void> => {
```
minotogna: don't we have already a migration in this task ? I would use only one migration

sorja: We do but I prefer keeping data migration and DDL migration separate, what do you think?

I can merge the migrations if you prefer

minotogna: Better one migration for when we update production

sorja: Agree, already working on it

minotogna (approved):

PR #5863, opened 2026-04-22
5767 - Use background-color over box-shadow

sorja:
image

### DIFF
--- a/src/client/components/RepositoryList/RepositoryListItem/RepositoryListItem.scss
+++ b/src/client/components/RepositoryList/RepositoryListItem/RepositoryListItem.scss
@@ -13,6 +13,6 @@
   }
 
   &:hover {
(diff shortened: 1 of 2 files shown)


src/client/components/RepositoryList/RepositoryListItem/RepositoryListItem.scss
```diff
@@ -13,6 +13,6 @@
   }
 
   &:hover {
-    box-shadow: inset 0 0 50px 10px $ui-accent-light;
+    background-color: $ui-accent-light;
```
minotogna: Better to use css var ? Pan European should be different with css var . Can you please double check ?

sorja: Good point, checked and it was different. Fixed accordingly:

image

minotogna (approved):

PR #5864, opened 2026-04-22
5767 - Disable removing folders containing items

sorja:
image

Developers note: Can't add tooltip using props for disabled buttons

### DIFF
--- a/src/client/components/Buttons/Button/Button.scss
+++ b/src/client/components/Buttons/Button/Button.scss
@@ -188,8 +188,3 @@
     background-color: transparent;
   }
 }
-
-// Allow tooltip to appear on a disabled button (disabled blocks clicks, not hover)
-.btn--tooltip-on-disabled:disabled {
-  pointer-events: auto;
-}
--- a/src/client/components/Buttons/Button/Button.scss
+++ b/src/client/components/Buttons/Button/Button.scss
@@ -188,3 +188,8 @@
     background-color: transparent;
   }
 }
+
+// Allow tooltip to appear on a disabled button (disabled blocks clicks, not hover)
+.btn--tooltip-on-disabled:disabled {
+  pointer-events: auto;
+}
--- a/src/client/components/RepositoryList/EditForm/RepositoryItemForm/RepositoryItemForm.tsx
+++ b/src/client/components/RepositoryList/EditForm/RepositoryItemForm/RepositoryItemForm.tsx
@@ -1,9 +1,13 @@
 import React from 'react'
 import { useTranslation } from 'react-i18next'
+import { Tooltip } from 'react-tooltip'
+import classNames from 'classnames'
 
 import { ApiEndPoint } from 'meta/api/endpoint'
 import { CountryIso } from 'meta/area/countryIso'
-import { RepositoryItem } from 'meta/cycleData/repository/item'
+import { RepositoryItem, RepositoryItemTree } from 'meta/cycleData/repository/item'
+import { RepositoryItems } from 'meta/cycleData/repository/items'
+import { TooltipId } from 'meta/tooltip/id'
 import { Objects } from 'utils/objects'
 
 import { useCountryRouteParams } from 'client/hooks/routeParams'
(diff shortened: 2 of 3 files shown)


src/client/components/Buttons/Button/Button.scss
```diff
@@ -188,3 +188,8 @@
     background-color: transparent;
   }
 }
+
+// Allow tooltip to appear on a disabled button (disabled blocks clicks, not hover)
+.btn--tooltip-on-disabled:disabled {
```
minotogna: Not sure if it's the best idea to add this complicated class .
Maybe a div sorrounding the Button where to put the tooltip ?

sorja: Added more contained version now (removed generic class).
Displaying the tooltip on disabled button is a bit problematic because the browser sets mouse events to none, so the mouseEnter-mouseLeave events etc are never triggered
Even wrapping with a div doesn't get around this, as we're still hovering the button

minotogna: The tooltip on the surrounding div doesn't work ?

sorja: No

src/client/components/RepositoryList/EditForm/RepositoryItemForm/RepositoryItemForm.tsx
```diff
@@ -53,13 +58,20 @@ const RepositoryItemForm: React.FC<Props> = (props) => {
       />
       {isEditing && (
         <div className="repository-item-form__delete">
```
minotogna, line 60: The tooltip  content can be added here directly if matches the condition ? 
```
<div
  data-tooltip-id="my-tooltip"
  data-tooltip-content="Hello world!"
  data-tooltip-place="top"
>
```

src/client/components/Buttons/Button/Button.scss
minotogna: Strange . Do we have the Tooltip rendered somewhere ? We should use that id . See my other comment

minotogna: You can use this ?

https://github.com/openforis/fra-platform/blob/development/src/client/components/Tooltips/Tooltips.tsx

src/client/components/RepositoryList/EditForm/RepositoryItemForm/RepositoryItemForm.tsx
sorja: This doesnt work

See my previous comment (wed still be hovering the button)

src/client/components/Buttons/Button/Button.scss
sorja: Don't think this applies here

We can't use tooltip if we hover over (mouseEnter) a disabled button, the event doesn't fire. No matter if it's the wrapper div or the element itself

The Tooltips you linked is the generic component used here:
image

Which is triggered by the attributes specified in the component

I would suggest replacing this Tooltip component (or generally investigating the functioanlity a bit better)

minotogna: > Don't think this applies here
> 
> 
> 
> We can't use tooltip if we hover over (mouseEnter) a disabled button, the event doesn't fire. No matter if it's the wrapper div or the element itself
> 
> 
> 
> The Tooltips you linked is the generic component used here:
> 
> image
> 
> 
> 
> Which is triggered by the attributes specified in the component
> 
> 
> 
> I would suggest replacing this Tooltip component (or generally investigating the functioanlity a bit better)

It can apply there @sorja 
The tooltip attributes can be conditionally placed on the div around the button , not on the button

sorja: This is what I did.
It did not work, hence it the `does not apply here`

minotogna: Can u paste the code of what didn't work ? I'm curious 🧐

minotogna (approved):

PR #5867, opened 2026-04-23
5767 - Repository: Read only view

sorja:
image

### DIFF
--- a/src/client/components/RepositoryList/ColumnHeaders/ColumnHeaders.tsx
+++ b/src/client/components/RepositoryList/ColumnHeaders/ColumnHeaders.tsx
@@ -38,20 +38,22 @@ const ColumnHeaders: React.FC<Props> = (props) => {
         {t('common.name')}
         {!readOnly && <OrderBy column={col('name')} path={path} />}
       </div>
-      <div className="repository-column-headers__col">
-        {t('common.added')}
-        {!readOnly && <OrderBy column={col('createdAt')} path={path} />}
-      </div>
-      {!selectable && (
+      {!readOnly && (
         <div className="repository-column-headers__col">
-          {t('common.linked')}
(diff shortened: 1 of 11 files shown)


src/client/components/RepositoryList/ColumnHeaders/ColumnHeaders.tsx
```diff
@@ -36,19 +36,21 @@ const ColumnHeaders: React.FC<Props> = (props) => {
       </div>
       <div className="repository-column-headers__col repository-column-headers__col--name">
         {t('common.name')}
-        <OrderBy column={col('name')} path={path} />
+        {!readOnly && <OrderBy column={col('name')} path={path} />}
```
minotogna: For me readOnly refers to edit mode not allowed ,  not hiding order by or other options . 
I would say allowSorting ? Etc... what do you think ?

sorja: Sure, originally I thought of passing from outside readOnly prop, but then I decided to just deduct it from isGlobal.

Should I instead pass related props like:

allowSorting
allowFiltering
showColumns 

Etc?

sorja: What I mean:

Currently:

image

After:
image

sorja: Or are you referring to just renaming the variable used here?

minotogna: > What I mean:
> 
> 
> 
> Currently:
> 
> 
> 
> image
> 
> 
> 
> After:
> 
> image

@sorja the after looks more clear and flexible

minotogna (approved):

PR #5871, opened 2026-04-24
5767 - Modal layout

sorja:
https://github.com/user-attachments/assets/38b491b8-061d-42fc-b62c-ad83c893ee15

Note:
Added: BaseFileSummary to fix some type issue (to avoid Partial type when we only need name/size/etc)

### DIFF
--- a/src/client/components/EditorWYSIWYG/AddFromRepository/AddFromRepository.scss
+++ b/src/client/components/EditorWYSIWYG/AddFromRepository/AddFromRepository.scss
@@ -22,6 +22,7 @@
 .repository-modal .modal-content {
   height: 80vh;
   margin: 5% 30%;
+  min-width: 450px;
   overflow: hidden;
 }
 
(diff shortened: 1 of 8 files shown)


src/client/components/EditorWYSIWYG/AddFromRepository/AddFromRepository.scss
```diff
@@ -19,9 +19,19 @@
   }
 }
 
-.repository-modal {
-  .modal-header span {
-    font-size: $font-s;
-    font-weight: initial;
-  }
+.repository-modal .modal-content {
+  height: 80vh;
```
minotogna: better dvh (for safari )

src/client/components/EditorWYSIWYG/AddFromRepository/AddFromRepository.scss
```diff
@@ -19,9 +19,19 @@
...
+  margin: 5% 30%;
```
minotogna: Do we need Margin if we dynamically center it (with flex ? ) ?

minotogna (approved):

PR #5872, opened 2026-04-26
5758 - Navigation: Data errors - E2E Tests

yaguzmang:
Tests:
1. `shows validation errors on initial load`
   Seeds invalid `forestAreaNetChange` data through the backend, opens the `forestAreaChange` subsection, and verifies the validation warnings are rendered for the cell, table, subsection, section, and submit-to-review flow.

2. `clears validation errors after fixing the data`
   Starts from the same invalid seeded state, then seeds the corrected values and verifies the `forestAreaNetChange['2020-2025']` validation error disappears in the rendered UI.

3. `editing extentOfForest updates forestAreaChange validation errors`
   Uses the browser to edit `extentOfForest.forestArea['2025']`, verifies the `forestAreaChange` navbar warning is displayed, navigates to the target subsection, and confirms the target cell/table become invalid and then clear again after fixing the source value.

4. `editing forestArea validates against the previous cycle`
   Seeds `fra 2020 / extentOfForest.forestArea['2020']`, seeds the matching `fra 2025` value, then edits the `fra 2025` cell in the browser to verify the previous-cycle validation appears when the values differ and clears again when they match.  
   Note: editing `fra 2020` will not trigger validations in `fra 2025`. We support reading data from another cycle as validation input, but we do not propagate validation updates across cycles. Because of that, this case needs to be tested by editing the `2020` value in the `fra 2025` cycle itself.

5. `does not show validation UI on the print route`
   Seeds invalid validation state, opens the print tables route, and verifies validation UI is not rendered.


src/test/e2e/utils/DOM.ts
```diff
@@ -33,7 +33,7 @@ const elementNotExists = async (locator: Locator): Promise<void> => {
 
 // Internal utility to find datatable cell
 const cellLocator = (page: Page, variableName: string, colName: string): Locator =>
-  page.locator(`[id*="variableName_${variableName}"][id*="colName_${colName}"]`)
+  page.locator(`[id$="variableName_${variableName}_colName_${colName}"]`)
```
yaguzmang, line 36: DOM ids look like:
- `...variableName_forestAreaNetChange_colName_2020-2025`
- `...variableName_forestAreaNetChangeFrom1a_colName_2020-2025`

With the old selector `forestAreaNetChange` also matched `forestAreaNetChangeFrom1a`.

This new version fixes that.

src/test/e2e/tests/05-validations.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    await _seedForestAreaNetChangeValidation(page, { valid: false })
+
+    await page.goto(printTablesPath)
+    await expect(DOMUtils.tableContainer(page, 'forestAreaChange')).toBeVisible({ timeout: 20000 })
+    await expect(DOMUtils.tableValidationErrors(page, 'forestAreaChange')).toHaveCount(0)
+    await expect(page.locator('.validation-error-indicator')).toHaveCount(0)
+    await DOMUtils.expectCellHasNoValidationError(page, 'forestAreaNetChange', '2020-2025')
+  })
+})
```
yaguzmang: I can add a few more data tests besides the `'forestAreaNetChange'.'2020-2025'` one, though this file is getting big

minotogna: This is ok for now .

I would test forest area with previous cycle if possible. What do you think ?

yaguzmang: Added 👍 . Since we skip that check for atlantis countries: `shouldSkipValidationFormula.ts`:

```ts
export const shouldSkipValidationFormula = (props: Props): boolean => {
  const { countryIso, formula } = props

  return Areas.isAtlantis(countryIso) && formula.includes(validatorEqualToPreviousCycleForestArea.name)
}
```
I used the countryIso `ALB`. I wanted to just use one countryIso but all countries are published in 2025, and I needed a non-published country to test the submit-to-review warning.

minotogna: Sure 👍 but can we have one file per table validation ? Otherwise this file will become super huge 😅

yaguzmang: You mean one file per validation test so we separate the 5 here?

yaguzmang: what do you think if this split:


folder `05-validations/`
- `table-validations.spec.ts` ->  initial load, validation clearing, propagation and print route test
- `cross-cycle-validations.spec.ts` -> New prev cycle check test

Add new util for tests: `NodeValues` and extract `_patchNodeValues` to it, so:

```
export const NodeValues = {
  patch,
}
```

That also fixes the issue of using two different countries in one file :)

minotogna: Call the tests by validation function , it will more clear what they test .

yaguzmang: I ended up splitting them into:
`rendering-validations.spec.ts`: Checks the initial validations load and the print route.
`update-validations.spec.ts`: Checks data changes and propagation (e.g forestAreaNetChange).
`cross-cycle-validations.spec.ts`: Checks the prev cycle data validation. 
also added the `NodeValues` util and a common helper for those files.

Now the tests are about 60 lines each :)
let me know what you think

src/test/e2e/tests/05-validations/01-cross-cycle-validations.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+import { NodeValues } from 'test/e2e/utils/NodeValues'
+
+const assessmentName = AssessmentNames.fra
+const cycleName = CycleNames._2025
+const countryIso = 'ALB' // Non-atlantis country to test cross-cycle validations (see: shouldSkipValidationFormula.ts).
+const previousCycleForestArea2020 = '2000'
+
+const extentOfForestPath = `/assessments/${assessmentName}/${cycleName}/${countryIso}/sections/${SectionNames.extentOfForest}`
+
+test.describe.serial('Cross-cycle validations', () => {
```
minotogna, line 17: Should we order the files with their own numbers ? Difficult to see the order if they belong all to 05.
What do you think ?

yaguzmang: image

Added 👍 . Though since every file runs in parallel, the order of the logs might not be the same

minotogna (approved):

PR #5874, opened 2026-04-27
5767 - Repository/Migration: Backfill created_at

sorja:
S/s: sent in Slack

Use `LEFT JOIN LATERAL` to run a subquery once per row, referencing the current row's columns inside it

### DIFF
--- a/src/tools/migrations/steps/steps/20260323110159-step-repository-folder-columns.ts
+++ b/src/tools/migrations/steps/steps/20260323110159-step-repository-folder-columns.ts
@@ -1,3 +1,5 @@
+import pgPromise from 'pg-promise'
+
 import { ActivityLogMessage } from 'meta/assessment/activityLog'
 import { Assessment } from 'meta/assessment/assessment'
 import { Cycle } from 'meta/assessment/cycle'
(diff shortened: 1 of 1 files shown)


src/tools/migrations/steps/steps/20260323110159-step-repository-folder-columns.ts
```diff
@@ -1,29 +1,80 @@
...
+  const prevSchema = prevCycle ? Schemas.getNameCycle(assessment, prevCycle) : null
+  const cycleCreatedAt = cycle.props.dateCreated
+
+  const prevFallback = prevSchema
+    ? `(select prev.created_at from ${prevSchema}.repository prev where prev.uuid = r.uuid limit 1),
+       (select prev.created_at from ${prevSchema}.repository prev where r.file_uuid is not null and prev.file_uuid = r.file_uuid limit 1),
+       (select prev.created_at from ${prevSchema}.repository prev where r.link is not null and prev.link = r.link limit 1),`
+    : ''
+
+  // NOTE: Slow query!
```
minotogna: How much slow ? Will prod handle it ? If not , we should find another way

minotogna (approved):

PR #5877, opened 2026-04-29
5767 - Repository: E2E

sorja:


src/test/e2e/tests/05-repository.spec.ts
```diff
@@ -0,0 +1,104 @@
+import { expect, test } from '../fixtures/auth'
+import { DOMUtils } from '../utils/DOM'
+
+const rootLinkLabel = 'E2E test link'
+const rootLinkLabelEdited = 'E2E test link edited'
+const folderName = 'E2E test folder'
+const nestedLinkLabel = 'E2E nested link'
+
```
minotogna, line 8: 05 is already in the validations test branch . Should this be 06 or validations 06?
What do you think ?

sorja: Hmm...

I think we should go top down (UI) so this would be first.

My first thought was putting this to 06 though, what do you prefer?

So logic order would be:
- Country home (repostiory, collab, etc)
- Sections/tables/etc


Reasoning: You can use the repository files in sections where also tables exist

minotogna (approved):

PR #5882, opened 2026-04-29
5758 - Navigation: Data errors - All cols validations

yaguzmang:
Fixing validations for `maxForestArea()` and `maxLandArea()` by storing them as all-column validation dependencies in the meta cache.

Previously, these helpers were normal column-scoped dependencies, so editing `extentOfForest.forestArea['2025']` would not revalidate targets that used `maxForestArea()` in other years. e.g `disturbances.insects['2005']`. Now we store those helpers under an all-columns dependency (`*`), and validation lookup includes both exact column and all-column dependants.

Changes:

- Add `AllColumnsDependencyKey = '*'`
- Store `maxForestArea()` under `extentOfForest.forestArea['*']`
- Store `maxLandArea()` under `extentOfForest.totalLandArea['*']`
- Include `*` dependants during validation lookup
- Rename `customDepenendencies` to `customDependencies`

TODO in different PR: Test

### DIFF
--- a/src/server/cache/repository/metaCache/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
+++ b/src/server/cache/repository/metaCache/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
@@ -63,8 +63,7 @@ export class MemberEvaluator extends ExpressionNodeEvaluator<Context, MemberExpr
         // Prevent validating nodes from other asessments/cycles
         if (assessment.props.name !== assessmentName || cycle.name !== cycleName) return
 
-        dependants =
-          AssessmentMetaCaches.getTableValidationsDependants(propsDependants)?.[variable.variableName]?.[colName] ?? []
+        dependants = AssessmentMetaCaches.getValidationsDependantsAtCol({ ...propsDependants, colName })
       }
       const external = assessmentName !== assessment.props.name
       const dependant: VariableCache = {
--- a/src/server/cache/repository/metaCache/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
+++ b/src/server/cache/repository/metaCache/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
@@ -63,7 +63,8 @@ export class MemberEvaluator extends ExpressionNodeEvaluator<Context, MemberExpr
         // Prevent validating nodes from other asessments/cycles
         if (assessment.props.name !== assessmentName || cycle.name !== cycleName) return
 
-        dependants = AssessmentMetaCaches.getValidationsDependants({ ...propsDependants, colName })
+        dependants =
+          AssessmentMetaCaches.getTableValidationsDependants(propsDependants)?.[variable.variableName]?.[colName] ?? []
       }
       const external = assessmentName !== assessment.props.name
       const dependant: VariableCache = {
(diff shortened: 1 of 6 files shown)


src/server/cache/repository/metaCache/generateMetaCache/dependencyEvaluator/evalDependencies/member.ts
```diff
@@ -63,7 +63,8 @@ export class MemberEvaluator extends ExpressionNodeEvaluator<Context, MemberExpr
         // Prevent validating nodes from other asessments/cycles
         if (assessment.props.name !== assessmentName || cycle.name !== cycleName) return
 
-        dependants = AssessmentMetaCaches.getValidationsDependants({ ...propsDependants, colName })
+        dependants =
+          AssessmentMetaCaches.getTableValidationsDependants(propsDependants)?.[variable.variableName]?.[colName] ?? []
```
minotogna: Isn't the default already returned from this method ? Should this method accept variableName and colName ?

yaguzmang: We use `getTableValidationsDependants` more generally elsewhere and we have to keep it like that, so I added `getValidationsDependantsAtCol`. That takes the col prop and handles the default 👍

minotogna: @yaguzmang I had a better look. can we have only one between getValidationsDependantsAtCol and getValidationsDependants.
There's currently only one usage of getValidationsDependants and I think it can be only one method. please have a look

yaguzmang: Removed `getValidationsDependantsAtCol` and replaced it with just `getValidationsDependants` + `includeAllColumnsDependants` prop, which we set to `false` during graph generation so `*` dependants stay separate from year-specific dependants

Let me know what you think

yaguzmang: Example for why we need `includeAllColumnsDependants`: based on real data:

`extentOfForest.forestArea['*']` contains targets like `disturbances.insects['2005']` because it uses `maxForestArea()`.

`extentOfForest.forestArea['2025']` contains normal same year targets like `forestCharacteristics.totalForestArea['2025']`.

During graph generation, when we process a dependency on `extentOfForest.forestArea['2025']`, we read the existing dependants for that exact entry, append the new dependant, and write the list back to `forestArea['2025']`. 

So if that read also includes the `*` dependants, we would write those back into the concrete `2025` entry too. For example, `disturbances.insects['2005']` belongs under `forestArea['*']`, but it would also get copied into `forestArea['2025']`.

We would have: 

```
forestArea['*']    -> disturbances.insects['2005']  correct
forestArea['2025'] -> disturbances.insects['2005']  wrong duplicate
```

The `*` entry already means “edits to any forestArea column should revalidate this target”. So we should not also copy that target into a concrete year like `2025`.

When validating at runtime, we want the opposite: if a user edits `forestArea['2025']`, we do want both the `2025` dependants and the `*` dependants to revalidate.

minotogna (approved):

PR #5884, opened 2026-04-29
5758 - Navigation: Data errors - Prevent stale data

yaguzmang:
Clearing the country validation cache before rebuilding it.

Why:

`validateAll` only queues cells that currently have `validateFns`. If a validator is removed from metadata, cells that used to have that validator are no longer queued, so their old validation results are never revisited or cleared. That can leave stale errors in Redis after a full rebuild.

## Example

Suppose an older metadata version has a validator on:

```ts
tableA.oldVariable["2025"]
```

That validator fails, so Redis stores an error for the country:

```json
{
  "tableA": {
    "2025": {
      "oldVariable": { "valid": false }
    }
  }
}
```

Later, metadata changes and `oldVariable` no longer has that validator, or the variable is removed entirely.

When `validateAll` runs again, `buildTablesNodeUpdates` only queues cells with current `validateFns`, so `oldVariable` is never validated again. Because it is never queued, `updateValidations` never removes the old Redis entry.

The result is stale validation state: the table can still appear invalid in the left navigation or submit-to-review warning.

Fix: clearing validations before rebuilding them in `validateCountryTables`.

### DIFF
--- a/src/server/controller/cycleData/validations/country/validateCountryTables.ts
+++ b/src/server/controller/cycleData/validations/country/validateCountryTables.ts
@@ -1,5 +1,6 @@
 import { CountryIso } from 'meta/area/countryIso'
 import { AssessmentName } from 'meta/assessment/assessment'
+import { Assessments } from 'meta/assessment/assessments'
 import { CycleName } from 'meta/assessment/cycle'
 
 import { TableRedisRepository } from 'server/cache/repository/table'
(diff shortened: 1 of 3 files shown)


src/server/controller/cycleData/validations/country/validateCountryTables.ts
```diff
@@ -19,16 +21,17 @@ type Props = {
...
 
-  const { assessment, cycle } = await AssessmentController.getOneWithCycle(
-    { assessmentName, cycleName, metaCache: true },
-    client
-  )
+  const assessment = await AssessmentController.getOne({ assessmentName, metaCache: true }, client)
+  const cycle = Assessments.getCycle({ assessment, cycleName })
+
   const [country, tables] = await Promise.all([
     AreaController.getCountry({ assessment, countryIso, cycle }, client),
```
minotogna: Don't we have a redis area repo ?

yaguzmang: Fixed :+1:

minotogna (approved):

PR #5889, opened 2026-05-04
5844 be 1

sorja:
The general idea is:

```

if (Users.isPersonalInfoRequiredForRole) redirect/goTo LoginInvitationUserInfo


```

- Allow creating user without accepting role

### DIFF
--- a/src/meta/user/users/isPersonalInfoRequired.ts
+++ b/src/meta/user/users/isPersonalInfoRequired.ts
@@ -1,6 +1,3 @@
-import { Objects } from 'utils/objects'
-import { RegExps } from 'utils/regExps'
-
 import { RoleName } from 'meta/user/role/name'
 import {
   UserContactPreference,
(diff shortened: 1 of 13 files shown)


src/meta/user/users/isPersonalInfoRequired.ts
```diff
@@ -11,6 +8,8 @@ import {
 import { UserRole } from 'meta/user/role/role'
 import { User, UserProps } from 'meta/user/user'
 import { isAdministrator } from 'meta/user/users/isRole'
+import { Objects } from 'utils/objects'
+import { RegExps } from 'utils/regExps'
 
 export const isPersonalInfoRequired = (user: User, role: UserRole): boolean => {
```
minotogna, line 14: I suspect this one won't be needed anymore. you might deprecate this one and add it (same name) to src/meta/user/roles/roles.ts and not here. what do you think ?

PR #5893, opened 2026-05-05
5844 - Invitation flow: require user info before accepting invitation

sorja:
Tests need to be fixed later

### DIFF
--- a/src/meta/user/roles/roles.ts
+++ b/src/meta/user/roles/roles.ts
@@ -9,7 +9,7 @@ import { UserRole } from 'meta/user/role/role'
 import { User } from 'meta/user/user'
 import { Dates } from 'utils/dates'
 
-const isPersonalInfoRequiredForRole = (roleName: RoleName): boolean =>
+const isPersonalInfoRequired = (roleName: RoleName): boolean =>
   [RoleName.NATIONAL_CORRESPONDENT, RoleName.ALTERNATE_NATIONAL_CORRESPONDENT, RoleName.COLLABORATOR].includes(roleName)
 
 const noRole = { role: 'NONE', labelKey: 'user.roles.noRole' }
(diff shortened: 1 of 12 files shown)


src/meta/user/users/index.ts
```diff
@@ -27,6 +27,7 @@ export const Users = {
   isCollaborator,
   isNationalCorrespondent,
   isPersonalInfoRequired,
+  isPersonalInfoRequiredForRole,
```
minotogna: Should we add this method to UserRoles instead ?

src/server/api/auth/loginGoogle.ts
```diff
@@ -11,6 +12,32 @@ import { ProcessEnv } from 'server/utils'
...
+const _getRedirectUrl = (props: RedirectUrlProps): string => {
+  const { assessmentName, countryIso, cycleName, msg } = props
+
+  const parsedMsg = msg?.message ? (JSON.parse(msg.message) as Record<string, string>) : {}
+  const { invitationUuid, requiresUserInfo } = parsedMsg
+  const invitedCountryIso = (parsedMsg.countryIso as AreaCode) ?? countryIso
+
+  let redirectPath: string
+  if (requiresUserInfo && invitationUuid) {
+    redirectPath = Routes.LoginInvitationUserInfo.generatePath({ assessmentName, cycleName, invitationUuid })
```
minotogna, line 31: do we redirect in the local as well ?

sorja: Handled on frontend

src/meta/user/roles/roles.ts
```diff
@@ -81,6 +84,7 @@ export const UserRoles = {
   noRole,
   getRecipientRoles,
   getLastRole,
+  isPersonalInfoRequiredForRole,
```
minotogna: I would remove forRole since this utility would work only on user roles

minotogna:
@sorja is this pr still valid or should we close it ?

PR #5894, opened 2026-05-06
5844 - Invitation page

sorja:


src/client/pages/Authentication/Authentication.tsx
```diff
@@ -16,7 +17,10 @@ const Authentication: React.FC = () => {
       <div className="login">
         <RouterRoutes>
           <Route element={<Login />} index />
-          <Route element={<Invitation />} path={Routes.LoginInvitation.path.relative} />
+          <Route path={Routes.LoginInvitation.path.relative}>
+            <Route element={<Invitation />} index />
```
minotogna: Don't we need only one AcceptInvitation page ?

src/client/pages/Authentication/Invitation/Invitation.tsx
```diff
@@ -39,14 +38,15 @@ const Invitation: React.FC = () => {
...
 
+  if (isInvitedUser) {
+    return <div>Accept invitation page</div>
+  }
+
   return (
     <div className="login-form">
-      {isInvitedUser && <Accept data={data} />}
-      {!isInvitedUser && <Register />}
+      <Register />
```
minotogna, line 49: I think the login form will eventually be moved out ? (And be only in the login page )

The invitation page should only include the accept invitation button + user form when needed .

Also the privacy notice should be also be moved to login when registration is needed ?

PR #5899, opened 2026-05-07
5758 - Navigation: Data errors - Descriptions - 3

yaguzmang:
Validating description text links by:
- Extracting code to get links to a common util `Htmls.getLinks(html)` (Didn't add this to the DOMs util because we can't use that one in the backend)
- Exporting `validateLink` from the visit links workflow and using it in `validateDescriptionText`
- Adding a new translation `invalidLinkWithReason` so we can show in the frontend why the link is invalid


src/server/controller/cycleData/validations/descriptions/validateDescriptionText.ts
```diff
@@ -12,11 +18,35 @@ type Props = {
...
-  void value
+  const links = Htmls.getLinks(value.text)
+  if (Objects.isEmpty(links)) return undefined
+
+  const linkValidations = await Promise.all(
+    links.map(async ({ link }) => ({
+      code: await validateLink(link),
+      link,
+    }))
+  )
```
yaguzmang: Here instead of calling `validateLink` in a loop like this we could do something like:
1. Adding a function `validateLinks`, which handles batching (just like we do here in [visitLinks](https://github.com/openforis/fra-platform/blob/development/src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/visitLinks.ts#L47 )).

2. Using that function here like this:
```ts
  const links = Htmls.getLinks(value.text)
  if (Objects.isEmpty(links)) return undefined

  const linkValidations = await validateLinks(links)
  const messages = linkValidations.reduce<Validation['messages']>((acc, { code, link }) => {
...
```
3. Refactoring `visitLinks` to use that common function too:
```ts
export const visitLinks = async (links: Array<LinkToVisit>): Promise<Array<VisitedLink>> => {
  const timestamp = Date.now().toString()
  const validatedLinks = await validateLinks(links)

  return validatedLinks.map((link) => ({ ...link, timestamp }))
}
```

I can do that though it makes this PR big and refactors some of the visit links workflow. This is the cleanest way I saw to validate because directly calling the visit links workflow introduces a lot of complexity, due to the fact that workflow was thought to be for country or assessment batches, not incremental validations for individual descriptions.

src/server/controller/cycleData/validations/descriptions/validateDescriptionText.ts
```diff
@@ -12,11 +18,35 @@ type Props = {
...
   const { value } = props
 
-  // TODO: Validate description text links through the existing links validation workflow.
-  void value
+  const links = Htmls.getLinks(value.text)
+  if (Objects.isEmpty(links)) return undefined
+
+  const linkValidations = await Promise.all(
+    links.map(async ({ link }) => ({
+      code: await validateLink(link),
```
minotogna: I remembered we mentioned validating links here might take 5 seconds ? We should post a request to the external worker to execute the link validation ?

yaguzmang: Yes tough 5s was worst case scenario I saw when validating links for an entire assessment. 
I did some testing and the highest I got was 846ms, but yes it could be more depending on the dns lookup resolution.
image

The current verify links visits all links in a country. If we want to use a queue, we would need to add a worker that's specific to descriptions, so it can run only with the few links passed to it, write to the descriptions validations cache,  and emit a socket event if there are errors. 
Or we could do the validation inline like we do here and add a max timeout (like 3s for example).

The queue system is more robust though it's much more code.

What do you think? if we want the queue, I can pivot this PR to that

minotogna: One request cannot take 3s , we'll crash the system .

If the worst case of validating one description takes more than 200ms we should use a separate job @yaguzmang

yaguzmang: Okay 👍 so this is the plan now: I removed this function and replaced it with 
`updateDescriptionTextValidations`: it will simply chedule the worker if the description text has links. 
The worker stuff I will do in following PRs so this one stays atomic.

minotogna: Nice 😊 thank you

src/server/controller/cycleData/validations/descriptions/updateDescriptionValidations.ts
```diff
@@ -31,19 +31,19 @@ export const updateDescriptionValidations = async (props: Props): Promise<void>
...
-    // TODO: Validate data sources and store results under sectionValidations.dataSources.
+    cachedValidations[sectionName] = sectionValidations
+
+    const { scheduled } = updateDescriptionTextValidations({ assessment, cycle, ...description })
 
-    if (!Objects.isEmpty(textValidation)) {
-      sectionValidations.descriptions ??= {}
-      sectionValidations.descriptions[descriptionName] = textValidation
-    } else {
+    if (!scheduled) {
```
minotogna: Should we really do anything ?
We won't anyway set any validation to this description , right ? 
We'll get the links validation from the links table , what do you think ?

yaguzmang: if the links table will is the source of truth, yes 👍 updated

minotogna (approved):

PR #5900, opened 2026-05-07
5758 - Navigation: Data errors - Descriptions - Redis repository

yaguzmang:
Adding redis set and get for descriptions


src/server/cache/repository/validation/getDescriptionValidations.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Objects } from 'utils/objects'
+
+import { getKeyCountry, Keys } from 'server/cache/repository/keys'
+import { RedisData } from 'server/cache/repository/redisData'
+
+type Props = {
+  assessment: Assessment
+  countryIso: CountryIso
+  cycle: Cycle
+  sectionNames?: Array<SectionName>
```
minotogna: should this be required instead ? (removing two ifs below) ?

yaguzmang: This is so that, if we don't pass sectionNames, we get all sections
We do the same in for example:
- getManyMetadata.ts (explorer)
- getTableValidations.ts

minotogna: Line 33 you return empty object , not all sections . Am I missing something maybe ?

yaguzmang: Yes, though line 33 returns empty [] if an empty sectionNames array is passed:
```
sectionNames === undefined  // fetch all sections
sectionNames === []         // fetch no sections, return {}
```

This is so that this doesn't through an error:
`const values = await redis.hmget(key, ...sectionNames)`

minotogna: That's why I asked the question . Should we always pass section names ? Why are we te question validation for no sections ?

yaguzmang: Sure, when we need all sections, I can query them first. For example when updating after the links job runs for the entire country.
I simplified it now. The check I added `Objects.isEmpty(sectionNames)` is defensive to avoid runtime errors.

minotogna (approved):

PR #5903, opened 2026-05-08
5844 - New user local flow

sorja:
- Fix also CountryHome/Collaborators link

https://github.com/user-attachments/assets/a5de7b0f-6d26-4a1e-9e7f-06f75176462a

### DIFF
--- a/src/client/pages/Authentication/Login/Login.tsx
+++ b/src/client/pages/Authentication/Login/Login.tsx
@@ -41,6 +41,7 @@ const Login: React.FC = () => {
   const invitationData = useGetInvitation()
   const { invitationUuid } = useSearchParams<LoginQueryParams>()
 
+  const isNewUser = Boolean(invitationData && Objects.isEmpty(invitationData.userProviders))
   const redirectUrl = useGetRedirectUrl(invitationData)
   const onSuccess = useOnSuccess({ redirectTo: redirectUrl })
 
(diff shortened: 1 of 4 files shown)


src/client/pages/Authentication/Login/Login.tsx
```diff
@@ -61,9 +62,10 @@ const Login: React.FC = () => {
           action={ApiEndPoint.Auth.login()}
           disableEmail={Boolean(invitationData)}
           email={invitationData?.user.email}
+          invitationUuid={isNewUser ? invitationUuid : undefined}
```
minotogna, line 65: What if is an invitation for an existing user ?

sorja: Only needed for new users here.
It is used in the backend to create auth providers

minotogna: Do we redirect client or server side , even for existing users ?

sorja: FE

minotogna: With google as well ? 
Should we try BE both google and local ?
From chat gpt 
image

sorja: Even with SPA you would use server side redirect?

sorja: If you want, I can do it

minotogna (approved):

PR #5914, opened 2026-05-12
5758 - Navigation: Data errors - Descriptions - 9 

yaguzmang:


src/server/worker/tasks/verifyLinks/visitDescriptionLinks/worker.ts
```diff
@@ -71,6 +73,17 @@ export default async (job: VerifyDescriptionLinksJob): Promise<void> => {
...
+    // Get the unique section names affected by this validation run.
+    const sectionNames = Array.from(
+      new Set(linksToVisit.flatMap(({ locations }) => locations.map(({ sectionName }) => sectionName)))
+    )
+    const eventName = Sockets.getDescriptionLinksValidationUpdateEvent({
+      assessmentName: assessment.props.name,
+      countryIso,
+      cycleName: cycle.name,
+    })
+    SocketServer.emit(eventName, { countryIso, descriptionIds, sectionNames })
```
yaguzmang: Since we will have the links table be the source of truth for description links validation, here we will emit description ids affected, so the client can re fetch their links validation state.

minotogna: can we pass the validation result instead of re fetching ?

yaguzmang: Added 👍 I will add the redis method in the next PR

src/server/worker/tasks/verifyLinks/visitDescriptionLinks/worker.ts
```diff
@@ -71,6 +75,21 @@ export default async (job: VerifyDescriptionLinksJob): Promise<void> => {
       linksToVisit,
     })
 
+    const descriptionValidations = buildDescriptionLinkValidations({ approvedLinks, linkVisits, linksToVisit })
+    // TODO Next PR: await ValidationRedisRepository.setDescriptionValidations({
```
minotogna, line 79: Should we also persist in database in table links ?

yaguzmang: Yes, we do it in line 71 with `upsertDescriptionLinks` ✌️

minotogna (approved):

PR #5915, opened 2026-05-13
5908 - Flags – Bulk download files (urgent issue)

### LINKED ISSUE #5908: Flags – Bulk download files (urgent issue)
@minotogna 

This is considered a high-priority issue.

**Is your feature request related to a problem? And its solution.**
Please add an individual flag column for each data entry point, and ensure that only one flag symbol (syntax below) is displayed per data entry point in the following files of the bulk download:

Files:
“Annual_yyyy_mm_dd” (6 new columns would be added to this file)
“FRA_Years_yyyy_mm_dd” (50+ columns would be added to this file)
“Intervals_yyyy_mm_dd” (4 new columns would be added)
Image

Flags (this also needs to be added as a legend on the **top right side** of the files):
- “A” = Normal value, official data reported by country/area;
- “I” = Imputed value, desk study, data compiled by FAO;
- “M” = Missing value, cannot exist, denominator is zero or missing;
- “O” = Missing value.

Syntax (only the preferred flag will be shown):
- The delineation of “I” and “A” flags is straight forward;
- “M” in case if forest area = 0.00, then the flags for _Growing stock (m³/ha) (2a)_, _Forest biomass (tonnes/ha) (2c)_, and _Forest carbon (tonnes/ha) (2d)_ should be “M” (regardless of whether the value is reported by a country or derived from a desk study);
- “O” If a country does not report a value and the information is missing;
- If it is a desk study and the information is missing, then “I” is sufficient and should be used on its own.

sorja:
ok'd by OJ reply

### DIFF
--- a/src/server/controller/cycleData/getBulkDownload/csvContent/_flag.ts
+++ b/src/server/controller/cycleData/getBulkDownload/csvContent/_flag.ts
@@ -1,0 +1,60 @@
+import { CountryIso } from 'meta/area/countryIso'
+import { AssessmentName } from 'meta/assessment/assessment'
+import { CycleName } from 'meta/assessment/cycle'
+import { TableName, TableNames } from 'meta/assessment/table'
+import { RecordAssessmentDatas } from 'meta/data/recordDatas'
+import { Numbers } from 'utils/numbers'
+
+import { BulkDownloadData } from 'server/controller/cycleData/getBulkDownload/types'
+
+const mFlagTables = new Set<TableName>([
+  TableNames.biomassStock,
+  TableNames.biomassStockAvg,
+  TableNames.carbonStock,
+  TableNames.carbonStockAvg,
+  TableNames.growingStockAvg,
+])
+
+type Props = {
+  assessmentName: AssessmentName
+  colName: string
+  countryIso: CountryIso
+  cycleName: CycleName
+  data: BulkDownloadData
+  deskStudy?: boolean
+  tableName: TableName
+  variableName: string
+}
+
+export const getFlag = (props: Props): string => {
+  const { assessmentName, colName, countryIso, cycleName, data, deskStudy, tableName, variableName } = props
+
+  if (mFlagTables.has(tableName)) {
+    const forestArea = RecordAssessmentDatas.getDatum({
+      assessmentName,
+      colName,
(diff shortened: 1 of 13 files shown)


src/server/controller/cycleData/getBulkDownload/metadata/_buildAnnualYears.ts
```diff
@@ -13,5 +13,5 @@ export const buildAnnualYears: BulkDownloadFilesFactory = (props) => {
   const years = Years.annual(cycle)
   const builders = [DisturbancesBuilder, AreaAffectedByFireBuilder]
 
-  return buildYears({ ...props, builders, fileName, includeDeskStudy, years })
+  return buildYears({ ...props, builders, fileName, includeDeskStudy, includeFlagLegend: true, withFlag: true, years })
```
minotogna: Why 2 Booleans include flag
And with flag ? One is not enough ?

src/server/controller/cycleData/getBulkDownload/metadata/_buildYears.ts
```diff
@@ -2,21 +2,52 @@ import { BulkDownloadFileYearsBuilderConstructor } from 'server/controller/cycle
...
+    includeDeskStudy,
+    includeFlagLegend,
+    tables,
+    withFlag,
+    years,
+  } = props
+
+  let csvPostProcessor: CSVPostProcessor | undefined
+  if (includeFlagLegend) {
+    const legendEntries = [
```
minotogna: Why here ? 
Better where the csv is generated ?

src/server/controller/cycleData/getBulkDownload/csvContent/_flag.ts
```diff
@@ -0,0 +1,60 @@
...
+    countryIso,
+    cycleName,
+    data: data.tables,
+    tableName,
+    variableName,
+  })
+
+  const isImputed = deskStudy || nodeValue?.faoEstimate
+  if (isImputed) return 'I'
+  if (nodeValue?.raw == null) return 'O'
```
minotogna: should we use Objects.isNil ?

minotogna (approved):

PR #5921, opened 2026-05-14
5917 - FAOSTAT Land area script

### LINKED ISSUE #5917: FAOSTAT Land area script import
- [ ] Create a CSV file with land area from current FAOSTAT for FRA latest 
- [ ] create tools/faostatLandArea/import.ts and move src/tools/migrations/steps/steps/utils/updateTotalLandArea.ts under it

sorja:


src/tools/faostat/updateTotalLandArea.ts
```diff
@@ -118,7 +114,7 @@ export const updateTotalLandArea = async (props: Props, client: BaseProtocol): P
           acc.push(query)
         }
         return acc
-      }, []).join(`; 
+      }, []).join(`;
     `)
       return client.query(query)
     })
```
minotogna: @sorja please. let's insert activity log as well of type ```ActivityLogMessage.nodeValueImport```

sorja: Ok.
For three years of land area, we will insert around ~700-800 rows.

sorja: For 1950...2025 we will add roughly 18750 rows

( 

cols/years to add per country: 2025 - 1950 = 75
countries roughly 250
250 * 75 = 18750

)

src/tools/faostat/import.ts
```diff
@@ -0,0 +1,90 @@
...
+import { AssessmentController } from 'server/controller/assessment'
+import { UserController } from 'server/controller/user'
+import { DB } from 'server/db/db'
+import { Logger } from 'server/utils/logger'
+
+// Usage:
+// Simply download the csv from fao stat and add it to ./csv/ and update __FILENAME__
+// https://www.fao.org/faostat/en/#data/RL
+
+const __FILENAME__ = 'FAOSTAT_data_en_5-14-2026.csv'
```
minotogna: any chance to have the file named `FAOSTAT_data.csv` only ? so that we don't need to manually update this file . what do you think ? or do you have another suggestion @sorja  ?

src/tools/faostat/import.ts
```diff
@@ -0,0 +1,90 @@
...
+const __LAST_YEAR__ = new Date().getFullYear()
+
+const assessmentName = 'fra'
+const cycleName = '2025'
```
minotogna: also, can we pass this cycle name from command line arg ? and throw error if not passed ? @sorja

src/tools/faostat/import.ts
sorja: Sure, I can change it so.
The idea was the download -> drag & drop file, but instead I can change the naming here

minotogna (approved):

PR #5930, opened 2026-05-20
5758 - Navigation: Data errors - Descriptions - Redux 

yaguzmang:


src/client/store/data/tableData/validations/slice/extraReducers/setDescriptionValidationsReducer.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (Objects.isEmpty(value)) {
+        delete currentValue[sectionName]
+      } else {
+        currentValue[sectionName] = value
+      }
+    })
+
+    Objects.setInPath({ obj: state.descriptions, path, value: currentValue })
+  })
+}
```
yaguzmang: This is to handle the flow where: A user edits a description and we trigger the links check job for that description only. In that case the worker returns partial `descriptionValidations` + `sectionNames`, so here we update only the affected sections instead of replacing the whole description validations state. We could standardize this to always send the full description cache too, but that would require an extra Redis read in `visitDescriptionLinks/worker.ts`.`

src/client/pages/Country/hooks/useDescriptionValidationsListener.ts
```diff
@@ -5,38 +5,47 @@ import { SectionName } from 'meta/assessment/section'
...
 import { useCountryRouteParams } from 'client/hooks/routeParams'
 import { SocketClient } from 'client/service/socket/client'
 
 type DescriptionValidationsListenerArgs = [
   {
-    countryIso: CountryIso
+    countryIso?: CountryIso
     descriptionValidations: RecordDescriptionValidations
-    sectionNames: Array<SectionName>
+    sectionNames?: Array<SectionName>
```
minotogna, line 17: Aren't we always passing sections from the socket ? I.e. this should be required .

src/client/pages/Country/hooks/useDescriptionValidationsListener.ts
```diff
@@ -5,38 +5,47 @@ import { SectionName } from 'meta/assessment/section'
...
+import { ValidationsActions } from 'client/store/data/tableData/validations/actions'
+import { useAppDispatch } from 'client/store/hooks'
 import { useCanEditCycleData } from 'client/store/user/hooks/auth'
 import { useCountryRouteParams } from 'client/hooks/routeParams'
 import { SocketClient } from 'client/service/socket/client'
 
 type DescriptionValidationsListenerArgs = [
   {
-    countryIso: CountryIso
+    countryIso?: CountryIso
```
minotogna: And this as well

src/client/store/data/tableData/validations/slice/extraReducers/setDescriptionValidationsReducer.ts
```diff
@@ -0,0 +1,54 @@
...
+  return value
+}
+
+export const setDescriptionValidationsReducer = (builder: ActionReducerMapBuilder<ValidationsState>): void => {
+  builder.addCase(setDescriptionValidations, (state, action) => {
+    const { assessmentName, countryIso, cycleName, descriptionValidations, sectionNames } = action.payload
+
+    const path = [assessmentName, cycleName, countryIso]
+
+    // Replace all description validations after a full refresh
```
minotogna: What's a full refresh ?
For me it's strange not to pass sections here . 
There should be another action if you want to reset the validations

src/client/store/data/tableData/validations/slice/extraReducers/setDescriptionValidationsReducer.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (Objects.isEmpty(sectionNames)) {
+      Objects.setInPath({ obj: state.descriptions, path, value: descriptionValidations })
+      return
+    }
+
+    // Update only the affected sections after a single link check
+    const currentValue = Objects.cloneDeep(Objects.getInPath(state.descriptions, path) ?? {})
+    const targetSectionNames = Array.from(new Set([...sectionNames, ...Object.keys(descriptionValidations)]))
+
```
minotogna, line 43: It seems over complicated to clone deep + iterate on all sections .

Can't we just update what's passed from the server ?

src/client/pages/Country/hooks/useDescriptionValidationsListener.ts
yaguzmang: There are two flows for description links validation:
- When the job to verify links for the entire Assessment or Country runs ("full refresh").
- When the user edits one description and we run the job for that description only.

When we run the the job for entire country, we process all descriptions, so we have all cache in hand and we send it in the event. There we don't need the section names because in the client we can just replace with all the data.

When we run the job for a single description, we do send the section name to know which singular entry to update in the client.

what do you think of those two flows? I thought for a second maybe we could send all the cache in both flows and then in the client we just replace everything always, no need to iterate. But maybe that's too much data to send after just editing one description.

minotogna: Add some comments then .

yaguzmang: Added 👍 , let me know if I can add anything else to make it better to follow

minotogna (approved):

PR #5935, opened 2026-05-21
5931 - EditorWYSIWYG - Pass validations from outside

### LINKED ISSUE #5931: EditorWYSIWYG pass validations from outside

yaguzmang:


src/client/components/EditorWYSIWYG/hooks/_processPaste.ts
```diff
@@ -3,61 +3,63 @@ import type { IJodit } from 'jodit/esm/types/jodit'
...
-    console.error('Error processing paste:', error)
-    return html
+      return textParts.join(' ')
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error processing paste:', error)
+      return html
+    }
   }
-}
```
yaguzmang, line 63: Easier to read diff:

Image

I edited 
`const href = node.getAttribute('href')` to 
`const href = node.getAttribute('href')?.trim() ?? ''`

Why:
Pasting for example `<a>Broken Example</a>`, was resulting on this value pasted:
`<a href="null" rel="nofollow" target="_blank">Broken Example</a>`
Which is an issue because `href="null"` doesn't really mean "empty href".

Now instead we paste:
`<a href="" rel="nofollow">Broken Example</a>`

yaguzmang:
This PR got too big so I'll close it and split it up in smaller PRs.

PR #5940, opened 2026-05-21
5931 - EditorWYSIWYG - Pass validations from outside - CellLink

yaguzmang:
https://github.com/user-attachments/assets/b4757347-7b3d-43a8-af88-92bccda6f3ed

### DIFF
--- a/src/client/components/EditorWYSIWYG/hooks/useValidationTooltip.tsx
+++ b/src/client/components/EditorWYSIWYG/hooks/useValidationTooltip.tsx
@@ -28,6 +28,12 @@ export const useValidationTooltip = (props: Props): Returned => {
       }
     }
 
+    if (validationErrors.length === 1) {
+      const dataTooltipHtml = ReactDOMServer.renderToStaticMarkup(<div>{validationErrors[0]}</div>)
+
+      return { dataTooltipHtml, dataTooltipId: TooltipId.error, hasValidationErrors }
+    }
+
     const dataTooltipHtml = ReactDOMServer.renderToStaticMarkup(
       <ul>
         {validationErrors.map((validationError, index) => (
--- a/src/client/components/TableNodeExt/CellNodeExt/CellLink/CellLink.tsx
+++ b/src/client/components/TableNodeExt/CellNodeExt/CellLink/CellLink.tsx
@@ -2,7 +2,10 @@ import React from 'react'
 
 import { NodeExtCellType } from 'meta/nodeExt/cellType'
 
+import { useCanEditCycleData } from 'client/store/user/hooks/auth'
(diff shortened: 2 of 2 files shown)


src/client/components/TableNodeExt/CellNodeExt/CellLink/CellLink.tsx
```diff
@@ -2,17 +2,25 @@ import React from 'react'
...
   const { disabled, nodeExt, onChange } = props
 
+  const canEditCycleData = useCanEditCycleData()
+  const { print } = useIsPrintRoute()
   const value = nodeExt?.value?.raw ?? ''
+  const validationErrors = useLinkValidationErrors({ enabled: canEditCycleData && !print, value })
 
-  return <EditorWYSIWYGLinks disabled={disabled} onChange={onChange} value={value} />
+  return (
+    <EditorWYSIWYGLinks disabled={disabled} onChange={onChange} validationErrors={validationErrors} value={value} />
```
yaguzmang, line 22: @minotogna @sorja The other place to add this empty link validation is in `TextLinkField`, using  `useRoleValidationSchema`. Should I add it there too? or will that collide with your task?

src/client/components/EditorWYSIWYG/hooks/useValidationTooltip.tsx
```diff
@@ -28,6 +28,12 @@ export const useValidationTooltip = (props: Props): Returned => {
       }
     }
 
+    if (validationErrors.length === 1) {
+      const dataTooltipHtml = ReactDOMServer.renderToStaticMarkup(<div>{validationErrors[0]}</div>)
+
+      return { dataTooltipHtml, dataTooltipId: TooltipId.error, hasValidationErrors }
+    }
+
```
yaguzmang, line 36: To match how it's currently implemented, I'm removing the list style when there is just one error:
This:
Image
Instead of this:

Image

src/client/components/TableNodeExt/CellNodeExt/CellLink/CellLink.tsx
minotogna: Wait @yaguzmang 
That hook is used outside the country route . Let's talk Monday also with @sorja .
We can have the daily Monday , what do you think ?

yaguzmang: I will be away on Monday, but maybe we can talk today?

minotogna: > I will be away on Monday, but maybe we can talk today? 

Tuesday @yaguzmang @sorja ?

minotogna (approved):

PR #5945, opened 2026-05-26
5944 - Fix: User table limited to 30 users

### LINKED ISSUE #5944: User table limited to 30 users
Users table (TablePaginated) defaults to 30 entries.

sorja:
Note: This targets development branch (critical bug fix)

TablePaginated was defaulting to 30 entries/page.
This limited collaborators view to only 30 items

kuva
kuva


src/client/components/TablePaginated/TablePaginated.tsx
```diff
@@ -42,7 +42,7 @@ const defaults: Readonly<Partial<Props<object>>> = {
 const TablePaginated = <Datum extends object>(props: Props<Datum>): React.ReactElement => {
   const { className, gridTemplateColumns: gridTemplateColumnsProps } = props // HTMLDivElement Props
   const { marginPagesDisplayed, pageRangeDisplayed } = props // Paginator Props
-  const { columns, filterFn, filters = defaults.filters, groups, limit = 30, path } = props // Base Props
+  const { columns, filterFn, filters = defaults.filters, groups, limit = 30, noLimit, path } = props // Base Props
```
minotogna: Can we just pass as prop to limit : number | 'NO' 
And the you pass undefined to the BE ? (We avoid an additional prop . )
What do you think ?

sorja: should we pass -1 ? it was the other option i thought so we can keep it as limit: number

let me know, either is fine

(noLimit added because it's more human readable)

minotogna: yes! -1 it is ;) @sorja

minotogna (dismissed):

sorja:
@minotogna fixed skeleton

minotogna (approved):

PR #5949, opened 2026-05-27
5844 - Register: Show video tutorials and update e2e

sorja:
image

### DIFF
--- a/src/client/pages/Authentication/Authentication.scss
+++ b/src/client/pages/Authentication/Authentication.scss
@@ -81,7 +81,7 @@ $login-input-width: $login-form-width - (2 * $spacing-s);
   }
 
   .button {
-    width: $login-input-width;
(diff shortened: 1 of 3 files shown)


src/client/pages/Authentication/Authentication.scss
```diff
@@ -78,6 +80,10 @@ $login-height: 50vh;
     background-color: $ui-bg-hover;
   }
 
+  .button {
+    width: $login-input-width;
```
minotogna: Can they be less wide ?

sorja: image

Better like this?

minotogna: Yes @sorja . Very nice 👍

minotogna (approved):

PR #5954, opened 2026-05-29
5758 - Navigation: Data errors - Add removeLocations query

yaguzmang:
As explained in slack, moving the locations to a new table is a medium/big refactor. So here I am adding this draft PR to show how instead of doing that, we could use a query "removeLocations" which we can run when we delete a link in a description. That way we keep the links table up to date. If we still want to split the links table now, I can close this PR.

### DIFF
--- a/src/server/db/repository/assessmentCycle/links/removeLocations.ts
+++ b/src/server/db/repository/assessmentCycle/links/removeLocations.ts
@@ -1,0 +1,75 @@
+import { CountryIso } from 'meta/area/countryIso'
+import { Assessment } from 'meta/assessment/assessment'
+import { Cycle } from 'meta/assessment/cycle'
+import { Objects } from 'utils/objects'
+
+import { BaseProtocol, DB } from 'server/db/db'
+import { Schemas } from 'server/db/schemas'
+
+type LocationToRemove = {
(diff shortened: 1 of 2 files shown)


src/server/db/repository/assessmentCycle/links/removeLocations.ts
```diff
@@ -0,0 +1,75 @@
...
+            ),
+            '[]'::jsonb
+          ) as locations
+        from ${schemaCycle}.link
+             cross join lateral jsonb_array_elements(coalesce(link.locations, '[]'::jsonb))
+               with ordinality as location(value, ordinality)
+        where link.country_iso = $(countryIso)
+        group by link.id
+      )
+      update ${schemaCycle}.link as link
```
minotogna, line 58: Should we also remove links that have no more locations ?

yaguzmang: Yes. For the links we do soft deletion. There is a prop called 'deleted', and in line 65-66 we set it to true if there are no remaining locations: `to_jsonb(jsonb_array_length(updated_links.locations) = 0)`

minotogna (approved):

PR #5958, opened 2026-05-29
5844 - Update validations translation on language change

sorja:


src/client/components/Form/Form.tsx
```diff
@@ -50,6 +52,10 @@ const Form: React.FC<FormProps> = (props) => {
   const { errors, isDirty, isSubmitting } = formState
   const watchValues = watch()
 
+  useEffect(() => {
+    trigger()
+  }, [i18n.language, trigger])
```
minotogna: Can we use useLanguage ?

minotogna (approved):

PR #5962, opened 2026-05-31
5758 - Navigation: Data errors - Descriptions - Add get description validations endpoint 

yaguzmang:
Adding an endpoint to fetch description validations, following the same pattern we use for getTableValidations. These are loaded per section, so we fetch only section the user is viewing.

After this, I will add the redux code to load, and after that I will actually display the errors in the FE.

### DIFF
--- a/src/meta/api/endpoint/cycleData.ts
+++ b/src/meta/api/endpoint/cycleData.ts
@@ -89,7 +89,7 @@ export const CycleData = {
   },
 
   Validations: {
-    descriptionData: (): string => apiPath('cycle-data', 'validations', 'description-data'),
+    descriptions: (): string => apiPath('cycle-data', 'validations', 'descriptions'),
     summary: (): string => apiPath('cycle-data', 'validations', 'summary'),
     tableData: (): string => apiPath('cycle-data', 'validations', 'table-data'),
   },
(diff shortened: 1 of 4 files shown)


src/meta/api/endpoint/cycleData.ts
```diff
@@ -89,6 +89,7 @@ export const CycleData = {
   },
 
   Validations: {
+    descriptionData: (): string => apiPath('cycle-data', 'validations', 'description-data'),
```
minotogna: The word descriptions is enough , not description-data

src/server/controller/cycleData/index.ts
```diff
@@ -64,6 +65,7 @@ export const CycleDataController = {
   updateOriginalDataPointYear,
 
   // ===== review
+  getDescriptionValidations: DescriptionValidationRedisRepository.getDescriptionValidations,
```
minotogna: Description validations and table validations should not go under review section , but rather under their Own domain . I.e. descriptions under descriptions , table data under table data

yaguzmang: Done 👍

minotogna (approved):

PR #5964, opened 2026-06-01
5844 - PanEuropean with no form accepting invitation, fix validation error rendering on first load, redirect 

sorja:
- Update fixtures with pan eu '25
- test pan eu invitation
- redirect login correctly


src/client/components/Form/Form.tsx
```diff
@@ -52,7 +52,13 @@ const Form: React.FC<FormProps> = (props) => {
   const { errors, isDirty, isSubmitting } = formState
   const watchValues = watch()
 
+  // Trigger validations on language change after first render
+  const isMounted = useRef(false)
```
minotogna, line 56: Do we have a hook useIsMounted already ?

sorja: I tried it, but in this case it doesn't work as expected:

```
  const isMounted = useIsMounted()
  useEffect(() => {
    if (!isMounted.current) return
    trigger()
  }, [isMounted, language, trigger])

```

=> it triggers rerenders and validation errors are shown.

I'd keep as it is with useRef

minotogna (approved):

PR #5993, opened 2026-06-04
5844 - Invitation Form Skeleton

sorja:
https://github.com/user-attachments/assets/57a990f6-d5b9-4468-8e9e-a9645cb94739

With also skeleton for google button:

image

### DIFF
--- a/src/client/pages/Authentication/Login/Login.tsx
+++ b/src/client/pages/Authentication/Login/Login.tsx
@@ -20,9 +20,20 @@ import ButtonGoogle from 'client/pages/Authentication/Login/ButtonGoogle'
 import { useRedirect } from 'client/pages/Authentication/Login/hooks/useRedirect'
 import { videoResources } from 'client/pages/Tutorials'
 
-export const useGetInvitation = (): InvitationData => {
+export const useGetInvitation = (): {
+  error: unknown
+  invitationData: InvitationData
+  loaded: boolean
+  loading: boolean
+} => {
   const { invitationUuid } = useSearchParams<LoginQueryParams>()
-  const { data: invitationData, dispatch: fetchData } = useGetRequest(ApiEndPoint.User.invitation(), {
+  const {
+    data: invitationData,
+    dispatch: fetchData,
+    error,
+    loaded,
+    loading,
+  } = useGetRequest(ApiEndPoint.User.invitation(), {
     params: { invitationUuid },
   })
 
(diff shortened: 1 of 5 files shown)


src/client/pages/Authentication/Login/Login.tsx
```diff
@@ -1,70 +1,53 @@
...
-import { useGetRequest } from 'client/hooks/getRequest'
 import { useCycleRouteParams } from 'client/hooks/routeParams'
 import { useSearchParams } from 'client/hooks/searchParams'
 import Icon from 'client/components/Icon'
 import Divider from 'client/pages/Authentication/Divider'
 import FormLogin from 'client/pages/Authentication/FormLogin'
 import { useOnSuccess } from 'client/pages/Authentication/FormLogin/hooks/useOnSuccess'
 import ButtonGoogle from 'client/pages/Authentication/Login/ButtonGoogle'
+import { useGetInvitation } from 'client/pages/Authentication/Login/hooks/useGetInvitation'
 import { useRedirect } from 'client/pages/Authentication/Login/hooks/useRedirect'
```
minotogna: relative imports ?

src/client/pages/Authentication/Login/Login.tsx
```diff
@@ -1,70 +1,53 @@
...
 
-  const { redirectUrl } = useRedirect({ invitationData })
+  const hasInvitationUuid = !Objects.isEmpty(invitationUuid)
+  const { redirectUrl } = useRedirect({ error, invitationData, loaded })
   const onSuccess = useOnSuccess({ redirectTo: redirectUrl })
 
   return (
     <div className="login-form">
-      {invitationUuid && !invitationData && <div>Add a FormLoginSkeleton</div>}
+      <FormLogin
```
minotogna, line 42: sorry @sorja where is the skeleton? 😅

src/client/pages/Authentication/Login/hooks/useRedirect.ts
```diff
@@ -39,5 +46,21 @@ export const useRedirect = (props: Props): Returned => {
     }
   }, [isDifferentUser, navigate, t, toaster])
 
+  const redirectHandled = useRef(false)
+
+  useEffect(() => {
+    if (redirectHandled.current) return
```
minotogna: this code can't go to the hook above ?

src/client/pages/Authentication/Login/Login.tsx
sorja: Automatically handled by FormLogin when loading is set to true 😃

minotogna: ok.... post a video in case ;)

src/client/pages/Authentication/Login/hooks/useRedirect.ts
sorja: added all to same useEffect and also added missing expired invitation

all cases should be ok now

- no invitation (wrong uuid)
- already accepted
- expired
- wrong user

src/client/pages/Authentication/Login/Login.tsx
sorja: added in pr desc

src/client/pages/Authentication/Login/Login.tsx
minotogna, line 73: if it's loading a skeleton here as well ? the button google should not be rendered anyway

sorja:
Try prod without useRef

src/client/pages/Authentication/Login/hooks/useRedirect.ts
```diff
@@ -1,43 +1,70 @@
...
       navigate(redirectUrl, { replace: true })
     }
   }, [isSameUser, navigate, redirectUrl])
 
+  const redirectHandled = useRef(false)
+
   useEffect(() => {
+    if (redirectHandled.current) return
+
     if (isDifferentUser) {
```
minotogna: the if else should only create the toast error message and route.
Then if toaster message and route exists , ->
```
type Redirect = {errorMessage:string; path:string }
redirectHandled.current = true
toaster.error(redirect.errorMessage)
navigate(redirect.path, { replace: true })
```

src/client/pages/Authentication/Login/hooks/useRedirect.ts
minotogna: better to have only one hook

minotogna (approved):

minotogna (changes requested):
wrongly approved

sorja:
@minotogna Even in prod build, it triggers it twice:

image


I will leave useRef

minotogna:
> @minotogna Even in prod build, it triggers it twice:
> 
> 
> 
> image
> 
> 
> 
> 
> 
> I will leave useRef

The hook is triggered twice then .
Can u check which params get updated ? @sorja

sorja:
> > @minotogna Even in prod build, it triggers it twice:
> > image
> > I will leave useRef
> 
> The hook is triggered twice then . Can u check which params get updated ? @sorja

Seems like it's the `react-router`s `navigate` fn, should i remove it from the deps?

minotogna:
> > > @minotogna Even in prod build, it triggers it twice:
> 
> > > image
> 
> > > I will leave useRef
> 
> > 
> 
> > The hook is triggered twice then . Can u check which params get updated ? @sorja
> 
> 
> 
> Seems like it's the `react-router`s `navigate` fn, should i remove it from the deps?

You can ref it @sorja ?

sorja:
> > > > @minotogna Even in prod build, it triggers it twice:
> > 
> > 
> > > > image
> > 
> > 
> > > > I will leave useRef
> > 
> > 
> > > 
> > 
> > 
> > > The hook is triggered twice then . Can u check which params get updated ? @sorja
> > 
> > 
> > Seems like it's the `react-router`s `navigate` fn, should i remove it from the deps?
> 
> You can ref it @sorja ?

Investigated a bit more:

Problem fixed by `return null` when invitation not found in backend.

I added the `throw error` when invitation not found which caused the double error message (1 from server, 1 from client)

👍 

So I had accidentally added twice the same error message..!

minotogna:
> > > > > @minotogna Even in prod build, it triggers it twice:
> 
> > > 
> 
> > > 
> 
> > > > > image
> 
> > > 
> 
> > > 
> 
> > > > > I will leave useRef
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > The hook is triggered twice then . Can u check which params get updated ? @sorja
> 
> > > 
> 
> > > 
> 
> > > Seems like it's the `react-router`s `navigate` fn, should i remove it from the deps?
> 
> > 
> 
> > You can ref it @sorja ?
> 
> 
> 
> Investigated a bit more:
> 
> 
> 
> Problem fixed by `return null` when invitation not found in backend.
> 
> 
> 
> I added the `throw error` when invitation not found which caused the double error message (1 from server, 1 from client)
> 
> 
> 
> 👍 
> 
> 
> 
> So I had accidentally added twice the same error message..!

Okkkk 👍

minotogna (approved):

PR #5996, opened 2026-06-04
5758 - Navigation: Data errors - Data sources - Update summary

yaguzmang:
In both frontend and backend we apply updates to the validations data when a new result is calculated. So, here I am adding a couple of helpers to handle that, while also considering the data sources in the logic.

Next PR: Display errors.


src/meta/assessment/validation/descriptionValidations/index.ts
```diff
@@ -0,0 +1,77 @@
...
+      acc[descriptionName] = { valid: textValid && dataSourcesValid }
+      return acc
+    }
+
+    acc[descriptionName] = { valid: textValid }
+    return acc
+  }, {} as ValidationSummaryDescription)
+}
+
+export const DescriptionValidations = {
```
minotogna, line 4: I suggest to export sub methods in sub files , like we do for our utilities.

src/server/cache/repository/validation/description/setDescriptionValidations.ts
```diff
@@ -43,7 +31,7 @@ export const setDescriptionValidations = async (props: Props): Promise<void> =>
   const validationsToSet = sectionNames.reduce<Record<string, string>>((acc, sectionName) => {
     const current = currentValidations[sectionName] ?? {}
     const update = descriptionValidations[sectionName]
-    const value = _mergeSectionDescriptionValidations(current, update)
+    const value = DescriptionValidations.applySectionUpdate({ current, update })
```
minotogna: I would call the method mergeValidations ? that's what it is, right ?

yaguzmang: Yes, renamed it now 👍

minotogna (approved):

PR #6005, opened 2026-06-05
5758 - Navigation: Data errors - Reference links - Clean link utils code

yaguzmang:
Before updating the descriptions worker to validate reference links, I'm cleaning up the links utils. I will add more, so better split it in different files:

- getI18nValidationStatusLabelKey -> same code in a different file.
- getLocationLabel -> same code in a different file.
- getKey -> new util used in a couple of places to get a link unique key.

I'm also adding: 
```ts
export const DescriptionLinkLocationPath = {
  dataSourceReference: ['dataSources', 'reference'],
  text: ['text'],
}
```

So we don't have to manually write ['dataSources', 'reference'] and ['text'] locations.


src/meta/cycleData/links/getLocationLabel.ts
```diff
@@ -0,0 +1,64 @@
...
+import { TableNames } from 'meta/assessment/table'
+import { LinkLocation } from 'meta/cycleData/links/link'
+
+import { ODPCommentColumns } from 'server/db/repository/assessmentCycle/originalDataPoint/commentColumns'
+
+type GetLocationLabelProps = {
+  countryIso: CountryIso
+  cycle: Cycle
+  includeCountryIso?: boolean
+  isPanEuropean: boolean
```
minotogna: Better to pass assessment instead ?

src/meta/cycleData/links/getLocationLabel.ts
```diff
@@ -0,0 +1,64 @@
...
+    return includeCountryIso ? `${countryIso} - ${label}` : label
+  }
+
+  const { descriptionName, path } = location
+
+  const descriptionLabelKey = path.includes('reference')
+    ? 'dataSource.referenceToTataSource'
+    : Descriptions.getLabelKey({
+        name: descriptionName as CommentableDescriptionName,
+        isPanEuropean,
```
minotogna: Also here

src/meta/cycleData/links/getLocationLabel.ts
yaguzmang: Yes, added 👍

minotogna (approved):

PR #6006, opened 2026-06-06
5758 - Navigation: Data errors - Reference links - Validate in worker

yaguzmang:
Handling the reference links in the worker, by adding them to the list of links to validate and by updating the reference validation under dataSources. 
Here we also validate that the reference is not empty. 

https://github.com/user-attachments/assets/bf8035b9-32d3-4e40-9dda-ee6063caf79d

### DIFF
--- a/src/meta/cycleData/links/isLocationPath.ts
+++ b/src/meta/cycleData/links/isLocationPath.ts
@@ -1,3 +1,0 @@
-export const isLocationPath = (path: Array<string>, locationPath: Array<string>): boolean => {
-  return path.length === locationPath.length && path.every((pathItem, index) => pathItem === locationPath[index])
-}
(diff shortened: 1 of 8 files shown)


src/meta/cycleData/links/isLocationPath.ts
```diff
@@ -0,0 +1,3 @@
+export const isLocationPath = (path: Array<string>, locationPath: Array<string>): boolean => {
+  return path.length === locationPath.length && path.every((pathItem, index) => pathItem === locationPath[index])
```
minotogna: This seems to compare two arrays . Check if we can use isEquals from lodash (and then let's add Objects.isEquals wrapper )

yaguzmang: Yes we can, and we already have `Objects.isEqual`, so even no need to add `isLocationPath` to wrap it. Thanks for pointing it out 👍

minotogna (approved):

PR #6007, opened 2026-06-08
6001: Dependencies upgrade: react-tooltip -> CountryMultiSelect

minotogna:
#### data-tooltip-html must be replaced with children
https://react-tooltip.com/docs/upgrade-guide/changelog-v5-v6#replace-data-tooltip-html-with-tooltip-children

image

#### PR # 1 CountryMultiSelect

- [x] Added client/components/CountryMultiSelect/TooltipCountries
- [x] Updated old **useTooltipContent** by removing html content and renamed to **useTooltipProps**

https://github.com/user-attachments/assets/774276bc-9f0c-47ca-8d53-b6fe26787809

https://github.com/user-attachments/assets/1858c046-178a-4782-8437-36d798182640


src/client/components/CountryMultiSelect/CountryMultiSelect.tsx
```diff
@@ -68,6 +62,15 @@ const CountryMultiSelect: React.FC<Props> = (props) => {
...
+      <TooltipCountries
+        allowAtlantis={allowAtlantis}
+        allowedCountries={allowedCountries}
+        canDisplayTooltip={canDisplayTooltip}
+        error={error}
+        isMulti={isMulti}
+        tooltipId={dataTooltipId}
+        value={value}
+      />
     </div>
```
sorja, line 74: Would it be better if we had a custom component we can use to wrap around elements?

```suggestion
      <TooltipCountries
        allowAtlantis={allowAtlantis}
        allowedCountries={allowedCountries}
        canDisplayTooltip={canDisplayTooltip}
        error={error}
        isMulti={isMulti}
        tooltipId={dataTooltipId}
        value={value}>
 <Select
        // eslint-disable-next-line react/jsx-props-no-spreading
        {...otherProps}
        toggleAll
        value={value}
      />
</Tooltip>
```

minotogna: Sure I can try ... facing another problem now . Tooltip behind grid ...

sorja: Sure, don't worry.

My comment more related to the generic way of using Tooltip anyway (thinking about the future: what if we replace it later, how should we be using it _now_?)

minotogna: @sorja i don't think it's easy to have a generic component , especially for country selector

sorja (approved):

PR #6012, opened 2026-06-08
5758 - Navigation: Data errors - Remove buildEmail isPanEuropean prop for link locations

yaguzmang:
Updating the code introduced in #6008 because we pass `assessment` instead of `isPanEuropean` in the updated `Links.getLocationLabel`

PR #6013, opened 2026-06-08
5758 - Navigation: Data errors - Descriptions - Remove getManyByIds

yaguzmang:
Replacing `getManyByIds` with `getValue`. Replacing `name` with `names`.

Also, removing the description ids from the links flow.


yaguzmang:
Tests will pass after we merge [5758 - Navigation: Data errors - Remove buildEmail isPanEuropean prop for link locations](https://github.com/openforis/fra-platform/pull/6012)

src/server/db/repository/assessmentCycle/descriptions/getMany.ts
```diff
@@ -0,0 +1,42 @@
...
+  assessment: Assessment
+  countryIso: CountryIso
+  cycle: Cycle
+  descriptionTargets: Array<{
+    name: CommentableDescriptionName
+    sectionName: SectionName
+  }>
+}
+
+export const getMany = async (props: Props, client: BaseProtocol = DB): Promise<Array<CommentableDescription>> => {
```
minotogna: can you use getValues instead ? what do you think ?

src/server/worker/tasks/verifyLinks/visitDescriptionLinks/props.ts
```diff
@@ -3,14 +3,19 @@ import { Job } from 'bullmq'
...
+import { SectionName } from 'meta/assessment/section'
 
 import { VerifyLinksJobName } from 'server/worker/tasks/verifyLinks/jobNames'
 
 export type VerifyDescriptionLinksJobProps = {
   assessment: Assessment
   countryIso: CountryIso
   cycle: Cycle
-  descriptionIds: Array<number>
+  descriptionTargets: Array<{
```
minotogna: I have seen this type duplicated over this PR. duplicated code should be avoided .
Let's introduce a type 
```
DescriptionIdentifier = {
    name: CommentableDescriptionName
    sectionName: SectionName
  }
``` ?

File: src/server/db/repository/assessmentCycle/descriptions/getMany.ts (hunk 1 above)
Comment (yaguzmang) [reply src/server/db/repository/assessmentCycle/descriptions/getMany.ts]: In getValues we don't return the full row like we do here, it returns something like 

```ts
{
  FIN: {
    forestCharacteristics: {
      generalComments: { text: '...' }
    }
  }
}
```

we can't reuse it because we need the description id for the links validation (we need it to build the link location). 

Also, I meant to ask:  later when we add the descriptions to the validateAll flow, we will need to query all descriptions in a country. Should we keep this as getMany and have `descriptionTargets` as optional? if not passed -> gets all descriptions in the country. If passed -> we query the requested records.
What do you think?

minotogna: Why are we still using id for the links validation ? Can we remove those and use only get values method ?

src/server/controller/cycleData/links/update.ts
```diff
@@ -39,10 +39,14 @@ export const update = async (props: Props): Promise<Link> => {
   })
 
   // If the link has description locations, we trigger the flow that updates validation cache.
-  const descriptionIds = updatedLink.locations.filter(Links.isDescriptionLocation).map(({ id }) => id)
-  if (!Objects.isEmpty(descriptionIds)) {
+  const descriptionLocations = updatedLink.locations.filter(Links.isDescriptionLocation)
+  if (!Objects.isEmpty(descriptionLocations)) {
     const { countryIso } = updatedLink
-    await visitDescriptionLinks({ assessment, countryIso, cycle, descriptionIds })
+    const descriptionTargets = descriptionLocations.map(({ descriptionName, sectionName }) => ({
```
minotogna: Add type to map

src/server/controller/cycleData/validations/descriptions/updateDescriptionLinkValidations.ts
```diff
@@ -23,6 +23,6 @@ export const updateDescriptionLinkValidations = async (props: Props): Promise<vo
     assessment,
     countryIso,
     cycle,
-    descriptionIds: descriptions.map(({ id }) => id),
+    descriptionTargets: descriptions.map(({ name, sectionName }) => ({ name, sectionName })),
```
minotogna: Add type to map

src/server/worker/tasks/verifyLinks/visitDescriptionLinks/utils/syncDescriptionLinks.ts
```diff
@@ -33,10 +34,10 @@ type Returned = {
 // 2. Visit the links that are not approved.
 // 3. When the descriptions still contain links, update the database in one transaction: remove old locations and save the current ones.
 export const syncDescriptionLinks = async (props: Props): Promise<Returned> => {
-  const { assessment, countryIso, cycle, descriptionIds, linksToVisit: rawLinksToVisit } = props
+  const { assessment, countryIso, cycle, descriptions, linksToVisit: rawLinksToVisit } = props
 
   const locationPaths = [DescriptionLinkLocationPath.text, DescriptionLinkLocationPath.dataSourceReference]
-  const locations: Array<LocationToRefresh> = descriptionIds.flatMap((id) =>
+  const locations: Array<LocationToRefresh> = descriptions.flatMap(({ id }) =>
```
minotogna: Add type to map not to locations

src/server/db/repository/assessmentCycle/descriptions/getMany.ts
yaguzmang: In the links table, we save locations like this:
```ts
type LinkLocationBase = {
  id: number // Here we store the id of the description
  url: string
}

export type DescriptionsLocation = LinkLocationBase & {
  colName: string
  descriptionName: CommentableDescriptionName
  path: Array<string>
  sectionName: string
  uuid?: string
}
```
With some refactoring we can remove the id though.


---
Looking into `DescriptionRepository.getValues`, the current props are these:
```ts
type Props = {
  assessment: Assessment
  countryISOs: Array<CountryIso>
  cycle: Cycle
  name?: CommentableDescriptionName
  sectionNames?: Array<SectionName>
}

```

With that shape we can't really query by sectionName/name pairs. 
But we can add the prop `descriptionIdentifiers` and use it optionally:

```ts
type Props = {
  assessment: Assessment
  countryISOs: Array<CountryIso>
  cycle: Cycle
  descriptionIdentifiers?: Array<DescriptionIdentifier>
  name?: CommentableDescriptionName
  sectionNames?: Array<SectionName>
}
```
That way we don't need to refactor the existing consumers of getValues. What do you think?

src/meta/assessment/originalDataPoint/originalDataPoint.ts
```diff
@@ -34,6 +35,7 @@ export interface OriginalDataPoint {
   dataSourceReferences?: string
   id: number
   nationalClasses?: Array<ODPNationalClass>
+  readonly uuid?: UUID
```
yaguzmang, line 38: Adding this so the code builds and the tests pass. It will be updated from optional to required when we merge the new ODP type shape.

minotogna (approved):

PR #6020, opened 2026-06-09
5994 - Repository error messages: displayed under the field

### LINKED ISSUE #5994: Repository error messages: displayed under the field
Image

- [ ] Add client side validation 
- [ ] Disable save while uploading 
- [ ] Toaster should anyway appear on top (check z index ?)

sorja:
Validation shows correctly:

image

Validation is cleared correctly:

image

Button is disabled correctly:

image

Note: No more notification as the validation has been moved to frontend.
If for some reason the user was able to send the malformed request, the popup would now show on top of the panel.

```
$z-index-toaster: 10000; <----- This was changed from 25 to 10000 - the highest value in the platform now.
$z-index-overlay-panel: 9998;
$z-index-loader: 9999;
```


src/client/style/_partials/_config.scss
```diff
@@ -71,7 +71,7 @@ $z-index-area-selector: 10;
 $z-index-panel-review: 15;
 $z-index-popover: 20;
 $z-index-tooltip: 21;
-$z-index-toaster: 25;
+$z-index-toaster: 10000;
```
minotogna: put it under loader

minotogna (approved):

PR #6037, opened 2026-06-11
3596 v2 history

sorja:
image

- Show history for V2
- Fix migration UUIDs

Developers note: `type: string | Array<string>` caused some problems with the diffLines which I tried to get around


sorja:
@minotogna 

From my side this PR is ready
Can you be so kind as to test it locally?

After if it's ok, you can `Ready for review` and approve.

Thanks,

minotogna:
> @minotogna 
> 
> 
> 
> From my side this PR is ready
> 
> Can you be so kind as to test it locally?
> 
> 
> 
> After if it's ok, you can `Ready for review` and approve.
> 
> 
> 
> Thanks,
> 
> 

Sure , thanks @sorja 
I'll check later

src/tools/migrations/steps/steps/20260613000000-step-odp-format-v2.ts
```diff
@@ -77,33 +77,42 @@ export default async (): Promise<void> => {
...
-                       al.target
-                       , '{dataSources}'
-                       , jsonb_build_array(jsonb_build_object(
-                   'uuid', uuid_generate_v5(uuid_nil(), concat('ndp_', al.target ->> 'id'::text)),
+    // 5. update activity log ndp format, using the UUID already stored in descriptions (step 3)
+    await DB.query(`
+      with uuids as (
+        select odp.id as odp_id
+             , coalesce(d.value -> 0 ->> 'uuid', uuid_generate_v5(uuid_nil(), concat('ndp_', odp.id))::text) as uuid
+        from ${schemaName}.original_data_point odp
```
minotogna: @sorja what if the odp has been deleted ? 
we need to fetch them from activity log, right ?

minotogna: plus @sorja this update is executed for every cycle....not what we want. why did you update this migration ?

sorja: we need to have real uuids per odp datasource otherwise history will show it as new (uuid compare)
odps are per cycle

minotogna: How long did it take to run locally this @sorja ?

Now I understand the reason, but I propose we use uuid v5 when we insert the datasources as well. would that work?

minotogna (approved):

PR #6038, opened 2026-06-11
5758 - Navigation: Data errors - Descriptions - validateAll

yaguzmang:
Validating descriptions in `validateAll`


src/server/controller/cycleData/validations/descriptions/updateDataSourceFieldValidations.ts
```diff
@@ -1,20 +1,22 @@
...
-import { CommentableDescription, DataSource } from 'meta/assessment/descriptionValue'
+import { DataSource } from 'meta/assessment/descriptionValue'
 import { RecordDescriptionValidations } from 'meta/assessment/validation/description'
 import { Validation } from 'meta/assessment/validation/validation'
 import { Sockets } from 'meta/socket/sockets'
 import { Objects } from 'utils/objects'
 
 import { DescriptionValidationRedisRepository } from 'server/cache/repository/validation/description'
 import { SocketServer } from 'server/service/socket'
+import { DescriptionLinkSource } from 'server/worker/tasks/verifyLinks/visitDescriptionLinks/types'
```
yaguzmang: This type `DescriptionLinkSource` is basically the same as `CommentableDescription` but without the id.

``` ts
export interface CommentableDescription {
  id: number
  countryIso: CountryIso
  sectionName: string
  name: CommentableDescriptionName
  value: CommentableDescriptionValue
}
```

We don't really use the description id anywhere besides raw queries. Should we remove it from `CommentableDescription`, and add something like `CommentableDescriptionRow` in case we need it later?

``` ts
type CommentableDescriptionRow = CommentableDescription & {
  id: number
}
```

minotogna: @yaguzmang we should not duplicate the code. The DescriptionLinkSource must be removed and we will investigate later on to remove the id from CommentableDescription type in a separate task.

yaguzmang: Removed `DescriptionLinkSource` and replaced it with `Omit<CommentableDescription, 'id'>` 👍

src/server/controller/cycleData/validations/descriptions/validateCountryDescriptions.ts
```diff
@@ -0,0 +1,38 @@
...
+import { updateDescriptionValidations } from 'server/controller/cycleData/validations/descriptions/updateDescriptionValidations'
+
+type Props = {
+  assessment: Assessment
+  country: Country
+  cycle: Cycle
+  descriptionsByCountry: DescriptionCountryValues
+}
+
+export const validateCountryDescriptions = async (props: Props): Promise<void> => {
```
minotogna, line 20: should this file go under server/controller/cycleData/validations/descriptions/ folder ?
It's still about descriptions

yaguzmang: In this folder `country` we also have `validateCountryTables.ts` and a its helper `buildTablesNodeUpdates.ts`. Seems a good place for country-wide operations. Or we could move all files to other places

minotogna: They should be grouped by domain . Domain are tables and descriptions

yaguzmang: Done :+1: Moved the files to `tables/` and `descriptions/`. I will move more files to `tables/` next PR so it's more organized.

minotogna (approved):

PR #6059, opened 2026-06-16
6050: Split TableDataApi + DescriptionsApi

minotogna:


src/server/api/cycleData/table/tableDataApi.ts
```diff
@@ -0,0 +1,34 @@
+import { Express } from 'express'
+// @ts-ignore
```
sorja, line 2: Why is this needed?

sorja (approved):

PR #6071, opened 2026-06-17
6070 - Country Recent Activity: User Removed

### LINKED ISSUE #6070: Country Recent Activity: User Removed
this should apply when admin or RFC removes a user role from the country user page 

Add recent activity entry : 
$User removed $RremovedUser as $Role

sorja:
image

Activity log (aligned with other acitivty logs for user (e.g. invite, accept))
```
{
  "role": "COLLABORATOR",
  "user": "Example",
  "userUuid": "dafa9248-f639-41f6-9e2c-cc3910b6a0c9",
  "countryIso": "ITA"
}
```

### DIFF
--- a/src/server/controller/user/removeRole.ts
+++ b/src/server/controller/user/removeRole.ts
@@ -24,12 +24,8 @@ export const removeRole = async (props: Props, client: BaseProtocol = DB): Promi
     const role = Users.getRole(target, countryIso, cycle)
     await UserRoleRepository.remove({ uuid: role.uuid }, t)
 
-    // @ts-ignore
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const { _: roles, ...userNoRoles } = user
-
     const activityLog = {
-      target: { user: userNoRoles, countryIso, role: role.role },
+      target: { userUuid: target.uuid, user: target.props.name, countryIso, role: role.role },
       section: 'users',
       message: ActivityLogMessage.userRoleDeleteRole,
       countryIso,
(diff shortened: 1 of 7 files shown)


src/server/controller/user/removeRole.ts
```diff
@@ -24,12 +24,8 @@ export const removeRole = async (props: Props, client: BaseProtocol = DB): Promi
     const role = Users.getRole(target, countryIso, cycle)
     await UserRoleRepository.remove({ uuid: role.uuid }, t)
 
-    // @ts-ignore
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const { _: roles, ...userNoRoles } = user
-
     const activityLog = {
-      target: { user: userNoRoles, countryIso, role: role.role },
+      target: { userUuid: target.uuid, user: target.props.name, countryIso, role: role.role },
```
sorja: @minotogna 

BUG:
Previously we persisted `currentUser` in the activity log and not `targetUser`.
Therefore the activity log is malformed for these entries.
Malformed: `user = targetUser` in column `target`.

There is no way to get the original data (I think).

Luckily, there are only 8 entries

What should we do?
Delete the malfored activity log entry?

minotogna: @sorja can u try to fix the data? E.g. check prev cycle, invitations ?

sorja: I have been trying, there is no clues of what was the target unfortunately, but I will continue debugging.

In case, what do you suggest?

minotogna (approved):

PR #6077, opened 2026-06-17
5758 - Navigation: Data errors - Display backend validation in the new data sources components

yaguzmang:
Adding the prop `validationErrors` to the new common Data Source components. This replaces the `validator` prop as we now validate backend side, not client side. 

image


src/client/components/DataSources/hooks/useComponentsOrder.ts
```diff
@@ -1,10 +1,10 @@
 import { useMemo } from 'react'
 
-import { DataSource } from 'meta/assessment/descriptionValue/dataSource'
+import { DataSourceValidationErrors } from 'meta/assessment/descriptionValue/dataSource'
 
 import { PropsDataSources } from 'client/components/DataSources/types'
 
-type Returned = Array<keyof DataSource>
+type Returned = Array<keyof DataSourceValidationErrors>
```
minotogna: Why this ? This shouldn't change , the components are based on data source fields , not validation

yaguzmang: I was avoiding a cast here: `const content = validationErrors[componentKey]`, since `componentKey` typed as `Array<keyof DataSource>` can be other things also like `uuid` or `type`.

But now I left as it was and simply added the cast:
`const content = validationErrors[componentKey as keyof DataSourceValidationErrors]`

minotogna: Errors can be key as well ?

yaguzmang: image


Typescript complains because `DataSourceValidationErrors` has `comments`, `reference`, `type`, `variables`, `year`, while the `DataSource` type has those and also `uuid`, `placeholder`.

We could add a common type `DataSourceComponentKey` for data source properties that have a component and are validated:
```ts
type DataSourceComponentKey = keyof Pick<DataSource, 'reference' | 'type' | 'variables' | 'year' | 'comments'>

componentKey: DataSourceComponentKey
```
Which makes sense because for `uuid` and `placeholder` there aren't any components, so this definition  `componentKey: keyof DataSource` is misleading.

What do you think?

minotogna: `DataSourceEditableField` is more appropriate as name

yaguzmang: Added `DataSourceEditableField` 👍

minotogna (approved):

PR #6078, opened 2026-06-18
5758 - Navigation: Data errors - ODP validation types

yaguzmang:


src/meta/assessment/validation/odp.ts
```diff
@@ -0,0 +1,26 @@
...
+export type ODPDataSourceValidation = Partial<Record<ODPDataSourceValidationField, Validation>>
+
+export type ODPValidation = {
+  comments?: Partial<Record<OriginalDataPointCommentKey, Validation>>
+  dataSources?: Record<UUID, ODPDataSourceValidation>
+  nationalClasses?: Record<UUID, ODPNationalClassValidation>
+  year?: Validation
+}
+
+export type RecordODPValidations = Record<UUID, ODPValidation>
```
minotogna: We now use NDP (cc @sorja ) 
Let's use national data point from now on 
We'll need to go through all code afterwards

minotogna (approved):

PR #6079, opened 2026-06-18
5758 - Navigation: Data errors - Add initial validateNationalDataPoint

yaguzmang:
Adding the initial `validateNationalDataPoint` function.


src/meta/assessment/odps/validateODP.ts
```diff
@@ -72,7 +72,7 @@ export const validateNationalClass = (originalDataPoint: OriginalDataPoint, inde
   if (!nationalClass) return { error: false } as NationalClassValidation
   const isPlaceHolder = nationalClass.placeHolder
   const validClassName = isPlaceHolder || _validateClassName(nationalClass)
-  const validArea = validClassName || _validateArea(nationalClass)
+  const validArea = isPlaceHolder || _validateArea(nationalClass)
```
yaguzmang, line 75: The comment above `_validateArea` says:

`// == Area must not be null`

So seems like this line had a bug
```ts
const validArea = validClassName || _validateArea(nationalClass)
```
because if the national class name is valid (not empty and below 1024 chrs), then the `_validateArea` doesn't even run. We could have null areas without errors.

minotogna: from BE, we don't pass placeholder , right?

src/server/controller/cycleData/validations/nationalDataPoint/validateNationalDataPoint.ts
```diff
@@ -0,0 +1,32 @@
...
+import { NDPValidation } from 'meta/assessment/validation/ndp'
+import { Objects } from 'utils/objects'
+
+import { buildNationalClassValidation } from './buildNationalClassValidation'
+
+type Props = {
+  nationalDataPoint: OriginalDataPoint
+}
+
+export const validateNationalDataPoint = (props: Props): NDPValidation => {
```
minotogna, line 9: I think all validation method should be moved outside old odps object

`NationalDataPointValidator.validate(ndp)
NationalDataPointValidator.validateNationalClasses(ndp)
NationalDataPointValidator.validateNationalClass(ndp)
NationalDataPointValidator.validateYear(ndp)
`
I suspect these are the 4 methods we will expose. 
However, since we don't know yet, I recommend to follow the flow.
i.e. in the task, the first issue reads 
` 
BE: Validate all ODPs and store them in redis
`
therefore we start from there. 
To validate all ODPs now we need to expose only validateNationalDataPoint.

src/server/controller/cycleData/validations/ndp/buildNationalClassValidation.ts
```diff
@@ -0,0 +1,70 @@
...
+): Validation | undefined => {
+  const { messageKey, nationalClassName, valid } = props
+
+  if (valid) return undefined
+
+  const message = { key: `generalValidation.${messageKey}`, params: { name: nationalClassName, value: '100%' } }
+  return { valid: false, messages: [message] }
+}
+
+export const buildNationalClassValidation = (props: Props): NDPNationalClassValidation => {
```
minotogna: we only want to have validateNationalClass.

minotogna (changes requested):

src/meta/assessment/odps/validateODP.ts
yaguzmang: Yes, we do skip placeholder in the BE

src/server/controller/cycleData/validations/nationalDataPoint/validateNationalDataPoint.ts
yaguzmang: Okay, I added `validateCountryNationalDataPoints` and `NationalDataPointValidator.validate`.

The flow is now:
`validateAll` -> `validateCountryNationalDataPoints` -> `validateNationalDataPoint` -> `NationalDataPointValidator.validate`

So this PR doesn't become too big, in the following PRs I will:
1. Move year and national-class validation logic out of the `ODPs` object. I left TODOs where the current ODPs calls are made.
2. Add the Redis validation repository method to persist the NDP validations.

src/meta/assessment/validation/nationalDataPoint/validate.ts
```diff
@@ -0,0 +1,32 @@
+import { ODPs } from 'meta/assessment/odps'
+import { ODPNationalClass, OriginalDataPoint } from 'meta/assessment/originalDataPoint'
+import { validateNationalClass } from 'meta/assessment/validation/nationalDataPoint/validateNationalClass'
```
minotogna: Either ndp or nationalDataPoint folder ,
Not both .
Validator can be exposed from its file (or folder ), not index

yaguzmang: Fixed it 👍 . I added a `nationalDataPointValidator` folder

src/meta/assessment/odps/validateODP.ts
minotogna: So the check will need to be removed ?

minotogna (approved):

src/meta/assessment/odps/validateODP.ts
yaguzmang: Since we will not use the old `ODPs` validations anymore, I guess we will end up removing this entire file anyway. I started adding and using the new NDP validations in PR #6087

PR #6088, opened 2026-06-20
5758 - Navigation: Data errors - NDP - Add redis setNationalDataPointValidations

yaguzmang:
Adding redis setNationalDataPointValidations.

Still left to do for this item ` BE: Validate all ODPs and store them in redis`:
- Validate NDP percentages.


src/server/cache/repository/validation/nationalDataPoint/index.ts
```diff
@@ -0,0 +1,5 @@
+import { setNationalDataPointValidations } from 'server/cache/repository/validation/nationalDataPoint/setNationalDataPointValidations'
+
+export const NationalDataPointValidationRedisRepository = {
+  setNationalDataPointValidations,
+}
```
yaguzmang, line 5: Added `index.ts` here because every main object we export in `src/server/cache/repository` we export from a barrel file. Shoud we avoid barrel files here too?

minotogna: this is not a barrel file. we export one object only

minotogna (approved):

PR #6092, opened 2026-06-23
5758 - Navigation: Data errors - NDP - Calculate summary

yaguzmang:


src/server/cache/repository/validation/nationalDataPoint/index.ts
```diff
@@ -1,5 +1,7 @@
+import { getNationalDataPointValidations } from 'server/cache/repository/validation/nationalDataPoint/getNationalDataPointValidations'
 import { setNationalDataPointValidations } from 'server/cache/repository/validation/nationalDataPoint/setNationalDataPointValidations'
 
 export const NationalDataPointValidationRedisRepository = {
+  getNationalDataPointValidations,
```
minotogna: The two methods can simply be called get and set validations.  Since the domain is already the NDP ...

yaguzmang: Updated this and the other validation repositories 👍

minotogna (approved):

PR #6095, opened 2026-06-23
5758 - Navigation: Data errors - Rename migration step & skip NDP validation in panEu

yaguzmang:
Renaming the migration step so it runs last, and skipping the NDP validations in panEuropean


src/server/controller/cycleData/validations/nationalDataPoint/validateCountryNationalDataPoints.ts
```diff
@@ -22,6 +22,8 @@ export const validateCountryNationalDataPoints = async (
   const { assessment, country, cycle } = props
   const { countryIso } = country
 
+  if (!cycle.props.ndp) return {}
```
minotogna: Better if we don't even invoke this method . The check should be done from the caller

minotogna (approved):

PR #6100, opened 2026-06-25
5758 - Navigation: Data errors - NDP - Get endpoint & initial redux

yaguzmang:


src/client/pages/Country/SectionWrapper/SectionWrapper.tsx
```diff
@@ -15,6 +16,7 @@ const SectionWrapper: React.FC<PropsWithChildren> = (props) => {
   const sections = useSections()
   const tableSections = useTableSections({ sectionName })
   useReviewStatusListener()
+  useGetNationalDataPointValidations()
```
yaguzmang, line 23: This seemed like the best place to call this hook because we need to fetch NDP validations if we are in section 1a/1b or if we are in the NDP page. The NDP page is a sibling of the section page, and the common component between them is this`SectionWrapper.tsx`

Any suggestions as to where to call the hook?

minotogna: We should call these where we call the table validations .
Where do we call table validations ?

yaguzmang: We call the table validations from ` src/client/pages/Section/Section.tsx`. If we call `useGetNationalDataPointValidations` from there, then we should also call it from `src/client/pages/OriginalDataPoint/OriginalDataPoint.tsx`.  Otherwise, when a user navigates straight to the NDP page or refreshes it, the validations won't be fetched. 

I can add the call from both components, let me know ✌️

src/client/store/meta/hooks/sections.ts
```diff
@@ -21,7 +21,7 @@ export const useSection = (sectionNameParam?: string): SubSection => {
 
   return useMemo<SubSection>(() => {
     let subSection: SubSection
-    for (let i = 0; i < sections.length; i += 1) {
+    for (let i = 0; i < sections?.length; i += 1) {
```
yaguzmang, line 24: This became needed after moving validation fetch into `SectionWrapper`. Those hooks call `useCanEdit` -> `useSection`, and `useSections()` can still be undefined on the initial render.

minotogna (approved):

PR #6104, opened 2026-06-29
5758 - Navigation: Data errors - NDP - Mark error in header cell

yaguzmang:
image

image

Regular tooltip unchanged 
image


src/client/pages/Country/SectionWrapper/SectionWrapper.tsx
```diff
@@ -21,6 +22,7 @@ const SectionWrapper: React.FC<PropsWithChildren> = (props) => {
   useGetTableValidations({ sectionName })
   useGetDescriptionValidations({ sectionName })
   useGetNationalDataPointValidations()
+  useGetOriginalDataPointReservedYears()
```
minotogna: Why here ?

src/i18n/resources/en/nationalDataPoint.json
```diff
@@ -10,6 +10,7 @@
   "categories": "Categories",
   "class": "Class",
   "clickOnNDP": "Click on the year to access original data",
+  "clickOnNDPWithErrors": "The original data has errors. Click on the year to open it",
```
minotogna: I don't know the message now . I don't think we should change it for now.
Let's use the same message with or without errors now . We go back to this later on maybe

minotogna (approved):

minotogna (changes requested):
Previous message

minotogna:
> Previous message

Sorry @yaguzmang i wrongly approved it before . See my previous messages

src/client/pages/Country/SectionWrapper/SectionWrapper.tsx
yaguzmang: We need it or the NDP validations, so I'm calling it next to `useGetNationalDataPointValidations`

src/i18n/resources/en/nationalDataPoint.json
yaguzmang: image

Done 👍

src/client/pages/Country/SectionWrapper/SectionWrapper.tsx
minotogna: Why do we need it for NDP validations ? Isn't the validation server side @yaguzmang

src/client/pages/Section/DataTable/Table/GridHeadCell/OdpHeaderCell/hooks/useOdpHeaderTooltip.ts
```diff
@@ -0,0 +1,38 @@
...
+  id: TooltipId
+}
+
+export const useOdpHeaderTooltip = (props: Props): OdpHeaderTooltip => {
+  const { odpId } = props
+
+  const { t } = useTranslation()
+  const historyLastApprovedIsActive = useHistoryLastApprovedIsActive()
+  const reservedYears = useOriginalDataPointReservedYears() ?? []
+  const uuid = reservedYears.find((reservedYear) => reservedYear.id === odpId)?.uuid
```
minotogna: I see now @yaguzmang 
Can we instead add the uuid in the data ? (If it's too complicated , then the id in the validation )
I don't like the idea to fetch the reserved years in section wrapper and inject the NDP slice there ... 
What do you think ?

yaguzmang: Yes, I was just writing:

> The validation itself is server-side, yes. We need reserved years only on the client to map the table header `odpId` to the NDP uuid.

> `OdpHeaderCell` only has `odpYear.id`, but NDP validations in Redux are keyed by `uuid`. The reserved-years payload gives us the id -> uuid mapping

I would go with id in the validation because adding the UUID in the data would require edits to the DDL, queries and a migration. 

Since NDP validations are typed like this:
`RecordNDPValidations = Record<UUID, NDPValidation>`

I would add the id here:

```ts
export type NDPValidation = {
  odpId?: OriginalDataPoint['id'] // <---
  comments?: Partial<Record<OriginalDataPointCommentKey, Validation>>
  dataSources?: Record<UUID, NDPDataSourceValidation>
  nationalClasses?: Record<UUID, NDPNationalClassValidation>
  year?: Validation
}
```


What do you think?

yaguzmang: Or, maybe cleaner:
```ts
export type NDPValidationMeta = {
  odpId: number
}

export type NDPValidation = {
  comments?: Partial<Record<OriginalDataPointCommentKey, Validation>>
  dataSources?: Record<UUID, NDPDataSourceValidation>
  meta?: NDPValidationMeta
  nationalClasses?: Record<UUID, NDPNationalClassValidation>
  year?: Validation
}
```

yaguzmang: I Added it with 
```ts 
export type NDPValidationMeta = {
  odpId: number
}
```

Let me know what you think

src/client/store/data/tableData/validations/hooks/nationalDataPoints.ts
```diff
@@ -30,6 +30,29 @@ export const useNationalDataPointValidation = (props: NationalDataPointValidatio
   })
 }
 
+type NationalDataPointValidationByOdpIdProps = {
```
minotogna: This can be called Props

src/client/store/data/tableData/validations/hooks/nationalDataPoints.ts
```diff
@@ -30,6 +30,29 @@ export const useNationalDataPointValidation = (props: NationalDataPointValidatio
...
+  odpId?: number
+}
+
+export const useNationalDataPointValidationByOdpId = (
```
minotogna: Is there another hook that returns the validation ? If not , no need to use byId

src/client/store/data/tableData/validations/selectors/index.ts
```diff
@@ -21,6 +22,7 @@ export const ValidationsSelectors = {
   getDescriptionValidation,
   getNationalClassValidation,
   getNationalDataPointValidation,
```
minotogna, line 24: Is this one used ?

src/server/controller/cycleData/validations/nationalDataPoint/validateYear.ts
```diff
@@ -26,10 +26,13 @@ export const validateYear = async (props: Props): Promise<void> => {
     uuid,
   })
 
-  const updatedValidation = NationalDataPointValidator.validateYear({
-    nationalDataPoint,
-    validation,
-  })
+  const updatedValidation = {
+    ...NationalDataPointValidator.validateYear({
```
minotogna: Can the odpId be returned from the validator method directly ?
And do we need the meta property ?

src/client/store/data/tableData/validations/selectors/index.ts
yaguzmang: Yes it is used when we get the validation with uuid, in `useYearErrorTooltip` for example.

I now removed `useNationalDataPointValidationByOdpId`
instead 
`useNationalDataPointValidation` takes `odpId` or  `uuid`, so we have only one hook

src/server/controller/cycleData/validations/nationalDataPoint/validateYear.ts
yaguzmang: The validator now return the odpId and I removed the meta property, it wasn't strictly necessary 👍

minotogna (approved):

PR #6116, opened 2026-07-01
5758 - Navigation: Data errors - NDP - Fix unresolved validation target on NDP year columns

yaguzmang:
Fixing these issues when toggling `Use national data points`

image

and when deleting and NDP or NDP class:
image

More info in the comments.


src/server/api/area/updateCountryProp.ts
```diff
@@ -44,9 +46,13 @@ export const updateCountryProp = async (req: Request, res: Response): Promise<vo
...
 
       // 3. schedule update dependencies
+      const variablesByTable = AssessmentMetaCaches.getVariablesByTables({ assessment, cycle })
       const nodeUpdates = Object.entries(tableData).reduce<NodeUpdates>(
         (acc, [colName, recordRowData]) => {
           Object.entries(recordRowData).forEach(([variableName, value]) => {
+            // The merged NDP data also carries variables that don't belong to this table (e.g. total, totalLandArea);
+            // only queue variables that are part of the table metadata.
+            if (Objects.isNil(variablesByTable[tableName]?.[variableName])) return
             acc.nodes.push({ tableName, variableName, colName, value })
```
yaguzmang: Found an issue while testing NDP, toggling `Use national data points`:

Since before the new BE validations, we were building node updates for every variable in the merged NDP data, which included variables that don't belong to table `forestCharacteristics` (total, totalLandArea, forestArea, ...). 

Now when we validate, we throw an error when targets can't be resolved. So, this bug was now causing this error: `Could not resolve validation target forestCharacteristics.total.2019` when toggling `Use national data points`. 
After the fix we only queue variables that are part of the table metadata.

PS: perhaps a more correct fix would be something like:

```ts
    if (Object.keys(countryProp).includes(SectionNames.forestCharacteristics)) {
      const odpYears = await DataRedisRepository.getODPYears({ assessment, cycle, countryIso })
      const variables = getVariables({ cycle, sectionName: SectionNames.forestCharacteristics })

      const nodes: Array<NodeUpdate> = []
      odpYears.forEach((colName) => {
        variables.forEach(({ tableName, variableName }) => {
          nodes.push({ tableName, variableName, colName, value: undefined })
        })
      })

      const nodeUpdates = { assessmentName, cycleName, countryIso, nodes }
      await updateDependents({ assessment, cycle, country: updatedCountry, isODP: true, nodeUpdates, user }, DB)
    }
```

That is similar to how `updateOriginalDataPointsDependentNodes` already builds the nodes when an ODP is saved, so the toggle would follow the same path instead of depending on what the merged data has. 

Any thoughts? @minotogna cc @sorja

src/server/controller/cycleData/validations/tables/validateNodeUpdates.ts
```diff
@@ -48,12 +50,18 @@ export const validateNodeUpdates = async (props: Props): Promise<Array<TableName
...
+    // NDP years that are not reporting years have no col in the row metadata,
+    // so a missing col is expected: validate with the row-level validateFns.
+    const isNDPYearTarget = Objects.isNil(col) && isODPVariable(cycle, variable)
+
+    if (Objects.isNil(row) || (Objects.isNil(col) && !isNDPYearTarget)) {
       throw new Error(`Could not resolve validation target ${tableName}.${variableName}.${colName}`)
     }
 
+    const validateFns = col?.props.validateFns?.[cycle.uuid] ?? row.props.validateFns?.[cycle.uuid]
+
```
yaguzmang, line 58: Found another issue while testing NDP: 
When a data point has a year that is not a reporting year (e.g. 2024), its node updates are queued with that year as `colName` (`extentOfForest.forestArea.2024`). That column doesn't exist in the row metadata, so this was throwing `Could not resolve validation target extentOfForest.forestArea.2024` when editing an ODP.

In that case we expect the `col` to be missing, so now we only throw if the target is not an NDP variable.

src/server/api/area/updateCountryProp.ts
minotogna: @yaguzmang we need to speak tue about this . Not sure now .
Can u go ahead ?

yaguzmang: Okay :+1: I put the PR in draft for now 
Yes, I can work around it. The next PR is a little related, It's about the issue I mentioned in slack: toggling Use ( or don't use) national data points. But I think I can work around this problem here 👍

minotogna: > Okay :+1: I put the PR in draft for now 
> 
> Yes, I can work around it. The next PR is a little related, It's about the issue I mentioned in slack: toggling Use ( or don't use) national data points. But I think I can work around this problem here 👍  

Anything else you can work on besides this toggle @yaguzmang ?

yaguzmang: Yes, NDP links

src/server/controller/cycleData/validations/tables/validateNodeUpdates.ts
```diff
@@ -48,12 +48,14 @@ export const validateNodeUpdates = async (props: Props): Promise<Array<TableName
 
     const row = rows[RowCaches.getKey({ tableName, variableName })]
     const col = row?.cols?.find((candidate: Col) => candidate.props.colName === colName)
-    const validateFns = col?.props.validateFns?.[cycle.uuid] ?? row?.props.validateFns?.[cycle.uuid]
-
+    // Same behavior as updateCalculationDependencies: queued targets that don't map
+    // to a concrete metadata column are not processed.
     if (Objects.isNil(row) || Objects.isNil(col)) {
-      throw new Error(`Could not resolve validation target ${tableName}.${variableName}.${colName}`)
+      return
```
yaguzmang, line 54: @minotogna Simplified to early return

minotogna (approved):

PR #6119, opened 2026-07-04
5758 - Navigation: Data errors - NDP - Add syncNationalDataPointLinks

yaguzmang:
Adding `syncNationalDataPointLinks`, which visits NDP links and keeps the `link` table in sync by removing stale locations and saving the current ones.

In this PR:
- Extracted a common `syncLinks.ts` used by descriptions and NDPs
- Renamed `upsertDescriptionLinks` to a generic `upsertLinks`
- Added location key types for matching description and NDP link locations
- Updated `removeLocations` so it can also remove NDP locations
- Cleaned all existing link table locations so we don't keep any with the old shape


src/meta/cycleData/links/linkLocationBase.ts
```diff
@@ -1,4 +1,3 @@
 export type LinkLocationBase = {
-  identifier?: string
```
yaguzmang, line 2: We weren't really using this for descriptions, so better to remove it and have an explicit `odpUuid: UUID` in `NationalDataPointLinkLocation`

src/meta/cycleData/links/nationalDataPointLink.ts
```diff
@@ -12,14 +12,19 @@ export const NDPLinkFields: Array<NDPLinkField> = Object.values(NDPLinkField)
...
   fields: Array<NDPLinkField>
-  year: number
 }
 
-// `identifier` from LinkLocationBase holds the odp uuid
 export type NationalDataPointLinkLocation = LinkLocationBase & {
   sectionName: 'originalDataPoint'
-  odpSection: string
+  odpSection: NDPLinkField
+  odpUuid: UUID
```
minotogna: Should we use ndp prefix ?

yaguzmang: Yes, updated these now 👍

minotogna (approved):

PR #6123, opened 2026-07-06
5758 - Navigation: Data errors - NDP - Trigger NDP link validation & Display errors

yaguzmang:
The NDP links job now runs whenever NDP links can change, and the errors are shown in the NDP page.

In this PR:

- Editing a 1a/1b comment or the data sources refreshes the link validations of the edited field
- Changing an NDP's year refreshes its link locations, because we need to update the year and a year-based url location
- Deleting an NDP removes its link locations
- Approving or unapproving a link refreshes the NDPs where the link is located
- Added `mergeTargets` to the worker: a link can be located in several fields of the same NDP, and the job should refresh each NDP only once
- The NDP comments and data sources now display the errors from the BE
- Link validations now carry the ndp id, which the component `OdpHeaderCell` needs 

https://github.com/user-attachments/assets/8c5cb31b-d1fe-4362-a6f4-eb66cccb483e

Coming next: Making the full links job reuse the logic from the ndp and description jobs so it handles validations as well.

### DIFF
--- a/src/client/pages/OriginalDataPoint/components/Comments/hooks/useValidationErrors.ts
+++ b/src/client/pages/OriginalDataPoint/components/Comments/hooks/useValidationErrors.ts
@@ -14,7 +14,9 @@ type Props = {
   field: OriginalDataPointCommentKey
 }
 
-export const useValidationErrors = (props: Props): Array<string> => {
+type Returned = Array<string>
+
+export const useValidationErrors = (props: Props): Returned => {
   const { field } = props
   const { t } = useTranslation()
   const canEditCycleData = useCanEditCycleData()
@@ -24,7 +26,7 @@ export const useValidationErrors = (props: Props): Array<string> => {
   const validation = useNationalDataPointValidation({ uuid: nationalDataPoint.uuid })
   const commentValidation = validation.comments?.[field]
 
-  return useMemo<Array<string>>(() => {
+  return useMemo<Returned>(() => {
     if (!canEditCycleData || print || displayHistory) return []
 
     return MessageParser.getMessages(t, commentValidation)
--- a/src/client/pages/OriginalDataPoint/components/Comments/hooks/useValidationErrors.ts
+++ b/src/client/pages/OriginalDataPoint/components/Comments/hooks/useValidationErrors.ts
@@ -1,0 +1,32 @@
+import { useMemo } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import { OriginalDataPointCommentKey } from 'meta/assessment/originalDataPoint'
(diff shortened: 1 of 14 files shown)


src/client/pages/OriginalDataPoint/components/Comments/hooks/useValidationErrors.ts
```diff
@@ -0,0 +1,34 @@
...
+type Props = {
+  field: OriginalDataPointCommentKey
+}
+
+type Returned = Array<string>
+
+export const useValidationErrors = (props: Props): Returned => {
+  const { field } = props
+  const { t } = useTranslation()
+  const canEditCycleData = useCanEditCycleData()
```
minotogna: I think here there's a specific can edit odp hook

src/client/pages/OriginalDataPoint/components/DataSources/DataSourcesV2/hooks/useValidationErrors.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useODPDisplayHistory } from 'client/pages/OriginalDataPoint/components/hooks/useODPDisplayHistory'
+
+type Props = {
+  nationalDataPoint: OriginalDataPoint
+}
+
+export const useValidationErrors = (props: Props): DataSourceValidationErrorsRecord => {
+  const { nationalDataPoint } = props
+  const { t } = useTranslation()
+  const canEditCycleData = useCanEditCycleData()
```
minotogna: And here as well

minotogna: Also I see a lot of common code with previous hook 
As always , we should avoid duplicated code

src/client/pages/OriginalDataPoint/components/Comments/hooks/useValidationErrors.ts
yaguzmang: Using now `const canEdit = useIsEditODPDescriptionEnabled()` 👍

src/client/pages/OriginalDataPoint/components/DataSources/DataSourcesV2/hooks/useValidationErrors.ts
yaguzmang: Using here now `const canEdit = useIsEditODPEnabled()` 👍 
I also added `useShowNDPValidationErrors` so the code reads cleaner. There isn't much more to deduplicate because `useValidationErrors` here in `DataSourcesV2` loops over the data sources validation record, and the hook in `Comments` simply returns the message directly

minotogna (approved):

PR #6125, opened 2026-07-07
5758 - Navigation: Data errors - visitCycleLinks - Update descriptions code

yaguzmang:
First step to update `visitCycleLinks`: making the descriptions code reusable for it. 

- Adding a common `notifyDescriptionValidationUpdate` like in NDP links.
- Removing `const currentValidations = await getValidations` from the redis set method, like we discussed while working on NDP

### DIFF
--- a/src/server/worker/tasks/verifyLinks/visitDescriptionLinks/worker.ts
+++ b/src/server/worker/tasks/verifyLinks/visitDescriptionLinks/worker.ts
@@ -1,8 +1,7 @@
-import { Sockets } from 'meta/socket/sockets'
 import { Objects } from 'utils/objects'
 
 import { DescriptionValidationRedisRepository } from 'server/cache/repository/validation/description'
-import { SocketServer } from 'server/service/socket'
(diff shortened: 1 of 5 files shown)


src/server/worker/tasks/verifyLinks/visitDescriptionLinks/worker.ts
```diff
@@ -54,20 +55,29 @@ export default async (job: VerifyDescriptionLinksJob): Promise<void> => {
...
+      countryIso,
+      cycle,
+      sectionNames,
+    })
+    const validations: RecordDescriptionValidations = {}
+    sectionNames.forEach((sectionName) => {
+      const current = currentValidations[sectionName] ?? {}
+      const update = descriptionValidations[sectionName] ?? {}
+      validations[sectionName] = DescriptionValidations.mergeValidations({ current, update })
+    })
```
yaguzmang, line 70: After removing the `get` from the respository `set` method, this is duplicated, but this duplication is removed in following PRs.

minotogna (approved):

PR #6134, opened 2026-07-09
5758 - Navigation: Data errors - Test - Fix tests passing/failing randomly

yaguzmang:
Some tests were passing or failing randomly. Fixing that here.

Why: the tests run in parallel, and many files were using the same country (e.g. X01). So two tests could be editing the same data at the same time and mess with each other.

For example, the 1d reload test `08-section-tables-1d-forest-area-change.spec.ts` fills `forestArea` 2025 and expects a `forestAreaNetChange` error. But for that error to show up, `forestArea` 2020 and `forest_expansion/deforestation` need values too, and those were being filled by the 1a test in `08-section-tables-1a-extent-of-forest.spec.ts`. So if 1d ran before 1a, the data wasn't there yet and the test failed.

The tooltip tests had a similar problem: the error was there, but another test editing the same country could wipe it. So, the cell was red when the test checked it, and by the time it hovered, the error was gone.

Changes:
- Each test file gets its own country now (X05–X07, X11–X14), so parallel runs don't step on each other
- `clearTable` waits for the request to finish.
- The 1d reload test seeds its own data, so it doesn't depend on another test.
- The tooltip helper moves the mouse away before hovering.
- `unlockEditing` does nothing if the page is already unlocked (this was throwing an error)

image


minotogna (commented):
I leave the review to @sorja

src/test/e2e/utils/dom/index.ts
```diff
@@ -35,7 +35,11 @@ const elementNotExists = async (locator: Locator): Promise<void> => {
 }
 
 const unlockEditing = async (page: Page): Promise<void> => {
-  await page.locator('.btn-lock.locked').click()
+  // Unlock editing if not already unlocked
+  await page.locator('.btn-lock').waitFor({ timeout: 5000 })
+
+  const lockedButton = page.locator('.btn-lock.locked')
+  if (await lockedButton.isVisible()) await lockedButton.click()
 }
```
sorja: @yaguzmang Can it ever be the case where the page is already unlocked unexpectedly? Would it be "coding error" then?

We can leave this, just curious when would this actually happen

src/test/e2e/utils/tooltip/index.ts
```diff
@@ -7,6 +7,8 @@ const expectValidationTooltip = async (page: Page, locator: Locator, text: strin
 
   // Retry hover before failing - tooltip is flaky
   await expect(async () => {
+    // Move the pointer away in case it is already hovering the location
+    await page.mouse.move(0, 0)
```
sorja, line 11: Does this happen?

src/test/e2e/utils/dom/index.ts
yaguzmang: If a page that is freshly loaded unlocked without any action, yes, that would be a coding error, but I don't see that happening now 🤔 

I added this check because we have a case where we unlock, then try to unlock again:
1. `CountryStatusUtils.expectSubmitToReviewWarning(page)` calls `DOMUtils.unlockEditing(page)`.
2. It opens the status modal and cancels it, leaving editing unlocked.
3. The 1d reload test then calls `DOMUtils.unlockEditing(page)` again before clearing the table.
4. Previously, the second call searched for .btn-lock.locked, which no longer existed, so Playwright timed out.

We could instead add a `lockEditing` function that we could call after step 2, or we can leave this `unlockEditing` as is. Let me know what you think ✌️

src/test/e2e/utils/tooltip/index.ts
yaguzmang: Yes, TableDomUtils.fillCell() does this:
```ts
await cellInput.click()
await cellInput.fill(value)
await cellInput.blur()
```
`blur()` removes focus but doesn't move the mouse. So, the pointer remains over the input/cell. The validation error and its tooltip can then be added asynchronously after the save and rerender, while the pointer is already inside the tooltip target

src/test/e2e/utils/dom/index.ts
minotogna: @sorja why the lock remains unlocked ?
Utilities should have a predictable behavior, that doesn't sound right

sorja: Thinking about this and I think @yaguzmang implementation is correct but maybe the function name could be better.

The idea of the FN should be "if the platform lock is locked then unlock it - if it's already unlocked - do nothing."

Thinking from the users perspective, some actions require the platform to be unlocked:

- The user would open the lock if the lock would be closed
- The user would not do anything if the lock was already open

Right?

yaguzmang: Yes, that's the idea. 
We could name it `ensureEditingUnlocked()`, that would be more accurate

sorja: Sounds good. This method could call internal method _unlockEditingLock or however it is named in the platform

👍

yaguzmang: Added 👍

sorja: Looks good to me.

@minotogna For your final approval

minotogna (approved):

PR #6135, opened 2026-07-10
5758 - Navigation: Data errors - NDP -  Exclude 1a/1b hidden descriptions/data sources from summary

yaguzmang:
Handling the last two cases where descriptions/data sources are hidden and shouldn't be considered in the summary:

1. As mentioned in the last PR, in 1b if 'use NDP' is enabled, we should not consider the 1b data sources errors in the summary:

https://github.com/user-attachments/assets/41254797-093d-4c75-9338-59e2d9a582cd

2. The last case we almost overlooked: The country uses NDPs but there aren't any yet, so in 1a the data sources/descriptions are shown. 
The same logic applies here: once an NDP is added and those descriptions/data sources can't be seen, then they shouldn't be considered in the summary. 

https://github.com/user-attachments/assets/7aab667c-df1f-4077-b837-dce27ead4a03

Coming next: validateLinks tool


src/meta/assessment/description/descriptions.ts
```diff
@@ -30,6 +31,34 @@ const getLabelKey = (props: GetLabelKeyProps): string => {
...
+}
+
+// Descriptions that remain visible when a section shows only comments
+const commentDescriptionNames = [
+  CommentableDescriptionName.generalComments,
+  CommentableDescriptionName.introductoryText,
+]
+
+// A section shows only comments when its data comes from national data points
+const isOnlyComments = (props: OnlyCommentsProps): boolean => {
```
minotogna: This should be called hasSectionOnlyComments 
And let's move this utility under meta/assessment/descriptions (with index , and sub files for each method )

yaguzmang: Done 👍

minotogna (approved):

PR #6136, opened 2026-07-10
5758 - Fix error visibility in NDP

yaguzmang:
Before
Fed and Agiculture Organisaten

After
image


src/client/pages/OriginalDataPoint/components/Comments/hooks/useValidationErrors.ts
```diff
@@ -18,8 +17,7 @@ type Returned = Array<string>
 export const useValidationErrors = (props: Props): Returned => {
   const { field } = props
   const { t } = useTranslation()
-  const canEdit = useIsEditODPDescriptionEnabled()
-  const showValidationErrors = useShowNDPValidationErrors({ canEdit })
+  const showValidationErrors = useShowNDPValidationErrors()
```
yaguzmang, line 20: `useIsEditODPDescriptionEnabled()` and `useIsEditODPEnabled()` return false if the data is locked, so we can't really use them as they are here. I didn't realize before because I had the data unlocked 😅 

I could rework those hooks and add something like `useCanEditODP()` that doesn't care if the data is locked, instead of using `const canEdit = useCanEditCycleData()` 
Let me know ✌️

minotogna: That's fine @yaguzmang

minotogna (approved):

PR #6137, opened 2026-07-11
5758 - Navigation: Data errors - Visit links tool - Extract common emitLinksVerificationEvent & insertLinksCheckActivityLog

yaguzmang:
First step to implement the visit links tool:

Extracting common `emitLinksVerificationEvent` and `insertLinksCheckActivityLog` so we deduplicate code and can reuse them later in the tool.

Coming next:
- Extract the verification logic out of the worker into a `verifyCycleLinks` function (worker becomes a thin wrapper), so it can be called without a queue job.


src/meta/socket/sockets/links.ts
```diff
@@ -2,6 +2,13 @@ import { CountryIso } from 'meta/area/countryIso'
 import { AssessmentName } from 'meta/assessment/assessment'
 import { CycleName } from 'meta/assessment/cycle'
 
+export enum LinksVerificationEvent {
```
minotogna: meta/socket/event/links.ts

yaguzmang: Moved it to `meta/socket/event/links.ts` 👍

minotogna (approved):

PR #6138, opened 2026-07-11
5758 - Navigation: Data errors - Visit links tool - Extract verifyAllLinks

yaguzmang:
Moving the code to verify links from `worker` to a callable `verifyAllLinks` function. No logic changes. Simply moving the code.  
`worker.ts` now just reads the data from the BullMQ job, builds the log key, and calls `verifyAllLinks`.

Also, renamed `visitCycleLinks` to `scheduleVerifyAllLinks`. That name better describes what it does, and it's not confusing with `verifyAllLinks`.

### DIFF
--- a/src/server/worker/tasks/verifyLinks/visitCycleLinks/worker.ts
+++ b/src/server/worker/tasks/verifyLinks/visitCycleLinks/worker.ts
@@ -1,26 +1,7 @@
-import { CountryIso } from 'meta/area/countryIso'
-import { Assessment, AssessmentNames } from 'meta/assessment/assessment'
-import { Cycle } from 'meta/assessment/cycle'
-import { OriginalDataPoint } from 'meta/assessment/originalDataPoint'
-import { LinkToVisit } from 'meta/cycleData/links/link'
 import { Objects } from 'utils/objects'
 
-import { AreaRedisRepository } from 'server/cache/repository/area'
-import { SectionRedisRepository } from 'server/cache/repository/section'
-import { DB } from 'server/db/db'
-import { DescriptionRepository } from 'server/db/repository/assessmentCycle/descriptions'
-import { LinkRepository } from 'server/db/repository/assessmentCycle/links'
-import { OriginalDataPointRepository } from 'server/db/repository/assessmentCycle/originalDataPoint'
-import { Logger } from 'server/utils/logger'
-import { insertLinksCheckActivityLog } from 'server/worker/tasks/verifyLinks/utils/insertLinksCheckActivityLog'
-import { refreshDescriptionValidations } from 'server/worker/tasks/verifyLinks/visitDescriptionLinks/utils/refreshDescriptionValidations'
-import { refreshNationalDataPointValidations } from 'server/worker/tasks/verifyLinks/visitNationalDataPointLinks/utils/refreshNationalDataPointValidations'
-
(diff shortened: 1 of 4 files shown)


src/server/worker/tasks/verifyLinks/visitCycleLinks/worker.ts
```diff
@@ -33,125 +14,9 @@ const _getLogKey = (job: VerifyAllLinksJob): string => {
...
 
-    const duration = (new Date().getTime() - time) / 1000
-    Logger.info(`${logKey} ended in ${duration} seconds with ${linkVisits.length} links visited.`)
-    return Promise.resolve()
-  } catch (error) {
-    Logger.error(`${logKey} Error.`)
-    Logger.error(error)
-    return Promise.reject(error)
-  }
+  return verifyAllLinks({ assessment, countryIso, cycle, logKey, user })
```
yaguzmang, line 21: For the next PR:

The plan is to add a `VerifyLinksJob.runInProcess` method that runs the verification without BullMQ, but still updates the status, sends the socket events, and writes the activity logs as usual.

```ts
VerifyLinksJob.runInProcess()
  -> set status: running
  -> emit event: active
  -> verifyAllLinks()
  -> set status: success | failed
  -> emit event: completed | failed
  -> insert activity log: completed | failed
```

Then the links validation tool can use it like this:

```ts
for each assessment/cycle:
   await new VerifyLinksJob({ assessment, cycle }).runInProcess(user, client)
```

Why add `runInProcess` and not just call `verifyAllLinks()` from the tool: the guard that asks users to run the links check before moving to review reads the last verification time from the job status and the `linksCheckComplete` activity log. Those are written by the job lifecycle.

---

please let me know any ideas/questions/suggestions :v:

minotogna: @yaguzmang I'm not sure I follow the logic why this is needed . Can you please detail the explanation please (before implementing it ?)
And please , tag me

yaguzmang: Sure @minotogna 
The reason is that to display this warning for example:
image

we need to know when the links were last verified. We get that by checking the job status and the `linksCheckComplete` activity log.

So when the tool verifies links, it should:
- insert the activity logs (started/completed/failed). This is what updates the warning.
- and ideally also set the job status (running/success/failed), so the UI can show when a verification is in progress and when it last finished.

It's not strictly necessary to do it in `VerifyLinksJob.runInProcess()`. We could also add the activity logs from the tool and that would work fine too. But `VerifyLinksJob.runInProcess()` seems more organized to me, that way there is only one place with the verification status logging logic.

minotogna: hi @yaguzmang thank you, now I understand.
I would then add service. `LinksService.verifyLinks` .
Would that work ?

yaguzmang: Yes, that would work :+1: , thank you. I will implement `LinksService.verifyLinks` next PR. 
This current PR is good to go as it is since it just extracts a callable `verifyAllLinks`

minotogna: Ok @yaguzmang 👍

minotogna (approved):

PR #6139, opened 2026-07-13
6097 - Dependencies upgrade - React router v8 + redlock-universal

### LINKED ISSUE #6097: Dependencies upgrade: react-router

yaguzmang:


tsconfig.json
```diff
@@ -4,6 +4,8 @@
   "compilerOptions": {
     "allowJs": true,
     "isolatedModules": true,
-    "jsx": "react"
+    "jsx": "react",
+    "module": "Preserve",
+    "moduleResolution": "Bundler"
```
yaguzmang, line 15: React Router 8 is ESM-only 

> To modernize the library, React Router is now published as an ESM-only module, and tsconfig target/lib fields have been updated to ES2022 across the board.

https://remix.run/blog/react-router-v8

This means we need to update the ts config so it can resolve React Router’s new package exports:

- "moduleResolution": "Bundler" -> find packages the way Vite does 
- "module": "Preserve" -> don't transform import/export, Vite handles that.

tsconfig.server.json
```diff
@@ -2,6 +2,8 @@
   "extends": "./tsconfig.base.json",
   "exclude": ["node_modules", "src/client", "src/test", "src/tools", "src/**/*.test.ts"],
   "compilerOptions": {
+    "module": "Node20",
+    "moduleResolution": "Node16",
```
yaguzmang, line 6: Similarly here for node:
- `"module": "Node20"` ->supports loading ESM packages.
- `"moduleResolution": "Node16"` -> resolve package `"exports"` using Node’s `import`/`require` rules.

package.json
yaguzmang: Now with the ts config updates, this package is breaking the build:

Image

I was looking into it and it's a known bug in the package: https://github.com/mike-marcacci/node-redlock/issues/307

And, according to the package creator, the package is not maintained anymore: https://github.com/mike-marcacci/node-redlock/issues/310#issuecomment-4078742458
Last published date in `npm` is 4 years ago. 

There are ways around that issue, but maybe it's better to use a different package. Should remove this package and use a different one before upgrading react router @minotogna ?

tsconfig.server.json
minotogna: do we need to update the tsconfig.server as well ? isn't react router affecting only client side ?

package.json
minotogna: hi @yaguzmang we could use `redlock-universal` ? feel free to replace it

tsconfig.server.json
yaguzmang: It also affects the backend because In `src/meta/routes/routes/createRoute.ts` we import `generatePath` from react-router:

image


And we use that backendside, for example:
```ts
res.redirect(
  Routes.Login.generatePath(
    { assessmentName, cycleName },
    { loginError: msg.message }
  )
)
```

I'm checking and this is the only import from react-router that affects the backend. This what it does:
image


Should I look into removing the import? or should we leave it and update the ts config

package.json
yaguzmang: Yes I think so, I will check and remove this old `redlock` 👍

tsconfig.server.json
minotogna: I think it's fine to update the ts config server if there are no other side effects @yaguzmang

tsconfig.json
```diff
@@ -1,9 +1,17 @@
 {
   "extends": "./tsconfig.base.json",
   "allowSyntheticDefaults": true,
+  "ts-node": {
+    "compilerOptions": {
+      "module": "Node20",
+      "moduleResolution": "Node16"
+    }
+  },
```
yaguzmang, line 9: This is so the tools work as usual with ts-node

tsconfig.server.json
yaguzmang: I checked and I saw no other side effects, so I will keep the new ts config 👍

src/server/start.ts
```diff
@@ -1,16 +1,26 @@
...
+
 serverInit()
 
 if (ProcessEnv.nodeEnv === NodeEnv.development) {
-  void import('./worker/tasks/verifyLinks/triggerVerifyLinksWorker')
-    .then(({ triggerVerifyLinksWorker }) => triggerVerifyLinksWorker())
-    .catch((error) => {
-      Logger.error(`[verifyLinks] failed to start local worker: ${JSON.stringify(error)}`)
-    })
+  const { triggerVerifyLinksWorker } = loadModule(
```
minotogna, line 19: @yaguzmang why this change ?

yaguzmang: With the new `tsconfig.server.json` change (`module: Node20 / moduleResolution: Node16`),  TypeScript treats this import as ESM: `import('./worker/tasks/verifyLinks/triggerVerifyLinksWorker')` and Node then requires a file extension (it expects ` import('.../triggerVerifyLinksWorker.js')`). 
But, if I add the `.js` extension, then the import fails in local because with `ts-node-dev` we use the `.ts` files directly, not a compiled `.js`.
image

With `createRequire` and `loadModule`, we don't need the file extension and that fixes the issue.

minotogna: Ok thank you 😊

minotogna (commented):
@sorja for your final approval . 
This is a substantial update

sorja (approved):

PR #6144, opened 2026-07-16
5758 - Navigation: Data errors - Move reducer outside tableData

yaguzmang:
Moving the validations reducer outside the `tableData` reducer. 
Now that the validations have NDPS, descriptions, and so on, it makes more sense that validations are not inside the `tableData` reducer. 

image


src/client/store/data/validations/slice/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import { setDescriptionValidationsReducer } from 'client/store/data/validations/slice/extraReducers/setDescriptionValidationsReducer'
+import { setNationalDataPointValidationsReducer } from 'client/store/data/validations/slice/extraReducers/setNationalDataPointValidationsReducer'
+import { setValidationsReducer } from 'client/store/data/validations/slice/extraReducers/setValidationsReducer'
+import { updateNationalDataPointValidationsReducer } from 'client/store/data/validations/slice/extraReducers/updateNationalDataPointValidationsReducer'
+import { updateSummaryReducer } from 'client/store/data/validations/slice/extraReducers/updateSummaryReducer'
+import { initialState } from 'client/store/data/validations/state'
+
+import { ValidationsReducerName } from './name'
+
+export const ValidationsReducer = createSlice({
```
minotogna: better to have a combine reducer

yaguzmang: In this PR or better next PR for easier review? The split to `tableValidationReducer`, `nationalDataPointValidationReducer`, `descriptionValidationReducer`, `summaryValidationReducer` will edit a lot of files

yaguzmang: As mentioned is slack: For the validations reducers split I think better multiple PRs. It's a lot of changes. The current PR just moves the validations out of `tableData`, so we can merge as is. It doesn't add any changes, just moves the reducer out. Let me know what you think ✌️

minotogna: I think if we combine reducer now , it's the same amount of changes . This pr is ok

minotogna: Or up to you . What do you prefer ? @yaguzmang

yaguzmang: @minotogna I did it all in this PR. The diff is big but it's mostly moving code around. This is the new shape of the validations store:
```txt
src/client/store/data/validations
├── actions/                                  
│   ├── index.ts                              # ValidationActions = { removeValidations }
│   └── removeValidations.ts        # shared action to remove all validations. e.g when unmounting
│
├── descriptions/
│   ├── actions/
│   │   ├── index.ts                          # DescriptionValidationActions
│   │   ├── getDescriptionValidations.ts      
│   │   └── setDescriptionValidations.ts
│   ├── hooks/descriptions.ts
│   ├── selectors/index.ts
│   ├── slice/
│   │   ├── index.ts                          # DescriptionValidationSlice
│   │   ├── name.ts
│   │   └── extraReducers/
│   │       ├── removeValidationsReducer.ts
│   │       └── setDescriptionValidationsReducer.ts
│   └── state.ts
│
├── nationalDataPoints/
│   ├── actions/
│   │   ├── index.ts                          # NationalDataPointValidationActions
│   │   ├── deleteNationalDataPointValidation.ts
│   │   ├── getNationalDataPointValidations.ts
│   │   ├── setNationalDataPointValidations.ts
│   │   └── updateNationalDataPointValidations.ts
│   ├── hooks/nationalDataPoints.ts
│   ├── selectors/index.ts
│   ├── slice/
│   │   ├── index.ts                          # NationalDataPointValidationSlice
│   │   ├── name.ts
│   │   └── extraReducers/
│   │       ├── deleteNationalDataPointValidationReducer.ts
│   │       ├── removeValidationsReducer.ts
│   │       ├── setNationalDataPointValidationsReducer.ts
│   │       └── updateNationalDataPointValidationsReducer.ts
│   └── state.ts
│
├── summary/
│   ├── actions/
│   │   ├── index.ts                          # SummaryValidationActions
│   │   ├── getSummary.ts
│   │   ├── setValidationSummary.ts
│   │   └── updateValidationSummary.ts       
│   ├── hooks/summary.ts
│   ├── selectors/index.ts
│   ├── slice/
│   │   ├── index.ts                          # SummaryValidationSlice
│   │   ├── name.ts
│   │   └── extraReducers/
│   │       ├── getSummaryReducer.ts
│   │       ├── removeValidationsReducer.ts
│   │       └── setValidationSummaryReducer.ts
│   └── state.ts
│
└── tables/
    ├── actions/
    │   ├── index.ts                          # TableValidationActions
    │   ├── getTableValidations.ts           
    │   └── setNodeValueValidations.ts
    ├── hooks/tables.ts
    ├── selectors/index.ts
    ├── slice/
    │   ├── index.ts                          # TableValidationSlice
    │   ├── name.ts
    │   └── extraReducers/
    │       ├── removeValidationsReducer.ts
    │       └── setNodeValueValidationsReducer.ts
    └── state.ts
```

The one real change is that to update the summary before we used 
```ts
  builder.addMatcher(
    isAnyOf(
      setDescriptionValidations,
      setNodeValueValidations,
      setNationalDataPointValidations,
      updateNationalDataPointValidations,
      deleteNationalDataPointValidation
    ),
```
however now that we use `combineReducers` and each slice is divided, we can't use `builder.addMatcher(sAnyOf(...` because summary can't access the state of table validations, for example. 
Now instead we call `updateValidationSummary` after every validation data update. e.g
```ts
      dispatch(NationalDataPointValidationActions.updateNationalDataPointValidations({...})
      dispatch(SummaryValidationActions.updateValidationSummary({...}))
```

minotogna: Wait @yaguzmang 
Why do we call this new update summary action ?

yaguzmang: Because before, the summary was in the same slice as the table/description/NDP validations, so the reducer had access to that data. For example we could do:

  ```ts
  const descriptionValidations =
  state.descriptions?.[assessmentName]?.[cycleName]?.[countryIso]
  const nationalDataPointValidations =
  state.nationalDataPoints?.[assessmentName]?.[cycleName]?.[countryIso]
  ```

  Now with the combined reducers we can't do `state.descriptions` or `state.nationalDataPoints` inside the summary reducer, each reducer only has access to its own state. So instead we call the action after updating the validations, passing what changed:

  ```ts
  dispatch(SummaryValidationActions.updateValidationSummary({
    assessmentName, countryIso, cycleName,
    updateNationalDataPoints: true,
  }))
  ```
The action reads the other validation states with `getState()` and recomputes the summary. It's the same logic the matcher was doing before, just called explicitly after each validation data update. 
  
We could also use a `listenerMiddleware` to keep it more similar to how it was before, with `matcher: isAnyOf(setNodeValueValidations, setDescriptionValidations, ...)` dispatching `updateValidationSummary` in the effect. Let me know if you would prefer that ✌️

minotogna: @yaguzmang 

two consecutive actions should be avoided:
```typescript       
    dispatch(NationalDataPointValidationActions.updateNationalDataPointValidations({...})
    dispatch(SummaryValidationActions.updateValidationSummary({...}))
```
I have seen calling the SummaryValidationActions.updateValidationSummary everywhere in the code.. we should not do that, not very maintainable .

If you need to update the summary when a validation gets updated (e.g. description, table, ndp), then a listener or an even  effect is a better approach. 
You can recalculate the summary when any of those validation state changes, calculate the summary and then dispatch an action to update the store.

Is this what you are trying to achieve?

src/client/store/data/validations/nationalDataPoints/actions/index.ts
```diff
@@ -0,0 +1,11 @@
+import { deleteNationalDataPointValidation } from 'client/store/data/validations/nationalDataPoints/actions/deleteNationalDataPointValidation'
+import { getNationalDataPointValidations } from 'client/store/data/validations/nationalDataPoints/actions/getNationalDataPointValidations'
+import { setNationalDataPointValidations } from 'client/store/data/validations/nationalDataPoints/actions/setNationalDataPointValidations'
+import { updateNationalDataPointValidations } from 'client/store/data/validations/nationalDataPoints/actions/updateNationalDataPointValidations'
+
+export const NationalDataPointValidationActions = {
+  deleteNationalDataPointValidation,
+  getNationalDataPointValidations,
+  setNationalDataPointValidations,
+  updateNationalDataPointValidations,
```
minotogna: the word nationalDataPoint can be removed.
The word NationalDataPoint is already in the object.
```NationalDataPointValidationActions.deleteValidation``` is more readable

File: src/client/store/data/validations/descriptions/actions/index.ts [hunk 3]
```diff
@@ -0,0 +1,7 @@
+import { getDescriptionValidations } from 'client/store/data/validations/descriptions/actions/getDescriptionValidations'
+import { setDescriptionValidations } from 'client/store/data/validations/descriptions/actions/setDescriptionValidations'
+
+export const DescriptionValidationActions = {
+  getDescriptionValidations,
+  setDescriptionValidations,
```
minotogna: the word description  can be removed.
The word Description is already in the object.
```DescriptionActions.setValidations``` is more readable

File: src/client/store/data/validations/summary/actions/index.ts [hunk 4]
```diff
@@ -0,0 +1,9 @@
+import { getSummary } from 'client/store/data/validations/summary/actions/getSummary'
+import { setValidationSummary } from 'client/store/data/validations/summary/actions/setValidationSummary'
+import { updateValidationSummary } from 'client/store/data/validations/summary/actions/updateValidationSummary'
+
+export const SummaryValidationActions = {
+  getSummary,
+  setValidationSummary,
+  updateValidationSummary,
```
minotogna: the word summary can be removed ...

src/client/store/data/validations/nationalDataPoints/slice/extraReducers/removeValidationsReducer.ts
```diff
@@ -0,0 +1,14 @@
+import { ActionReducerMapBuilder } from '@reduxjs/toolkit'
+
+import { Objects } from 'utils/objects'
+
+import { removeValidations } from 'client/store/data/validations/actions/removeValidations'
+import { NationalDataPointValidationState } from 'client/store/data/validations/nationalDataPoints/state'
+
+export const removeValidationsReducer = (builder: ActionReducerMapBuilder<NationalDataPointValidationState>): void => {
+  builder.addCase(removeValidations, (state, action) => {
```
minotogna: here we should import the full object since the domain is different. (also in the other slices )

src/client/store/data/validations/tables/actions/index.ts
```diff
@@ -0,0 +1,7 @@
+import { getTableValidations } from 'client/store/data/validations/tables/actions/getTableValidations'
+import { setNodeValueValidations } from 'client/store/data/validations/tables/actions/setNodeValueValidations'
+
+export const TableValidationActions = {
+  getTableValidations,
```
minotogna: the word table can be removed

src/client/store/data/validations/slice/index.ts
yaguzmang: Yes that's what I want to achieve. I now added [src/client/store/data/validations/summary/listener.ts](https://github.com/openforis/fra-platform/pull/6144/changes#diff-12147c6536f2d48d4c08529a4ab47da61b97352082fe5c1795e1cba3a094aab5) there we listen for those update events in one place 👍

src/client/store/data/validations/tables/actions/index.ts
yaguzmang: I renamed all of those and also the selectors, thanks 👍

src/client/store/data/validations/summary/recomputeSummary.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type RecomputeSummaryProps = {
+  summary: ValidationSummary
+  descriptionSectionNames?: Array<SectionName>
+  descriptionValidations?: RecordDescriptionValidations
+  nationalDataPointValidations?: RecordNDPValidations
+  tableValidations?: RecordTableValidationsState
+}
+
+export const recomputeSummary = (props: RecomputeSummaryProps): ValidationSummary => {
```
minotogna: better to have a utility `ValidationSummaries.compute`
under `src/meta/assessment/validation`.
Also this file is too big. in the utility better to split in sub files by responsibility.

src/client/store/data/validations/summary/listener.ts
```diff
@@ -0,0 +1,57 @@
...
+const isValidationUpdate = isAnyOf(
+  TableValidationActions.setNodeValueValidations,
+  DescriptionValidationActions.setValidations,
+  NationalDataPointValidationActions.setValidations,
+  NationalDataPointValidationActions.updateValidations,
+  NationalDataPointValidationActions.deleteValidation
+)
+
+export const startValidationSummaryListener = (): void => {
+  startAppListening({
```
minotogna: this should not be called in the store.
It should be in a useEffect and called where the summary is fetched.

The effect should also contain the return with the unsubscribe .

src/client/store/data/validations/summary/recomputeSummary.ts
yaguzmang: Added 👍 and now reusing `ValidationSummaries.compute` for the backend as well

src/client/pages/Country/hooks/useValidationSummaryListener.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const isValidationUpdate = isAnyOf(
+  TableValidationActions.setNodeValueValidations,
+  DescriptionValidationActions.setValidations,
+  NationalDataPointValidationActions.setValidations,
+  NationalDataPointValidationActions.updateValidations,
+  NationalDataPointValidationActions.deleteValidation
+)
+
+export const useValidationSummaryListener = (): void => {
```
minotogna, line 24: This file
Should be moved
Where it's used (with relative import )

yaguzmang: Done

minotogna (approved):

PR #6146, opened 2026-07-20
 5758 - Navigation: Data errors - Add DataValidationService & trigger validation only for affected description or data source

yaguzmang:
Addressing these points from the main PR todo list:
 - "DescriptionValidation: Only call the related method to validate correct description. E.g. when editing description, dont call datasourcevalidation (because we are not editing data sources)"

 - "Move to DataValidationService.validateDescriptions. Remove redundant updateDescriptionLinkValidations.ts line 21 (inline the code). Call updateDataSourceFieldValidations only if needed, which is renamed to validateDataSources."


src/server/service/dataValidation/validateDescriptions.ts
```diff
@@ -14,11 +20,21 @@ type Props = {
...
-export const updateDescriptionLinkValidations = async (props: Props): Promise<void> => {
+export const validateDescriptions = async (props: Props): Promise<void> => {
   const { assessment, country, cycle, descriptions, notifyClients } = props
   const { countryIso } = country
 
-  if (Objects.isEmpty(descriptions)) return
+  // Validate required data source fields first (not in parallel) to avoid cache race conditions.
+  const dataSourceDescriptions = descriptions.filter(({ name }) => name === CommentableDescriptionName.dataSources)
+  if (!Objects.isEmpty(dataSourceDescriptions)) {
+    await validateDataSources({
```
minotogna: validateDataSource should not be called in 2020. is it the case ?

yaguzmang: Yes that was the case, but I now added a check to skip them 👍

src/server/service/dataValidation/canValidateDataSources.ts
```diff
@@ -0,0 +1,20 @@
...
+import { CycleNames } from 'meta/assessment/cycle/names'
+import { CommentableDescription, CommentableDescriptionName } from 'meta/assessment/descriptionValue'
+
+type Props = {
+  assessment: Assessment
+  cycle: Cycle
+  description: Omit<CommentableDescription, 'id'>
+}
+
+export const canValidateDataSources = (props: Props): boolean => {
```
minotogna: we use metadata to know if the description uses data source tables.
`Boolean(nationalData.dataSources?.table)` 
defined in 
`src/meta/assessment/description/description.ts`

can we use this instead of hardcoding cycle name ?

src/tools/validations/validateDescriptions.ts
```diff
@@ -35,20 +35,26 @@ const _validateCountry = async (props: CountryProps): Promise<void> => {
...
+  >((acc, [sectionName, sectionValues]) => {
+    Object.entries(sectionValues).forEach(([name, value]) => {
+      const description: Omit<CommentableDescription, 'id'> = {
+        countryIso,
+        name: name as CommentableDescriptionName,
+        sectionName,
+        value,
+      }
+
+      if (DataValidationService.canValidateDataSources({ assessment, cycle, description })) acc.push(description)
```
minotogna: why isn't this centralized? we should call only one validate descriptions that checks what to validate.

src/server/service/dataValidation/canValidateDataSources.ts
sorja: Proposal:

something like this:

`cycle.section.dataSource.text = boolean`

yaguzmang: Image

src/tools/validations/validateDescriptions.ts
yaguzmang: It was redundant, good catch 👍 Now the only place we check this is in `validateDataSources`:
```ts 
if (Objects.isNil(subSection?.props.descriptions?.[cycle.uuid]?.nationalData?.dataSources?.table)) {
  return acc
}
```

minotogna (approved):

PR #6167, opened 2026-07-24
5758 - Navigation: Data errors - Add LinksService.enqueueAllLinksValidation

yaguzmang:
We have `LinksService`:
- `enqueueDescriptionLinksValidation`
- `enqueueNationalDataPointLinksValidation`

Those enqueue jobs and make sure there is a worker to process it.

So, for consistency, I am adding `enqueueAllLinksValidation` which does the same.

### DIFF
--- a/src/server/service/links/index.ts
+++ b/src/server/service/links/index.ts
@@ -1,8 +1,10 @@
 import { enqueueDescriptionLinksValidation } from 'server/service/links/enqueueDescriptionLinksValidation'
 import { enqueueNationalDataPointLinksValidation } from 'server/service/links/enqueueNationalDataPointLinksValidation'
 import { verifyLinks } from 'server/service/links/verifyLinks'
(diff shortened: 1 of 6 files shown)


src/server/service/links/index.ts
```diff
@@ -1,8 +1,10 @@
 import { enqueueDescriptionLinksValidation } from 'server/service/links/enqueueDescriptionLinksValidation'
 import { enqueueNationalDataPointLinksValidation } from 'server/service/links/enqueueNationalDataPointLinksValidation'
 import { verifyLinks } from 'server/service/links/verifyLinks'
+import { scheduleVerifyAllLinks } from 'server/worker/tasks/verifyLinks/visitCycleLinks/scheduleVerifyAllLinks'
```
minotogna: Better to rename this as well .
The exported method should have the same name

minotogna (approved):

PR #6170, opened 2026-07-27
5758 - Navigation: Data errors - DataValidationService.validateNodes

yaguzmang:
Addressing:

- context factory for validations:
  - why do we run two iterations in `#initQueue`
- `src/server/controller/cycleData/validations/tables/updateTableValidations.ts`
  - Move to `DataValidationService.validateNodes()`
- `_removeValidation` → `TableValidations.remove(props)`


src/server/service/dataValidation/validateNodes.ts
```diff
@@ -6,11 +6,10 @@ import { NodeUpdates } from 'meta/data/nodeUpdates'
 import { Sockets } from 'meta/socket/sockets'
 
 import { TableValidationRedisRepository } from 'server/cache/repository/validation/table'
+import { ContextFactory } from 'server/controller/cycleData/validations/tables/context/contextFactory'
+import { validateNodeUpdates } from 'server/controller/cycleData/validations/tables/validateNodeUpdates'
```
yaguzmang: Now that we have `DataValidationService.validateNodes`, should we move:

`src/server/controller/cycleData/validations/tables/`

- `context/`
  - `baseContextBuilder.ts`
  - `context.ts`
  - `contextBuilderProps.ts`
  - `contextFactory.ts`
  - `dataContextBuilder.ts`
- `shouldSkipValidationFormula.ts`
- `validateNodeUpdates.ts`

 to:

`src/server/service/dataValidation/tables/`
?

src/server/service/dataValidation/tables/context/contextFactory.ts
```diff
@@ -88,6 +88,8 @@ export class ContextFactory extends BaseContextBuilder {
       })
     })
 
+    // Add the dependants of each queued variable. The queue grows while we iterate it,
+    // so dependants of dependants are included as well.
     await Promises.each(this.#queue, async (variable) => {
       const variableKey = this.#getVariableKey(variable)
```
yaguzmang: "why do we run two iterations in `#initQueue`?":

The first loop adds the changed nodes to the queue.

The second loop takes each item in the queue and adds its dependants to the end of the queue. Those dependants get their turn in the same loop, so their own dependants are added too.

For example, if the user edits `extentOfForest.totalLandArea[1990]`:

1. The first loop queues `extentOfForest.totalLandArea[1990]` (the changed node).
2. The second loop processes it and queues its dependant `extentOfForest.otherLand[1990]`.
3. Later in the same loop it reaches `otherLand` and queues its dependant `otherLandWithTreeCover.palms[1990]`.

So one edit ends up re-validating the whole chain:
```txt
extentOfForest.totalLandArea[1990]
  → extentOfForest.otherLand[1990]
    → otherLandWithTreeCover.palms[1990]
```

minotogna: Can we have only one loop ? 
I think in the calculations we have one loop only .
Can you please double
Check @yaguzmang

src/server/service/dataValidation/validateNodes.ts
minotogna: If they are used only in validation service , then yes

src/server/service/dataValidation/tables/context/contextFactory.ts
yaguzmang: @minotogna 
In `addDependantsToQueue` we do only one loop but it is recursive, `addDependantsToQueue` calls ` await this.#addDependantsToQueue(dependant)`: 

```ts
  async #addDependantsToQueue(variable: VariableCache): Promise<boolean> {
    const { colName, tableName, variableName } = variable
    const { assessment, country, cycle } = this.props

    const dependants = AssessmentMetaCaches.getCalculationsDependants({ assessment, cycle, tableName, variableName })
    await Promises.each(dependants, async (variable) => {
     
     // ...
     // ...

      // internal variable
      const dependant = { tableName: variable.tableName, variableName: variable.variableName, colName }
      if (!externalVariable && !this.#isInQueue(dependant) && this.#mustAddToQueue(dependant)) {
        await this.#addToQueue(dependant)
        await this.#addDependantsToQueue(dependant) // <------------- Recursive
      }
    })

    return Boolean(
      dependants.find((dependant) => dependant.variableName === variableName && dependant.tableName === tableName)
    )
  }
```

So really it is like having two loops. We could make the validations recursive but I think two loops is easier to understand

src/server/service/dataValidation/validateNodes.ts
yaguzmang: That's the case, so I moved the files 👍

src/server/service/dataValidation/tables/context/contextFactory.ts
minotogna: @yaguzmang I just checked your branch.
but where do you add the dependents of the dependents to the queue? you stop at first level only?

in your first loop can you call #addDependantsToQueue (like we do in the calculation initQueue)?

yaguzmang: @minotogna The loop also covers dependants of dependants because `Promises.each` goes over the growing queue:
`Promises.each` does this:
`for (let i = 0; i < iterable.length; i += 1) {`

It checks `iterable.length` again on every iteration, so anything pushed to `this.#queue` in the loop still gets processed.

For example, editing `extentOfForest.totalLandArea[1990]`:

1. First loop seeds the queue: `[totalLandArea]`
2. Second loop, `i=0`: processes `totalLandArea` and pushes its dependant, so the queue is now `[totalLandArea, otherLand]`
3. `i=1`: the length check sees the new item, so the loop reaches `otherLand` and pushes its dependant: `[totalLandArea, otherLand, palms]`
4. `i=2`: reaches `palms`, which has no dependants, and finishes

I can add the change later to recursive though 👍

minotogna: right @yaguzmang no recursion, but one loop please ;)

yaguzmang: @minotogna Done, `#initQueue` now has only one awaited loop:

```ts
await Promises.each(this.#queue, async (variable) => {
  this.#tableNames.add(variable.tableName)
  await this.#dataContextBuilder.addVariable(variable)
  this.#rowKeys.add(RowCaches.getKey(variable))

  this.#addDependantsToQueue(variable)
})
```

The queue starts with the changed source nodes and grows while the loop runs, so it processes the dependants and their dependants.

Before the loop there is only data prep: a `map` to convert the node updates into queue items and a dedup. No awaits.

This also let me simplify the rest: `#addToQueue` and `#addDependantsToQueue` are synchronous now, and the `#queuedKeys`/`#visitedKeys` sets are gone. Dedup is a `#isInQueue` check, like in the calculations contextFactory.

minotogna (approved):

PR #6179, opened 2026-07-29
6111 - Cycle name change: FRA Latest 2 -> FRA in Progress

### LINKED ISSUE #6111: Cycle name change: FRA Latest (two countries) → FRA in Progress
@minotogna 

To avoid confusion caused by having the same name, the admin agreed to rename the layer “FRA Latest 2”, which contains only two countries mentioned, to “FRA in progression.”

Image

sorja:
https://github.com/user-attachments/assets/23e144ec-284b-4d29-9f48-d72bd10c3643


src/client/components/PageLayout/Header/CycleSwitcher/hooks/useDisplayName.ts
```diff
@@ -0,0 +1,34 @@
...
+  const isLastCreatedCycle = Assessments.getLastCreatedCycle(assessment).name === cycleName
+  if (isPanEuropean && isLastCreatedCycle) {
+    return ''
+  }
+
+  // TODO: NOTE: This logic is subject to change when we merge latest2 to latest
+  // Hide header cycle name for fra cycle named 'latest', show 'in Progress' for 'latest2'
+  const isFra = assessmentName === AssessmentNames.fra
+  if (isFra) {
+    if (cycleName === CycleNames.latest) return ''
```
minotogna: I can see that latest in both assessments is always ''
Should we simplify the logic ?

src/i18n/resources/en/common.json
```diff
@@ -75,6 +76,7 @@
   "lastEdit": "Last edit",
   "lastPublished": "Last published",
   "latest": "Latest",
+  "latestInProgress": "$t(common.latest) $t(common.inProgress)",
```
minotogna, line 79: Do we use this ?

src/client/components/PageLayout/Header/CycleSwitcher/hooks/useDisplayName.ts
sorja: panEuropean has no 'latest' (2020, 2025)
fra "real" latest is latest2

Not sure what you mean

src/i18n/resources/en/common.json
sorja: yes

src/client/components/PageLayout/Header/CycleSwitcher/hooks/useDisplayName.ts
sorja: The logic is

panEuropean:
2020 - show label
2025 - hide label

fra:
2020 - show label
2025 - show label
latest - hide label
latest2 - show custom label (in Progress)

src/i18n/resources/en/common.json
minotogna: Where ?

src/client/components/PageLayout/Header/CycleSwitcher/hooks/useDisplayName.ts
minotogna: Ok but actually @sorja even fra won't necessarily have latest cycle . 
It's possible we will have 2020, 2025,2029,2030

In that case 2030 will need to be ''

The logic should be generic , not assessment specific

src/i18n/resources/en/common.json
sorja: Image

in the drop down

src/client/components/PageLayout/Header/CycleSwitcher/hooks/useDisplayName.ts
sorja: What do you suggest?

Currently we hide panEuropean last cycle and FRA second last cycle

We cant hide label for lastCreated or lastPublished

if we have 2020, 2025, 2029 and 2030, then we wouldnt probably have latest2 after 2030?

Another possibility is to introduce cycle.props.headerLabel

minotogna: If name=== latest2 return common....

If isLastPublishedCycle or name===latest
   Then return ''

Return name

minotogna: @sorja i corrected the order . It's important

minotogna (approved):

PR #6181, opened 2026-07-29
5758 - Navigation: Data errors - Add DataValidationService.removeValidations methods

yaguzmang:
Addressing:
- [x] `src/tools/validations/nationalDataPoint/validateCountryNationalDataPoints.ts`
  - [x] Why do we get current validations if the object here is to validate completely all NDP?
    here we should remove all existing validations and validate from scratch all NDP

- [x] `src/tools/validations/tables/validateCountryTables.ts`
  line 15: expose the method in `DataValidationService.removeTableValidations` and rename repos method `TableValidationRedisRepository.removeValidations`


src/tools/validations/nationalDataPoint/validateCountryNationalDataPoints.ts
```diff
@@ -15,20 +16,13 @@ export const validateCountryNationalDataPoints = async (props: CountryProps, cli
...
 
   const validations: RecordNDPValidations = {}
   nationalDataPoints.forEach((nationalDataPoint) => {
     const { uuid } = nationalDataPoint
-    // NDP data source references are validated by the links flow.
-    validations[uuid] = NationalDataPointValidator.validate({
-      nationalDataPoint,
-      validation: currentValidations[uuid] ?? {},
-    })
+    validations[uuid] = NationalDataPointValidator.validate({ nationalDataPoint, validation: {} })
```
minotogna: do we need to pass a validation object here ?

yaguzmang: Not anymore. Removed it 👍

minotogna (approved):

PR #6182, opened 2026-07-29
5758 - Navigation: Data errors - NDP validation methods: validate, set, notify 

yaguzmang:
Addressing:
- [ ] NDP Validation methods:
  each validation should validate, set and notify client ~~(not get validation first, but return the validation from the repo)~~
  
 ---

As discussed in the comments, the validations are stored as a single value per NDP, so updating one field requires read-merge-write, and we agreed to keep the storage as is and the repo methods as plain get/set. The merge therefore stays in the service:

  1. `NationalDataPointValidator.validateNationalClasses({ nationalDataPoint })`  validates only its field, no cached input
  2. The service gets the stored validation, merges the result onto it and sets it back
  3. `notifyNationalDataPointValidationUpdate`  notifies clients with the complete stored validation

  `validateNDPYear` follows the same pattern.


src/server/service/dataValidation/nationalDataPoint/validateNDPNationalClasses.ts
```diff
@@ -17,26 +18,23 @@ type Props = {
...
-    countryIso,
-    cycle,
-    validations: { [uuid]: updatedValidation },
-  })
+  const validations = { [uuid]: updated }
+  await NationalDataPointValidationRedisRepository.setValidations({ assessment, countryIso, cycle, validations })
 
-  notifyNationalDataPointValidationUpdate({ assessment, countryIso, cycle, validations: { [uuid]: updatedValidation } })
+  notifyNationalDataPointValidationUpdate({ assessment, countryIso, cycle, validations })
 }
```
yaguzmang, line 40: @minotogna in the review notes it says:
"
- [ ] NDP Validation methods:
  each validation should validate, set and notify client (not get validation first, but return the validation from the repo)

  e.g for national classes:

  1. `const validation = NationalDataPointValidator.validateNationalClasses({..})`
  2. `const validations = NationalDataPointValidationRedisRepository.setNationalClassValidations({...,validation})`
  3. `notifyNationalDataPointValidationUpdate({...validation})`
"

However, to set the validations, we need to get the current validations, merge the new result, and set the merged validations. This is because in redis we store each NDP's validation as a single hash entry, with all the fields together in one JSON value:

```
HSET <countryKey>
  <ndpUuid>  '{ "year": {...}, "nationalClasses": {...}, "dataSources": {...}, "odpId": 123 }'
```
So there is no way to update only `nationalClasses` without first reading the whole value, otherwise we'd wipe the year and link validations stored next to it. We need to read-merge-write somewhere. Previously I had done it in the redis repo method itself but we agreed those methods should only set. 

We could change how we store validations: one hash field per validation field, instead of one JSON with everything.

```
HSET <countryKey>
  <ndpUuid>:year             '{ "valid": false, "messages": [...] }'
  <ndpUuid>:nationalClasses  '{ "<classUuid>": {...} }'
  <ndpUuid>:dataSources      '{ "<dataSourceUuid>": {...} }'
  <ndpUuid>:odpId            '123'
```

That way we don't need to read anything to set validations. We would need to edit how we read the validations, but maybe that's better

minotogna: right @yaguzmang I think we might not change that then
@sorja what do you think ?

sorja: Ok for me

minotogna (approved):

PR #6193, opened 2026-07-31
5758 - Navigation: Data errors - Add missing NDP reference/comment link tests

yaguzmang:


src/test/e2e/tests/06-national-data-point.spec.ts
```diff
@@ -11,7 +12,13 @@ import {
   x11ExtentOfForestPath,
 } from './06-national-data-point.fixture'
 
-// TODO: Validate links
```
yaguzmang, line 14: Adding this missing test. CC @sorja 
Also, I noticed that fra 2025 uses data sources v1, and that still uses client side validation for the references links. I will be updating that next so it uses backend validations @minotogna 
edit: added tests for reference links as well now that the update is merged.

minotogna: Sure 👍 thank you @yaguzmang

minotogna: @sorja for your final approve

sorja (approved):

PR #6199, opened 2026-08-03
6169 - MultiSelect: Align selected Tooltip values style

### LINKED ISSUE #6169: MultiSelect: Align selected Tooltip values style
Image

Use same style always as for countries (list with hr separating items, under component if room)

sorja:
https://github.com/user-attachments/assets/f04c54f5-0705-4a86-ac97-b0aab32396d0

Version 1:
- Only add custom tooltip to Multiselect component similar to what we have for CountryMultiSelect

### DIFF
--- a/src/client/components/Inputs/MultiSelect/MultiSelect.scss
+++ b/src/client/components/Inputs/MultiSelect/MultiSelect.scss
@@ -4,6 +4,19 @@
 
 $icon-size: 14px;
 
+.multiselect__tooltip-trigger {
+  height: 100%;
+}
+
+.multiselect__tooltip-values {
+  display: grid;
+  row-gap: $spacing-xxxs;
+}
+
+.multiselect__tooltip-value {
+  padding: $spacing-xxxs 0;
+}
+
 div.multiselect__container {
   border-radius: 2px;
   border: 1px solid $ui-border;
--- a/src/client/components/Inputs/MultiSelect/MultiSelect.tsx
+++ b/src/client/components/Inputs/MultiSelect/MultiSelect.tsx
@@ -1,9 +1,11 @@
 import './MultiSelect.scss'
-import React from 'react'
-import { PlacesType } from 'react-tooltip'
+import React, { useMemo } from 'react'
+import { PlacesType, Tooltip } from 'react-tooltip'
 import classNames from 'classnames'
 
 import { TooltipId } from 'meta/tooltip/id'
+import { UUIDs } from 'meta/uuid/uuids'
+import { Objects } from 'utils/objects'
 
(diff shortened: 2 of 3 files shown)


sorja:
Plan is to have a single tooltip component in `inputs/MultiSelect/TooltipMultiSelect` used by both `CountryMultiSelect` and `MultiSelect`.

This requires a bit more work/effort but I think it's a good idea.

Otherwise, if preferred, we can merge this as it is to have two separate tooltip components.

@minotogna Let me know what you prefer

minotogna:
> Plan is to have a single tooltip component in `inputs/MultiSelect/TooltipMultiSelect` used by both `CountryMultiSelect` and `MultiSelect`.
> 
> 
> 
> This requires a bit more work/effort but I think it's a good idea.
> 
> 
> 
> Otherwise, if preferred, we can merge this as it is to have two separate tooltip components.
> 
> 
> 
> @minotogna Let me know what you prefer

@sorja countries selector have multiple columns and region header . Not sure we should make it generic; what did you have in mind ?

sorja:
> > Plan is to have a single tooltip component in `inputs/MultiSelect/TooltipMultiSelect` used by both `CountryMultiSelect` and `MultiSelect`.
> > This requires a bit more work/effort but I think it's a good idea.
> > Otherwise, if preferred, we can merge this as it is to have two separate tooltip components.
> > @minotogna Let me know what you prefer
> 
> @sorja countries selector have multiple columns and region header . Not sure we should make it generic; what did you have in mind ?

Single component for multi select:
- Generic for grouped and non-grouped
- Region header is a group header, omit if not grouped
- Multiple columns omitted if no groups

Maybe it adds too much complexity?

minotogna:
> > > Plan is to have a single tooltip component in `inputs/MultiSelect/TooltipMultiSelect` used by both `CountryMultiSelect` and `MultiSelect`.
> 
> > > This requires a bit more work/effort but I think it's a good idea.
> 
> > > Otherwise, if preferred, we can merge this as it is to have two separate tooltip components.
> 
> > > @minotogna Let me know what you prefer
> 
> > 
> 
> > @sorja countries selector have multiple columns and region header . Not sure we should make it generic; what did you have in mind ?
> 
> 
> 
> Single component for multi select:
> 
> - Generic for grouped and non-grouped
> 
> - Region header is a group header, omit if not grouped
> 
> - Multiple columns omitted if no groups
> 
> 
> 
> Maybe it adds too much complexity?
> 
> 
> 
> 

I think so . We only have two cases now and they are different . I would do this later on if necessary

minotogna (approved):

PR #6200, opened 2026-08-03
5758 - Navigation: Data errors - Performance - Visit links

yaguzmang:
Introducing two big optimizations to the `visitLinks` code:

1. We don't do HTTP requests to validate links, we do a DNS lookup. That means that a link like `http://www.example-link.com/report-1.pdf` and a link like `http://www.example-link.com/a-forest/some-other-report.pdf` are the same to the DNS resolver, so we can treat those two links as one hostname to resolve. This lowers the amount of lookups a lot, since many links share the same hostname. For example, the 2,996 links of fra/2025 become 1,037 unique hostnames.

2. The way we were batching DNS lookups was not the best. In a batch of 50, if 49 links took 1 second to resolve and 1 link took 5 seconds, then the entire batch took 5 seconds. The solution: `Promises.pool`. With this new method, we resolve 25 hostnames at a time, all the time. As soon as one lookup finishes, the next one starts.

Performance (in `validateAll`)

| Cycle | Links | Before | After | Speedup |
|---|---:|---:|---:|---:|
| fra/2020 | 516 | 46.2 s | 36.3 s | 1.3× |
| fra/2025 | 2,996 | 210.0 s | 41.6 s | 5.0× |
| fra/latest | 2,890 | 133.7 s | 31.8 s | 4.2× |
| fra/latest2 | 2,893 | 125.1 s | 22.5 s | 5.6× |
| panEuropean/2020 | 0 | 0.04 s | 0.01 s | n/a |
| panEuropean/2025 | 177 | 5.3 s | 1.8 s | 3.0× |
| **Total (link validation)** | **9,472** | **520.3 s** | **134.0 s** | **3.9×** |

Full `validateAll` run:  ~9.6 minutes → ~3.2 minutes:    **577 s → 193 s** (the remaining ~59 s are the table/data-source/NDP validations, unchanged by this PR).


src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/parseLink.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (Objects.isEmpty(link)) return { code: LinkValidationStatusCode.empty }
+
+  if (link.trim().toLowerCase().startsWith('mailto:')) {
+    return { code: LinkValidationStatusCode.success }
+  }
+
+  if (
+    link.startsWith('#_') ||
+    link.startsWith('api/cycle-data/repository/file/') ||
+    link.startsWith('/api/cycle-data/repository/file/')
```
minotogna: We should use ApiEndpoints

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/resolveHostname.ts
```diff
@@ -0,0 +1,22 @@
...
+export type HostnameResolution = 'resolved' | 'transient' | 'unresolved'
+
+// Error codes that count as 'transient'. Links with transient resolution errors are retried.
+const transientErrorCodes = new Set<string>([dns.CANCELLED, dns.CONNREFUSED, dns.REFUSED, dns.SERVFAIL, dns.TIMEOUT])
+
+const _getErrorCode = (error: unknown): string => (error as NodeJS.ErrnoException).code ?? ''
+
+const _isTransientError = (error: unknown): boolean => transientErrorCodes.has(_getErrorCode(error))
+
+export const resolveHostname = async (hostname: string): Promise<HostnameResolution> => {
```
minotogna, line 15: Do we have a HTTP utility or something similar ?

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/parseLink.ts
yaguzmang: Done 👍

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/resolveHostname.ts
yaguzmang: Not really, this uses `dns.promises.Resolver` which is from node directly. We don't really use HTTP, we only resolve the dns (at least for now, we will see later)

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/parseLink.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const repositoryFilePrefix = ApiEndPoint.CycleData.Repository.File.one('') // '/api/cycle-data/repository/file/'
+
+// Parses a link before resolving the hostname. Returns either a final status code or the hostname still to be resolved.
+export const parseLink = (link: string | null): ParsedLink => {
+  if (Objects.isEmpty(link)) return { code: LinkValidationStatusCode.empty }
+
+  if (link.trim().toLowerCase().startsWith('mailto:')) {
+    return { code: LinkValidationStatusCode.success }
+  }
```
sorja: Are emails validated? Should they?
E.g. if it's malformed: mailto:http://example.com/

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/parseLink.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const isRepositoryFileLink = link.startsWith(repositoryFilePrefix) || link.startsWith(repositoryFilePrefix.slice(1))
```
sorja: Are there any cases for ```link.startsWith(repositoryFilePrefix.slice(1)```?

sorja:
Good idea grouping and removing duplicated hostnames

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/parseLink.ts
yaguzmang: Good idea, added a check for it 👍 

(I checked the DB and we don't currently have any malformed ones, but this is good in case)

src/server/worker/tasks/verifyLinks/visitCycleLinks/utils/parseLink.ts
yaguzmang: There were before but not anymore, good catch :)

src/i18n/resources/en/admin.json
```diff
@@ -7,6 +7,7 @@
   "country_two": "$t(admin.country_other)",
   "country_zero": "$t(admin.country_other)",
   "disabledUsers": "Disabled Users",
+  "emailParsingError": "Invalid email address",
```
minotogna: I wouldn't put this under admin , but rather common validation 
And the key better invalidEmailAddress

yaguzmang: Done 👍

src/meta/cycleData/links/getI18nValidationStatusLabelKey.ts
```diff
@@ -1,7 +1,13 @@
...
+  if (
+    [
+      LinkValidationStatusCode.success,
+      LinkValidationStatusCode.empty,
+      LinkValidationStatusCode.invalidEmailAddress,
+    ].includes(code)
+  ) {
     return `common.${code}`
   }
   return `admin.${code}`
```
minotogna: Why are we storing error messages in admin or common ?
Can you please double check , if they are only used in validation , we should have a specific error messages i18n key .

yaguzmang: We use them for:
- Links Status table
- Status filter options
- CSV export
- Validation messages

I agree better grouping them, like:
```
generalValidation.linkStatus.success
generalValidation.linkStatus.empty
generalValidation.linkStatus.enotfound
generalValidation.linkStatus.invalidEmailAddress
generalValidation.linkStatus.urlParsingError
```

yaguzmang: or since these labels are not used exclusively in validation messages, we could introduce a dedicated `linkValidation.json` namespace:

```text
linkValidation.status.success
linkValidation.status.empty
linkValidation.status.enotfound
linkValidation.status.invalidEmailAddress
linkValidation.status.urlParsingError
```
What do you think? @minotogna

minotogna: Yes @yaguzmang only linkValidation 
But without status and meaningful name (I.e. the key must represent the message )

yaguzmang: Added it:
```json
{
  "dnsError": "DNS error",
  "empty": "Empty",
  "invalidEmailAddress": "Invalid email address",
  "success": "Success",
  "urlParsingError": "URL parsing error"
}

```
(used Deepl for the new translations)

minotogna (commented):
@sorja for your final approve

src/meta/cycleData/links/getI18nValidationStatusLabelKey.ts
minotogna: Nice 👍 thanks @yaguzmang

sorja (approved):

PR #6201, opened 2026-08-03
5758 - Navigation: Data errors - Performance - validateTableData tool

yaguzmang:
Performance improvement (`validateTableData` only):

| Before | After | Improvement |
|---:|---:|---:|
| 71.6 s | 33.8 s | **2.1× faster** |

### DIFF
--- a/src/server/service/dataValidation/tables/context/contextFactory.ts
+++ b/src/server/service/dataValidation/tables/context/contextFactory.ts
@@ -3,7 +3,6 @@ import { AssessmentMetaCaches } from 'meta/assessment/metaCaches'
 import { RowCacheKey } from 'meta/assessment/rowCache'
 import { RowCaches } from 'meta/assessment/rowCaches'
 import { TableName } from 'meta/assessment/table'
-import { Arrays } from 'utils/arrays'
 import { Promises } from 'utils/promises'
 
 import { RowRedisRepository } from 'server/cache/repository/row'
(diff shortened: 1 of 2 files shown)


src/server/service/dataValidation/tables/context/contextFactory.ts
```diff
@@ -34,15 +35,15 @@ export class ContextFactory extends BaseContextBuilder {
   }
 
   #isInQueue(variable: VariableCache): boolean {
-    const variableKey = this.#getVariableKey(variable)
-    return this.#queue.some((queued) => this.#getVariableKey(queued) === variableKey)
+    return this.#queueKeys.has(this.#getVariableKey(variable))
```
yaguzmang, line 38: I had removed this `o(1)` lookup using `Set` to mirror how it is done in the table data `ContextFactory`. 

For small incremental updates with a few nodes `this.#queue.some(...)` is fine, but in the tool `validateTableData` we pass a lot of nodes for all countries and this `O(n)` lookup becomes very expensive. This change alone sped up the tool by 2x.

Maybe we should use a `Set` in `tableData/updateDependencies/context/contextFactory.ts` for consistency. Let me know @minotogna :v:

minotogna: Hi @yaguzmang 
Sure 👍 please update it in this pr as well 
Thank you 😊

yaguzmang: Done 👍

minotogna (approved):

PR #6203, opened 2026-08-04
5758 - Navigation: Data errors - Performance - Improve validateTableData memory usage

yaguzmang:
Optimizing `validateTableData` by reducing duplicated work and limiting how many validation contexts exist at the same time:

- Assessment cycles are validated sequentially.
- Countries within each cycle are validated using `Promises.pool` with a concurrency of 8.
- Table metadata is fetched once per cycle.
- The list of nodes to validate is built once per cycle and shared by all its countries. Previously, both operations were repeated for every country.

Results (running for all assessment/cycles):
Memory

As a baseline, I measured the process after importing `validateTableData` without running it. This used **927.4 MiB of RSS** (the process's physical memory footprint).

| Measurement | Before | After | Reduction |
|---|---:|---:|---:|
| Peak memory | 4,139.3 MiB | 1,476.7 MiB | **2,662.6 MiB (64.3%)** |

Time

| Before | After | Improvement |
|---:|---:|---:|
| 31.644 s | 17.873 s | **43.5% faster (1.77×)** |

Next PR: Apply the same workflow to the other validation tools for consistency -> validate cycles sequentially and validate countries within each cycle using `Promises.pool`.


src/tools/validations/common/validateCycleCountries.ts
```diff
@@ -0,0 +1,46 @@
+import { Country } from 'meta/area/country'
+import { Assessment } from 'meta/assessment/assessment'
+import { Cycle } from 'meta/assessment/cycle'
+import { Promises } from 'utils/promises'
+
+import { Logger } from 'server/utils/logger'
+
+import { Failure } from './types'
+
+const countryValidationConcurrency = 8
```
sorja, line 10: should this be upper case with _ ?

```suggestion
const COUNTRY_VALIDATION_CONCURRENCY = 8
```

yaguzmang: Based on previous reviews, we would use camelcase (cc @minotogna )
I would use upper with _

minotogna (approved):

PR #6217, opened 2026-08-07
6211 - Recent Activity: Repository

### LINKED ISSUE #6211: Recent Activity: Repository
- Add to `acceptedMessages`:

ActivityLogMessage.repositoryItemCreate,
ActivityLogMessage.repositoryItemDelete,
ActivityLogMessage.repositoryItemUpdate,

- Refresh views

- Add translations
- Remove unused (file) translations

Check public.activity_log.repositoryItemUpdateAccess

sorja:
image

Last edit ok:

image

image


src/meta/assessment/activityLogs/index.ts
```diff
@@ -83,6 +88,15 @@ const getLabelAction = (props: { activity: ActivityLog<unknown>; t: TFunction })
   return label !== labelActionKey ? label : t('landing.recentActivity.actions.edited')
 }
 
+// Repository activities are logged with the repository item as target
+const _getRepositoryItemKey = (activity: ActivityLog<unknown>): string => {
+  const repositoryItem = activity.target as RepositoryItem
+
+  if (RepositoryItems.isFolder(repositoryItem)) return 'landing.recentActivity.actions.repositoryFolder'
```
minotogna: do we want to hide added a folder ? Mayne it's only relevant if users add file or link ?
What do you think @sorja ?

src/server/db/repository/assessmentCycle/countrySummary/_lastEditActivities.ts
```diff
@@ -16,6 +16,9 @@ export const activitiesLastEdit = [
   ActivityLogMessage.originalDataPointUpdateNationalClasses,
   ActivityLogMessage.originalDataPointUpdateOriginalData,
   ActivityLogMessage.originalDataPointUpdateYear,
+  ActivityLogMessage.repositoryItemCreate,
+  ActivityLogMessage.repositoryItemDelete,
+  ActivityLogMessage.repositoryItemUpdate,
```
minotogna, line 21: I'm still wondering if these activities should be considered for the last edit update ? does the report change if user add/remove/delete a repository item?
What do you think @sorja ?

sorja: I think not - they might add files like "hey, look at this report" or "check out these notes", but the report itself never changes - so the report "data" is not updated in my opinion

the report only changes IF the file or link uploaded in the repository is used in the platform - but then anyway that case is handled correctly

minotogna: nice. thank you @sorja

src/server/db/repository/assessmentCycle/countryActivityLog/createMaterializedView.ts
```diff
@@ -45,7 +46,8 @@ export const createMaterializedView = async (props: Props, client: BaseProtocol
                   and a.assessment_uuid = $2
                   and a.cycle_uuid = $3
                   and a.message in ($4:list)
-                  and a.section not in ($5:list)) as a
+                  and a.section not in ($5:list)
+                  and ${excludeFoldersClause}) as a
```
minotogna, line 50: maybe I would not create complication in the view itself. but rather in the query that returns the element. what do you think ?

sorja: I think the partition would create problems otherwise, what do you think?

minotogna: I mean in the where clause of the select in src/server/db/repository/assessmentCycle/countryActivityLog/getMany.ts, not the partition. 
we are already excluding node calculations there.

sorja: Doesn't work in this case:
- nodeCalculations are a separate thing, so excluding them there is ok
- folders are grouped by repositoryItem{Create,Delete,Edit} so they are in the same partition as editing a file or link.

=> Meaning that the proposed solution would drop a row in Recent Activity -view (e.g. add file -> add folder => both hidden)

minotogna: ok Miró, but this code can be inline (since it's used only here) and add a comment please

minotogna: oh I see it's used in two places/ never mind

minotogna (approved):

PR #6230, opened 2026-08-13
5810 - Upgrade TypeScript to 6.0.2

sorja:
Short explanation of changes:

|change|explanation|future?|
|---|---|---|
|ignoreDeprecations|Ignore deprecated (but not removed) props in config - baseUrl is used by tsc-alias, tsconfig-paths/register |```+    "paths": { "*": ["./src/*"] },```|
|strict|TS<5 defaulted to strict:false - our biggest change|strict:true + separate issue to fix broken parts of code|
|types|`@types/<package>` are not autoimported anymore|
|noUncheckedSideEffectImports| `vite/client` declares *.scss|
|files|for ts-node to load our .d.ts files, which nothing imports (e.g. Request.user)|

TS7 upgrade blocked by eslint-typescript (they will add support probably at around TS7.1)


minotogna (commented):
@sorja 
- paths should be added now
- list all deprecations 
 please :)

sorja:
> @sorja
> 
>     * paths should be added now
> 
>     * list all deprecations
>       please :)

# Paths

After investigating this a bit more, the solution is "either/or" paths or baseUrl - and in our case it's baseUrl:

Paths build on top of baseUrl, so if you have what is shown in description and keep baseUrl it translates to `src/...src/` which breaks

`baseUrl` still required by tsc-alias

We could introduce another issue to get rid of tsc-alias? What do you think? Or this can be done together with TS7?

We have 2 options (possibly more...)

bundling with the server:
esbuild/rolldown/etc (i think vite8 has rolldown?)

subpath imports (node resolves natively, ok with ts):
https://medium.com/@vitaliypotapov/setting-up-subpath-import-aliases-in-a-typescript-project-3ee027b75f1d
https://nodejs.org/api/packages.html#subpath-imports

# Deprecated

in config, the only deprecated property is `baseUrl`

PR #6238, opened 2026-08-14
6233 - Data entry: Database lock

sorja:


src/server/controller/cycleData/tableData/persistNodeValues/persistNodeValues.ts
```diff
@@ -34,7 +35,11 @@ export const persistNodeValues = async (props: Props, client: BaseProtocol = DB)
 
   await client.tx(async (client) => {
     try {
-      await client.func('pg_advisory_xact_lock', [1])
+      // Serialize writes per country
+      await client.query('select pg_advisory_xact_lock($(namespace), hashtext($(countryIso)))', {
+        countryIso,
```
minotogna: Should it be by assessment/cycle/country ?

minotogna (commented):
👍

src/server/controller/cycleData/tableData/persistNodeValues/persistNodeValues.ts
sorja: We have validations spanning over cycles and calculations spanning over assessments (panEu showing fra data)
I would keep like this, it brings no benefit to tighten the rule, let me know

minotogna (approved):

Issue #6237, opened 2026-08-13
Redis validation storage: prefix validation


minotogna:
in redis validation let's use only validation prefix (without data)
it's easier to get to validation and to distinguish from data

PR #6239, opened 2026-08-14
6237 - Redis validation storage: prefix validation

yaguzmang:
Closes #6237

### DIFF
--- a/src/server/cache/repository/keys.ts
+++ b/src/server/cache/repository/keys.ts
@@ -17,9 +17,9 @@ enum Assessments {
 
 enum Data {
   data = 'data',
-  validationDescriptions = 'data:validation:descriptions',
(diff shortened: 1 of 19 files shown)


src/server/cache/repository/keys.ts
```diff
@@ -17,9 +17,9 @@ enum Assessments {
 
 enum Data {
   data = 'data',
-  validationDescriptions = 'data:validation:descriptions',
-  validationNationalDataPoints = 'data:validation:nationalDataPoints',
-  validationTables = 'data:validation:tables',
+  validationDescriptions = 'validation:descriptions',
```
minotogna: Should we have its enum Validation ? What do you think @yaguzmang ?

yaguzmang: Yes, better. Added it 👍

minotogna (approved):

Issue #6234, opened 2026-08-13
Stress test


minotogna:

sorja:
I was using this to test.

You can get the token if you have test user:
```
curl -s -c cookies.txt -X POST http://localhost:9001/api/auth/login -F email=test@test.com -F password=password123
```

You need k6 installation separately, note that this is not ts-node script, then you can run it with:

```
k6 run -e TOKEN=$(grep fra-auth-token cookies.txt | awk '{print $7}') src/tools/loadTest/persistNodeValues.ts

or with token inline
k6 run -e TOKEN=..token.. src/tools/loadTest/persistNodeValues.ts
```

```
/** Very basic stress test script */

import { check } from 'k6'
import http from 'k6/http'

const baseUrl = __ENV.BASE_URL || 'http://localhost:9001'
const countryISOs = ['ALB', 'AUT', 'BEL', 'BGR', 'HRV', 'CYP', 'CZE', 'DNK'] // random group of countries (preferred: either use a node script to pull subset of countries by random or specify certain countries like FIN, ITA, etc..)

const headers = {
  'Content-Type': 'application/json',
  Cookie: `fra-auth-token=${__ENV.TOKEN}`,
}


export const options = {
  scenarios: {
    canary: {
      duration: '4m',
      exec: 'canary',
      executor: 'constant-arrival-rate',
      preAllocatedVUs: 10,
      rate: 2,
      timeUnit: '1s',
    },
    writers: {
      exec: 'write',
      executor: 'ramping-vus',
      stages: [
        { duration: '1m', target: 5 },
        { duration: '1m', target: 20 },
        { duration: '2m', target: 40 },
      ],
    },
  },
}


export const write = () =>{
  const countryIso = countryISOs[__VU % countryISOs.length]
  const params = `assessmentName=fra&cycleName=2025&countryIso=${countryIso}&sectionName=extentOfForest`
  const url = `${baseUrl}/api/cycle-data/table/nodes?${params}`

  const body = JSON.stringify({
    tableName: 'extentOfForest',
    values: [{ colName: '2025', value: { raw: String(1000 + __ITER) }, variableName: 'forestArea' }],
  })

  const response = http.patch(url, body, { headers, tags: { name: 'nodes' } })

  check(response, { 'write ok': (res) => res.status === 200 })
}

export const canary = () =>{
  const params =
    'assessmentName=fra&cycleName=2025&countryIso=FIN&countryISOs[]=FIN' +
[...]

PR #6243, opened 2026-08-17
6234 - Stress test - Add initial auth/config for k6

yaguzmang:
Adding the client to be used for the stress test requests, with login.


sorja:
Are you using K6?

minotogna:
> Are you using K6?

Good question. We should use k6

src/tools/stressTest/config.ts
```diff
@@ -0,0 +1,9 @@
+export interface StressTestConfig {
+  assessmentName: string
+  cycleName: string
+}
+
+export const config: StressTestConfig = {
+  assessmentName: 'fra',
```
minotogna: Also , I would add the configs in env vars

yaguzmang:
> Are you using K6?

 >>Good question. We should use k6 


@sorja @minotogna 
Based on the comments on https://github.com/orgs/openforis/projects/8/views/1?pane=issue&itemId=228256119&issue=openforis%7Cfra-platform%7C6234 I thought we wouldn't use k6.
So it's possible to install k6 in the staging server? as Miró explained in the issue, it needs to be installed separately

minotogna:
> > Are you using K6?
> 
> 
> 
>  >>Good question. We should use k6 
> 
> 
> 
> 
> 
> @sorja @minotogna 
> 
> Based on the comments on https://github.com/orgs/openforis/projects/8/views/1?pane=issue&itemId=228256119&issue=openforis%7Cfra-platform%7C6234 I thought we wouldn't use k6.
> 
> So it's possible to install k6 in the staging server? as Miró explained in the issue, it needs to be installed separately

Can't we add it to dev dependency @sorja @yaguzmang ?

sorja:
We don't want to install the tool anywhere else than locally

As it's a tool we use to test, it would be weird to install it in review?

Maybe I'm not following

Running the tool in review would itself spend resources of the server, thus giving us incorrect results

sorja:
> > > Are you using K6?
> > 
> > 
> > 
> >  >>Good question. We should use k6 
> > 
> > 
> > 
> > 
> > 
> > @sorja @minotogna 
> > 
> > Based on the comments on https://github.com/orgs/openforis/projects/8/views/1?pane=issue&itemId=228256119&issue=openforis%7Cfra-platform%7C6234 I thought we wouldn't use k6.
> > 
> > So it's possible to install k6 in the staging server? as Miró explained in the issue, it needs to be installed separately
> 
> Can't we add it to dev dependency @sorja @yaguzmang ?

K6 is a standalone install outside of node packages

minotogna:
> > > Are you using K6?
> > 
> > 
> > 
> >  >>Good question. We should use k6 
> > 
> > 
> > 
> > 
> > 
> > @sorja @minotogna 
> > 
> > Based on the comments on https://github.com/orgs/openforis/projects/8/views/1?pane=issue&itemId=228256119&issue=openforis%7Cfra-platform%7C6234 I thought we wouldn't use k6.
> > 
> > So it's possible to install k6 in the staging server? as Miró explained in the issue, it needs to be installed separately
> 
> Can't we add it to dev dependency @sorja @yaguzmang ?

Probably not 
But it can be installed with brew for example .

minotogna:
> > > > Are you using K6?
> > > 
> > > 
> > > 
> > >  >>Good question. We should use k6 
> > > 
> > > 
> > > 
> > > 
> > > 
> > > @sorja @minotogna 
> > > 
> > > Based on the comments on https://github.com/orgs/openforis/projects/8/views/1?pane=issue&itemId=228256119&issue=openforis%7Cfra-platform%7C6234 I thought we wouldn't use k6.
> > > 
> > > So it's possible to install k6 in the staging server? as Miró explained in the issue, it needs to be installed separately
> > 
> > Can't we add it to dev dependency @sorja @yaguzmang ?
> 
> K6 is a standalone install outside of node packages
> 

I'm also not following.
K6 must be installed on the developer machine and then we load the env to target the relative server .
In that comment i don't see we shouldn't use k6 , it's actually an example of using it !

yaguzmang:
> Running the tool in review would itself spend resources of the server, thus giving us incorrect results

Yes. Because of that, if we want to use k6, the ideal scenario is we run the stress test remotely. But in the comments on the issue we said we shouldn't execute the stress test remotely.

src/tools/stressTestK6/auth.ts
```diff
@@ -0,0 +1,27 @@
+import http from 'k6/http'
+
+import { baseUrl } from './config.ts'
+
+// Runs in k6 setup(), once per run. Returns the fra-auth-token cookie value used by every request.
+export const getToken = (): string => {
+  if (baseUrl.toLowerCase().includes('fra-data.fao.org')) {
+    throw new Error(`refusing to run against production host ${baseUrl}`)
+  }
```
yaguzmang: Added this to avoid an oopsie

src/tools/stressTestK6/run.sh
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+# Read the env file and export its vars for k6
+while IFS= read -r line || [ -n "$line" ]; do
+  case "$line" in
+    [A-Za-z_]*=*) export "$line" ;;
+  esac
+done < "$envFile"
+
+# TODO: exec k6 run src/tools/stressTestK6/<stress test scenario>
```
yaguzmang: k6 is not a Node process, so it can't use `scriptInit`/`dotenv` and has no way to read env files itself. The only inputs a k6 script sees are `-e` flags and the process environment.  This wrapper reads `.env.$NODE_ENV_SCRIPT` (like we do for other tools) and exports the vars before launching k6. This way we don't need to type everything to the terminal.

sorja:
This looks quite complicated, should we pass the token from file?

```
k6 run -e TOKEN=$(grep fra-auth-token cookies.txt | awk '{print $7}') src/tools/loadTest/persistNodeValues.ts
```

You can get the token pretty easily with curl:

```
curl -s -c cookies.txt -X POST http://localhost:9001/api/auth/login -F email=test@test.com -F password=password123
```

yaguzmang:
> This looks quite complicated, should we pass the token from file?

@sorja Even if we pass the token, we still need to load the other env vars. 

From the issue: 
> [minotogna](https://github.com/minotogna)
[1d ago](https://github.com/openforis/fra-platform/issues/6234#issuecomment-5315968511)
from the local machine with tools. add it to dev dependencies and use NODE_ENV_SCRIPT to target the environment

We could also skip env files entirely and run it like this:

```bash
curl -s -c cookies.txt -X POST http://localhost:9001/auth/login -F email=test@test.com -F password=pass

k6 run \
  -e APP_URI=http://localhost:9001 \
  -e TOKEN=$(grep fra-auth-token cookies.txt | awk '{print $7}') \
  -e STRESS_TEST_ASSESSMENT_NAME=fra \
  -e STRESS_TEST_CYCLE_NAME=2025 \
  -e STRESS_TEST_COUNTRIES=X09,X10 \
  -e STRESS_TEST_EDITORS=100 \
  -e STRESS_TEST_DURATION=5m \
  src/tools/stressTestK6/<scenario>
```

versus:

```bash
NODE_ENV_SCRIPT=staging ./src/tools/stressTestK6/run.sh
```

eslint.config.mjs
```diff
@@ -15,7 +15,13 @@ import globals from 'globals'
 
 export default defineConfig([
   {
-    ignores: ['**/node_modules/**', '**/dist/**', '**/build/**', 'src/tools/migrations/steps/steps/template.ts'],
+    ignores: [
+      '**/node_modules/**',
+      '**/dist/**',
+      '**/build/**',
+      'src/tools/migrations/steps/steps/template.ts',
+      'src/tools/stressTestK6/**',
```
minotogna: Why should we ignore this ?

.env.template
```diff
@@ -63,5 +63,9 @@ FRA_AUTH_TOKEN=
 FRA_REPORT_ASSESSMENT=fra
 FRA_REPORT_CYCLE=2025
 
+# Stress test (src/tools/stressTestK6)
```
minotogna: I would append those at the end of the file . Please add new line at the end as well

minotogna:
> > This looks quite complicated, should we pass the token from file?
> 
> 
> 
> @sorja Even if we pass the token, we still need to load the other env vars. 
> 
> 
> 
> From the issue: 
> 
> > [minotogna](https://github.com/minotogna)
> 
> [1d ago](https://github.com/openforis/fra-platform/issues/6234#issuecomment-5315968511)
> 
> from the local machine with tools. add it to dev dependencies and use NODE_ENV_SCRIPT to target the environment
> 
> 
> 
> We could also skip env files entirely and run it like this:
> 
> 
> 
> ```bash
> 
> curl -s -c cookies.txt -X POST http://localhost:9001/auth/login -F email=test@test.com -F password=pass
> 
> 
> 
> k6 run \
> 
>   -e APP_URI=http://localhost:9001 \
> 
>   -e TOKEN=$(grep fra-auth-token cookies.txt | awk '{print $7}') \
> 
>   -e STRESS_TEST_ASSESSMENT_NAME=fra \
> 
>   -e STRESS_TEST_CYCLE_NAME=2025 \
> 
>   -e STRESS_TEST_COUNTRIES=X09,X10 \
> 
>   -e STRESS_TEST_EDITORS=100 \
> 
>   -e STRESS_TEST_DURATION=5m \
> 
>   src/tools/stressTestK6/<scenario>
> 
> ```
> 
> 
> 
> versus:
> 
> 
> 
> ```bash
> 
> NODE_ENV_SCRIPT=staging ./src/tools/stressTestK6/run.sh
> 
> ```

@sorja @yaguzmang we
Should keep things simple and documented 

The first stress test case has been  written . That is our first objective 
Maybe we don't need the env ? 
But do we need a wrapper script that store the token locally ?
E.g. ./path/command.sh host user password 

Also there should be a Readme inside the stress test folder to document it. 

What do you think ?

The important thing is must be simple to use

yaguzmang:
> > > This looks quite complicated, should we pass the token from file?
> > 
> > 
> > @sorja Even if we pass the token, we still need to load the other env vars.
> > From the issue:
> > > [minotogna](https://github.com/minotogna)
> > 
> > 
> > [1d ago](https://github.com/openforis/fra-platform/issues/6234#issuecomment-5315968511)
> > from the local machine with tools. add it to dev dependencies and use NODE_ENV_SCRIPT to target the environment
> > We could also skip env files entirely and run it like this:
> > ```shell
> > curl -s -c cookies.txt -X POST http://localhost:9001/auth/login -F email=test@test.com -F password=pass
> > 
> > 
> > 
> > k6 run \
> > 
> >   -e APP_URI=http://localhost:9001 \
> > 
> >   -e TOKEN=$(grep fra-auth-token cookies.txt | awk '{print $7}') \
> > 
> >   -e STRESS_TEST_ASSESSMENT_NAME=fra \
> > 
> >   -e STRESS_TEST_CYCLE_NAME=2025 \
> > 
> >   -e STRESS_TEST_COUNTRIES=X09,X10 \
> > 
> >   -e STRESS_TEST_EDITORS=100 \
> > 
> >   -e STRESS_TEST_DURATION=5m \
> > 
> >   src/tools/stressTestK6/<scenario>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > versus:
> > ```shell
> > NODE_ENV_SCRIPT=staging ./src/tools/stressTestK6/run.sh
> > ```
> 
> @sorja @yaguzmang we Should keep things simple and documented
> 
> The first stress test case has been written . That is our first objective Maybe we don't need the env ? But do we need a wrapper script that store the token locally ? E.g. ./path/command.sh host user password
> 
> Also there should be a Readme inside the stress test folder to document it.
> 
> What do you think ?
> 
> The important thing is must be simple to use

@minotogna I agree, we can use `./src/tools/stressTestK6/run.sh <host> <email> <password>`. And we don't need to store the token locally because the script logs in by itself in k6's `setup()` (see `auth.ts`) and reuses the cookie for the run. Nothing is written to disk.
Instead of env vars then we can have the other config vars in plain text in the code, like assessment, cycle, number of editors. They can just be constants at the top of the stress test script, and k6 flags like `-e EDITORS=5` can still override them.

eslint.config.mjs
yaguzmang: Because these files run inside k6, not Node, so some rules conflict with how k6 loads modules:
- `k6` and `k6/http` are k6 built-in modules, not npm packages. They don't exist in `package.json` or `node_modules`, so import rules flag them.
- The k6 loader requires explicit `.ts` extensions on local imports (`./config.ts`). Our `import/extensions` rules doesn't allow that.

We still get formatting when committing though

src/tools/stressTestK6/README.md
```diff
@@ -0,0 +1,51 @@
...
+edit the target countries (X09 and X10 by default), or an ADMINISTRATOR account.
+
+The script logs in once, in k6 `setup()`, and reuses the `fra-auth-token` cookie for the whole run.
+No token or cookie file is stored. If you already have a valid `fra-auth-token` value, you can skip the
+login by running k6 directly:
+
+```bash
+k6 run -e HOST=http://localhost:9001 -e TOKEN=token123 src/tools/stressTestK6/<stress test script>
+```
+
```
yaguzmang: CC @sorja  
We use `./src/tools/stressTestK6/run.sh http://localhost:9001 test@test.com password123`, but if you have the token already you can use this.

yaguzmang: I will update the readme in the next PR when I add the first actual stress test code

minotogna (commented):
@sorja for me this looks good 
For your final review / approval

src/tools/stressTestK6/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+# Runs the stress test against the given environment. See README.md.
+set -e
+
+if [ $# -ne 3 ]; then
+  echo 'Usage: ./src/tools/stressTestK6/run.sh <host> <email> <password>' >&2
+  echo 'e.g.:  ./src/tools/stressTestK6/run.sh http://localhost:9001 test@test.com password123' >&2
```
sorja: no need to redirect output to `stderr` here, i think?

src/tools/stressTestK6/run.sh
```diff
@@ -0,0 +1,13 @@
...
+  exit 1
+fi
+
+cd "$(dirname "$0")/../../.."
```
sorja: why you need to change the dir here? for write later?

tsconfig.base.json
```diff
@@ -1,6 +1,10 @@
 {
   "include": ["src"],
-  "exclude": ["src/tools/migrations/steps/steps/template.ts"],
+  "exclude": [
+    "src/tools/migrations/steps/steps/template.ts",
+    // k6 scripts run in k6's own runtime, outside the project.
```
sorja: json doesnt support comments, you can add (fake) comments using
"//": "k6...."

src/tools/stressTestK6/run.sh
sorja: also i noticed you are checking the params twice, in `config.ts` and here, should we do it only in one place?

maybe you can check here?

sorja:
some minor comments and thoughts

src/tools/stressTestK6/run.sh
yaguzmang: Usage/error text goes to stderr so if we run it and pipe to a file like ` ./run.sh ... > results.txt` the error would be lost.

and about the params check:
It looks duplicated but they cover different cases. `run.sh` only checks that you called it with the three arguments, so you get a quick usage message. The checks in `config.ts` and `auth.t`s are for when k6 is run directly with `-e TOKEN=...`, and that path doesn't go through `run.sh` at all. If we keep just one of them, the other way of running it can fail with some other less helpful error.

src/tools/stressTestK6/run.sh
yaguzmang: It was just for executing, so just `"$(dirname "$0")/<stress test script>"` would do. I made the change, thanks for pointing it out 👍

tsconfig.base.json
yaguzmang: tsconfig files are JSONC and allow comments, but I removed it anyway because it is redundant now that we have the readme 👍

sorja (approved):

PR #6254, opened 2026-08-20
6234 - Stress test - Data entry: Tables & NDP

yaguzmang:


src/tools/stressTest/config.ts
```diff
@@ -0,0 +1,25 @@
...
+export const assessmentName = __ENV.ASSESSMENT_NAME || 'fra'
+export const cycleName = __ENV.CYCLE_NAME || '2025'
+
+// Auth: either a ready-made token, or credentials to log in with (see auth.ts)
+export const token = __ENV.TOKEN
+export const email = __ENV.STRESS_TEST_EMAIL
+export const password = __ENV.STRESS_TEST_PASSWORD
+
+// Load shape
+export const users = Number(__ENV.USERS || 100)
```
yaguzmang: Even only at 40 users the run fails:

```bash
USERS=40 DURATION=1m ./src/tools/stressTest/run.sh http://localhost:9001 test@test.com password123
```

Image

Most requests fail due to timeouts. 
I am investigating and already found a couple of issues. Writes queue up waiting for the country lock, and each one holds a database connection while it waits. With enough concurrent editors all connections are held by the queue, so everything times out until the server is restarted. 

I will add some fixes in a coming PR. Leaving this one as draft for now.

yaguzmang: Fixed in #6264

src/tools/stressTest/dataEntry/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    getValidationSummary(headers, countryIso)
+  } else if (roll <= getValidationSummaryWeight + getTableDataWeight) {
+    getTableData(headers, countryIso)
+  } else if (roll <= getValidationSummaryWeight + getTableDataWeight + editTableCellsWeight || ndps.length === 0) {
+    editTableCells(headers, countryIso)
+  } else {
+    editNdp(headers, countryIso, ndps)
+  }
+
+  sleep(randomInt(pauseMinSeconds, pauseMaxSeconds))
```
minotogna: The if else logic is a bit difficult to understand.
Maybe it's better if we have one stress test per case ?
Data entry 
Edit NDP 
Edit descriptions 

@sorja what do you think ?

sorja: Sounds good

I would do something similar to:

src/tools/stressTest/tableData.ts
src/tools/stressTest/ndpData.ts
src/tools/stressTest/descriptions.ts


--

Something else:


Not sure _where_ we should fetch validations (GET request), for me it would make more sense to:

in case of {table data / ndp / description}:
write: {}
read: {}

Means:
when stress testing table data, I would also try to fetch the same data
when stress testing ndp (e.g. year 2020), i would try to fetch the same ndp to see that it responds
etc

Does this make sense?


-- After thought --
Maybe we can fetch both validations AND table data in the canary?

src/tools/stressTest/ndp/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    getValidationSummary(headers, countryIso)
+  } else if (roll <= getValidationSummaryWeight + getTableDataWeight) {
+    getTableData(headers, countryIso)
+  } else if (roll <= getValidationSummaryWeight + getTableDataWeight + editTableCellsWeight || ndps.length === 0) {
+    editTableCells(headers, countryIso)
+  } else {
+    editNdp(headers, countryIso, ndps)
+  }
+
+  sleep(randomInt(pauseMinSeconds, pauseMaxSeconds))
```
yaguzmang: Done. I split it into one test per case: `tableData/` and `ndp/`
run.sh now takes the test as an argument:

```bash
./src/tools/stressTest/run.sh <host> <email> <password>        # tableData (default)
./src/tools/stressTest/run.sh <host> <email> <password> ndp
```

@sorja I added a canary at a fixed rate (2/s) that fetches the data being written and its validations 👍

sorja:
Should we have util for urls?

Something like

```
Urls.getTableRead({ params })

```

src/tools/stressTest/tableData/getTableData.ts
```diff
@@ -0,0 +1,26 @@
...
+    `${baseUrl}/api/cycle-data/table/table-data?${params}&countryISOs[]=${countryIso}&mergeOdp=false`,
+    { headers, tags: { name: 'table/table-data GET' } }
+  )
+  check(tableData, { 'table data ok': (res) => res.status === 200 })
+
+  const validations = http.get(`${baseUrl}/api/cycle-data/validations/table-data?${params}`, {
+    headers,
+    tags: { name: 'validations/table-data GET' },
+  })
+  check(validations, { 'table validations ok': (res) => res.status === 200 })
```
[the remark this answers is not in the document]
yaguzmang: `Requests.sendOk` sends the requests, but `check` here is actually reading it. I will add   `RequestUtils.isOk` 👍

minotogna (commented):
@sorja for your final review .
To me  it looks good

sorja (approved):

